PR #1, opened 2020-09-24
Upgrade Font Awesome to 5.14.0

MakisH:
This upgrades [Font Awesome](https://fontawesome.com/) from version 4.7.0 to 5.14.0, giving us access to more and better [free icons](https://fontawesome.com/icons?d=gallery&m=free).

@chlorenz I still cannot figure out the external link icon (see review) and I a hope I have not missed any icons.


css/customstyles.css
```diff
@@ -929,8 +929,8 @@ span.soft {
 
 /* this part adds an icon after external links, using FontAwesome*/
 a:not(.notExternal)[href^="http://"]:after, a:not(.notExternal)[href^="https://"]:after {
-    content: "\f08e";
-    font-family: FontAwesome;
+    content: "\f35d";
+    font-family: "Font Awesome 5 Free Solid";
```
MakisH: I assume that these are the changes needed to upgrade the external link icon from [version 4](https://fontawesome.com/v4.7.0/icon/external-link) to [version 5](https://fontawesome.com/icons/external-link-alt?style=solid), but it does not work for me. Any ideas?

chlorenz: Turns out there is a "known issue" in FontAwesome 5: https://github.com/FortAwesome/Font-Awesome/issues/11946
To use Unicode code points the font-family has to be "Font Awesome\ 5 Free" and you have to set font-weight: 900 manually.
Credits go to https://stackoverflow.com/questions/17254339/font-awesome-unicode

chlorenz (approved):

css/customstyles.css
MakisH: Ah... Then good thing I did not spend any more time on this crazy bug. :sweat_smile: Thanks a lot!

PR #2, opened 2020-09-24
Add a disclaimer alert

MakisH:
Add an alert type for disclaimers. Main use case:

Screenshot_20200924_120424

I am not sure if I have edited everything I needed. This is still an information box and I guess we don't want to add more logic apart from distinguishing it (or similar alerts) from other alerts.

@chlorenz @uekerman what do you think?

:warning: Depends on #1.


chlorenz (approved):
I like the custom use of the callouts and approve the technical implementation. I agree that this alert should be on the same level as "information".

uekerman:
Looks good to me :+1: 
But let's not overuse the OpenFOAM one. It draws quite some attention to sth completely useless. Let's always put it at the very bottom.

PR #4, opened 2020-10-26
Literature migration

oguzziya:


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (6.0.3.2)
```
MakisH: Why the updates in this file? It seems to build without them.

I suggest that we remove the changes in this file.

_data/sidebars/docs_sidebar.yml
```diff
@@ -295,8 +295,20 @@ entries:
...
+    - title: Starting points
+      url: /literature-guide-starting-points.html
+      output: web, pdf
+
+    - title: preCICE features
+      url: /literature-guide-precice-features.html
+      output: web, pdf
+    
+    - title: Adapters and coupled codes
+      url: /literature-guide-adapters-and-coupled-codes.html
```
MakisH: I am not sure if we need so many small pages for each section. Let's better move everything to a single page.

MakisH (changes requested):
The page builds nicely with the latest master, but I have two comments regarding:
1. Unrelated changes to `Gemfile.lock`
2. Too fine structure for the amount of information we currently have.

pages/docs/literature-guide/literature-guide-overview.md
```diff
@@ -0,0 +1,34 @@
...
+
+* **Data mapping** For an introduction to the various techniques, have a look at the dissertations of Bernhard Gatzhammer (Sections 2.4 and 4.2) and Benjamin Uekermann (Section 4.3). For a more condensed overview of RBF mapping, see [Radial Basis Function Interpolation for Black-Box Multi-Physics Simulations](ftp://ftp.informatik.uni-stuttgart.de/pub/library/ncstrl.ustuttgart_fi/INPROC-2017-35/INPROC-2017-35.pdf)
+
+* **Communication** For an introduction to the various techniques, have a look at the dissertation of Bernhard Gatzhammer (Section 4.3). Have a look also at the master's thesis of Alexander Shukaev: "[A Fully Parallel Process-to-Process Intercommunication Technique for preCICE](https://www5.in.tum.de/pub/Shukaev2015_MasterThesis.pdf)".
+
+* **Time interpolation** This feature is currently under active development. Have a look at the publications, talks, and posters of [Benjamin Rüth](https://www5.in.tum.de/wiki/index.php/Benjamin_R%C3%BCth,_M.Sc._(hons)).
```
MakisH: New page in the meanwhile:

```suggestion
* **Time interpolation** This feature is currently under active development. Have a look at the publications, talks, and posters of [Benjamin Rüth](https://www.in.tum.de/en/i05/people/personen/benjamin-rueth/).
```

pages/docs/literature-guide/literature-guide-overview.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Literature guide
+permalink: literature-guide-overview.html
+keywords: literature, papers, numerics
+summary: "Theses and papers for theoretical background of preCICE"
+---
+
+Wherever the information in this documentation is not enough (in this case, [let us know](https://www.precice.org/resources/#contact)), you may find a wide spectrum of additional information in [publications](https://www.precice.org/publications/). This page will guide you through it.
```
MakisH, line 8: Two points that I currently don't know how to handle (@uekerman):
1. Where do we now link for the "contact us"? Maybe the top-level "community" page? Maybe "community/channels"? Inany case, we don't have this already. It would also make sense to just merge this as it is and then have a second iteration where we check/update all the links.
2.  Where do we link for the publications? From our [proposed structure](https://pad.gwdg.de/jxoCHHtJSYiZt6hS3XgPVA) I understand that in the community we put the "everything that cites preCICE". Should we add here a subpage with our publications?

pages/docs/literature-guide/literature-guide-overview.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Literature guide
+permalink: literature-guide-overview.html
+keywords: literature, papers, numerics
+summary: "Theses and papers for theoretical background of preCICE"
```
MakisH: Maybe something more general:
```suggestion
summary: "A guide to the main reference literature for each component and feature of preCICE"
```

MakisH (approved):
A small suggestion for the summary and an update on Benjamin Rüth's page. After fixing these, this PR can be merged.

But this is a good point to discuss what to do with the "contact us" and "publications" links.

pages/docs/literature-guide/literature-guide-overview.md
uekerman: concerning 1.
I think "community/channels" is good. For the moment you could simply put `community-channels.html`. We have to check all links in the end anyway. Yes, let's not make this a merge stopper.

concerning 2.
Yes, in our proposed structure, we put everything that cites preCICE to community. I guess our thinking was that the "literature overview" is indeed where we list all publications about preCICE. So I guess, we don't need another page there. Right?

MakisH: > I guess our thinking was that the "literature overview" is indeed where we list all publications about preCICE. So I guess, we don't need another page there. Right?

But still, [our current list of publications](https://www.precice.org/publications/) already has >50 entries. We probably want to organize this a bit better. We could merge this now and decide on what to do with the publications later.

PR #9, opened 2020-11-17
Add watch integral documentation

oguzziya:


pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Watch Integral Configuration
+permalink: configuration-watchintegral.html
+keywords: configuration, watchintegral
+summary: "With a watch integral, you can track the transient change of integral of a coupling data over the interface. This is especially useful when you want to track conserved quantities of your simulation, such as flow rate or total force at coupling mesh."
+---
+**NOTE:** Only available for preCICE version >= v2.2.0
```
uekerman:
```suggestion
{% include important.html content="This feature is only available for preCICE versions >= v2.2" %}
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Watch Integral Configuration
+permalink: configuration-watchintegral.html
+keywords: configuration, watchintegral
+summary: "With a watch integral, you can track the transient change of integral of a coupling data over the interface. This is especially useful when you want to track conserved quantities of your simulation, such as flow rate or total force at coupling mesh."
```
uekerman:
```suggestion
summary: "With a watch integral, you can track the transient change of integral values of coupling data over complete coupling meshes. This is especially useful when you want to track conserved quantities of your simulation, such as flow rate or the overall force acting on a geometry."
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
...
+```xml
+<participant name="MySolver1"> 
+    <use-mesh name="MyMesh1" provide="yes"/>
+    ...
+    <watch-integral mesh="MyMesh1" name="MyWatchIntegral" scale-with-connectivity="yes"/>
+    ...
+</participant>
+```
+
+This will create a logging file `precice-MySolver1-watchintegral-MyWatchIntegral.log` with one row per timestep. If `scale-with-connectivity` is set and there is a connectivity information, `SurfaceArea` is also going to be included as a column in the logging file.
```
uekerman:
```suggestion
This creates a log file `precice-MySolver1-watchintegral-MyWatchIntegral.log` with one row per time window and integral values of coupling data per column. If `scale-with-connectivity` is set to `yes` and there is connectivity information defined (find out more about mesh connectivity in [step 8 of the couple-your-code section](couple-your-code-defining-mesh-connectivity.html)), surface area is included as an additional column. 
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
...
+<participant name="MySolver1"> 
+    <use-mesh name="MyMesh1" provide="yes"/>
+    ...
+    <watch-integral mesh="MyMesh1" name="MyWatchIntegral" scale-with-connectivity="yes"/>
+    ...
+</participant>
+```
+
+This will create a logging file `precice-MySolver1-watchintegral-MyWatchIntegral.log` with one row per timestep. If `scale-with-connectivity` is set and there is a connectivity information, `SurfaceArea` is also going to be included as a column in the logging file. 
+* Only a participant that provides the respective mesh can set a watch integral on that mesh.
```
uekerman:
```suggestion
* Only a participant that provides a mesh can set a watch integral on this mesh.
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
...
+    * **No connectivity information:** Integral is the summation of coupling data over the mesh, no matter what `scale-with-connectivity` is.
+    * **2D - Edge Connectivity:** If `scale-with-connectivity` is set `yes`, data values will be scaled using midpoint rule, i.e the average of two vertex data is taken and multiplied with corresponding edge length. Otherwise, the integral will be the same as no connectivity case.
+    * **3D - Face Connectivity:** If `scale-with-connectivity` is set `yes`, data values will be scaled such that the average of three vertex data is taken and multiplied with corresponding face area. Otherwise, the integral will be the same as no connectivity case.
+* Some important points for the interpretation of the integral data:
+    * Integral calculation is based on weighted sum of vertex data and does not distinguish between conservative and consistent data. For example, watch integral is useful for calculating the total flow rate over the interface, or total force which are both conserved variables. However, using watch integral for stress data would not be useful since summing up total stress has no physical interpretation.
```
uekerman: I think it might be better to re-organize this part: First explain that there are two way how integral values are computed. Then explain when which case is needed. And finally, what is needed to use either case and how to switch it in on or off.

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,27 @@
...
+    * Scaling of the data is going to be always based on the mesh given in configuration.
```
uekerman: Probably better to move this up. You could also use the name `MyMesh1` to make it more obvious what you want to say here.

uekerman (changes requested):
Needs a bit of re-organization. In general, try to avoid future tense in the documentation ("will ..."). The only proper use case is to say that sth will happen in the future (e.g. what we write in our roadmap).

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,30 @@
...
+    <watch-integral mesh="MyMesh1" name="MyWatchIntegral" scale-with-connectivity="yes"/>
+    ...
+</participant>
+```
+
+This creates a log file `precice-MySolver1-watchintegral-MyWatchIntegral.log` with one row per time window and integral values of coupling data per column. If `scale-with-connectivity` is set to `yes` and there is connectivity information defined (find out more about mesh connectivity in [step 8 of the couple-your-code section](couple-your-code-defining-mesh-connectivity.html)), surface area is included as an additional column. 
+* Only a participant that provides a mesh can set a watch integral on this mesh.
+* You can freely choose the name `MyWatchIntegral`.
+* There are two ways to calculate integral data:
+    * **Calculate with scaling:** While calculating the integral values, area weighted sum of vertex data is calculated. It is useful when you want to calculate flow rate, for example.
```
davidscn:
```suggestion
    * **Calculate with scaling:** While calculating the integral values, area weighted sum of vertex data is calculated. This approach is useful when your data is not yet associated to any cell size such as a flow rate or a displacement field. In case your data is already associated with a cell size (e.g. a force acting on a cell face), this option is usually not required. Data, where a scaling is reasonable, is usually mapped `consistent` between participants.
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,30 @@
...
+    This option requires mesh connectivity information (edges for 2D, faces for 3D) and `scale-with-connectivity` option is set to `yes`. For 2D, edge lengths are used while for 3D face areas are used for scaling.
+
+    * **Calculate without scaling:** The coupling data is summed up over all vertices on the coupling mesh. This is useful when you want to calculate the overall force acting on the geometry. If there is no mesh connectivity information, this way is used to calculate the integral. If there is mesh connectivity information, you can switch of the scaling by setting `scale-with-connectivity` option to `no`.
```
davidscn:
```suggestion
    * **Calculate without scaling:** The coupling data is summed up over all vertices on the coupling mesh. This is useful when your coupling data is already associated to a certain cell size (e.g. a force acting on a cell face) of your coupling mesh since no additional weighting of the coupling data is applied. Data, where a scaling is not reasonable, is usually mapped `conservative` between participants. If there is no mesh connectivity information provided, no scaling is performed to calculate the integral. If there is mesh connectivity information, you can switch of the scaling by setting `scale-with-connectivity` option to `no`.
```

pages/docs/configuration/configuration-watchintegral.md
```diff
@@ -0,0 +1,30 @@
...
+
+* Some important points for the interpretation of the integral data:
+    * Integral calculation is based on weighted sum of vertex data and does not distinguish between conservative and consistent data. For example, watch integral is useful for calculating the total flow rate over the interface, or total force which are both conserved variables. However, using watch integral for stress data would not be useful since summing up total stress has no physical interpretation.
+    * Scaling of the data is going to be always based on the connectivity information of the mesh given in configuration. For example, in the given configuration file, `MyWatchIntegral` is defined on mesh `MyMesh1`. Each of the coupling data used by `Mesh1` is scaled with the connectivity information of `Mesh1`.
```
davidscn:
```suggestion
    * Scaling of the data is going to be always based on the connectivity information of the mesh given in configuration. For example, in the given configuration file, `MyWatchIntegral` is defined on mesh `MyMesh1`. Each of the coupling data used by `Mesh1` is scaled with the connectivity information of `Mesh1`. If the configuration and connectivity information of `Mesh1` is not updated throughout the simulation, the scaling is always based on the initial configuration of your geometry. You need to ensure that your exchanged coupling data refers to the same configuration.
```

davidscn (commented):
Just a suggestion.

pages/docs/configuration/configuration-watchintegral.md
uekerman: Looks better now :+1:

pages/docs/configuration/configuration-watchintegral.md
uekerman: In principle yes, but might be hard to understand:
* what do you mean by configuration of `Mesh1`? 
* connectivity can currently not be updated throughout the simulation. But this is also not the point, right? Connectivity (who is connected to whom) stays the same throughout the simulation. 

I guess it would just be enough to say  
"If your solver uses a Lagrangian or ALE description please note that the scaling is based on the initial reference vertex coordinates of your geometry."
and this could probably be a separate item here.

uekerman (approved):

PR #10, opened 2020-11-17
Add macOS documentation

oguzziya:


oguzziya:
I just followed this approach and it should work. Have you faced with an additional issue while installing on the chair mac @fsimonis ?

PR #24, opened 2021-02-21
Provide more information on vanilla CalculiX

BenjaminRodenberg:
For review:

* Ideally you have never built CalculiX before.
* Install dependencies via `apt`
* Try to build vanilla CalculiX using the included Makefile
* Hopefully the documentation updates help you to find out how you have to modify the Makefile.
* If you already have some experience on include and library flags: Try to imagine how a user that has no idea what a include and library flag is would work with the documentation. Is it understandable even in this case?

### DIFF
--- a/pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
@@ -129,7 +129,8 @@ The source code is now in the `~/CalculiX/ccx_VERSION/src` directory. The adapte
 
 ### Building the "vanilla" CalculiX (optional)
 
-If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
+If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as `spooles.h:26:10: fatal error: misc.h: No such file or directory
+`. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
 
 ### Building the modified CalculiX
 
--- a/pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
@@ -129,7 +129,7 @@ The source code is now in the `~/CalculiX/ccx_VERSION/src` directory. The adapte
 
 ### Building the "vanilla" CalculiX (optional)
 
-If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. You may still run into issues with ARPACK in this case, but this is out of scope for our needs, as we use a different Makefile.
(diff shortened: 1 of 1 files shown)


pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
```diff
@@ -129,7 +129,7 @@ The source code is now in the `~/CalculiX/ccx_VERSION/src` directory. The adapte
 
 ### Building the "vanilla" CalculiX (optional)
 
-If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. You may still run into issues with ARPACK in this case, but this is out of scope for our needs, as we use a different Makefile.
+If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
```
MakisH: `such as` what? :wink:

BenjaminRodenberg: I'm now giving an example. See https://github.com/precice/precice.github.io/pull/24/commits/ebe33bf42d5401ca8088dcb79cac3244d17f3cb8

pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
```diff
@@ -129,7 +129,8 @@ The source code is now in the `~/CalculiX/ccx_VERSION/src` directory. The adapte
...
 
-If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
+If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as `spooles.h:26:10: fatal error: misc.h: No such file or directory
+`. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
```
MakisH: I would not break the line in an inline code environment:

```suggestion
If you want to build the "vanilla" (i.e. without preCICE) CalculiX, you can now run `make` inside the `src/` directory. Depending on how you installed the dependencies above (using `apt` or from source), you might get compilation errors, such as `spooles.h:26:10: fatal error: misc.h: No such file or directory`. Often these errors can be easily fixed by modifying CalculiX `Makefile`. Please refer to [our adapter's makefile options](https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options) for a list of library and include flag you might have to set depending on your installation procedure.
```

If you prefer to keep the line break, then better move it after the dot.

MakisH: Also, when you are adding URLs in a page inside the the `precice.github.io` repository, it is better to skip the domain name:

`https://www.precice.org/adapter-calculix-get-adapter.html#makefile-options` --> `adapter-calculix-get-adapter.html#makefile-options`

Reasons are that the domain can change, even in smaller parts: we now, e.g. prefer the form without the `www.`. In the past, we changed everything from `http` to `https`. Redirections may apply. It is better to just leave this to the server. :wink:

When you are editing a file in the tutorials repository, then use absolute URLs (as the tutorials should also be readable in a different context, outside of the website).

MakisH (approved):
A few technicalities to fix, but I think this is a useful change. :smile: 

(I don't fall into the category of reviewer you asked for, though)

pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
BenjaminRodenberg: After first accidentally messing up the commit, I fixed this now in 6c71e6d.

@MakisH  If I understand you correctly: If `https://www.precice.org` appears in any link of this repo, it's wrong. Is this rule something that we can put into a test? I find the new tests for dead links etc. extremely useful. If we could also use something similar to avoid unnecessarily adding absolute links starting with `https://www.precice.org...` I imagine this might be helpful, as well.

MakisH: I don't think it would be easy to check or enforce the format of the links. I want to add the same link-checking workflow we already have in the tutorials, but this needs time, as I will need to fix any broken links and add exceptions. It's on my list.

PR #27, opened 2021-03-08
Style attribute removed from css

dcs-soni:
This is the first attempt to improve the errors generated during HTML site validation. A style attribute is removed from the website's page.


chlorenz:
Hi @dcs-soni 
It seems you have removed the `type` attribute from the wrong tag.

PR #36, opened 2021-03-11
Add the nearest-projection OpenFOAM example

davidscn:
Hopefully added it in the right way. Related to precice/tutorials#159.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -183,6 +183,7 @@ subprojects:
   - imported/tutorials/elastic-tube-3d
   - imported/tutorials/flow-over-heated-plate
   - imported/tutorials/flow-over-heated-plate-steady-state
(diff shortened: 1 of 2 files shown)


MakisH (changes requested):
The sidebar part is fine, but we still need to add this in the project config: https://github.com/precice/precice.github.io/blob/master/_config.yml#L184

Incidentally, we don't have documentation about this at the moment. I opened an issue for now https://github.com/precice/precice.github.io/issues/37

davidscn:
Done :ok_hand:

MakisH (approved):

davidscn:
Waiting with the merge here until we merge precice/tutorials#159. Otherwise the page will be not found since it was not part of the initial template.

PR #41, opened 2021-04-12
Update su2-adapter configuration

davidscn:


MakisH (approved):
Looks nice! I did not tried building it, but at least the alert boxes should work.

We should also remove most content from the adapter README.md and forward here: https://github.com/precice/su2-adapter/blob/master/README.md

PR #53, opened 2021-04-21
Add testimonial of SFB1313 (DuMuX)

ajaust:
I added a testimonial describing part of the work done in the SFB1313. I focused on the coupling of free and porous-medium flow using DuMuX and preCICE.


uekerman (approved):
Great, thanks :+1:

PR #55, opened 2021-04-22
Add actions to lint markdown files

MakisH:
Similarly to https://github.com/precice/openfoam-adapter/pull/169 and https://github.com/precice/tutorials/pull/206, this adds a GitHub action to lint markdown files.

Apart from the files in `.github` and two `.markdown*.json`, the rest are just fixing linting issues.


MakisH:
I removed the action to check links for now, as it will need some work to configure it. Other than that, ready for review.

pages/docs/configuration/configuration-basics.md
```diff
@@ -66,37 +71,45 @@ Each solver that participates in the coupled simulation needs a participant defi
...
+
 ```xml
 <mapping:nearest-neighbor direction="read" from="MyMesh2" to="MyMesh1" constraint="consistent"/> 
 ```
-`nearest-neighbor` means that the nearest-neighbor mapping method is used to map data from `MyMesh1` to `MyMesh2`. 
 
-Read more about the [mapping configuration](configuration-mapping.htmln).
+`nearest-neighbor` means that the nearest-neighbor mapping method is used to map data from `MyMesh1` to `MyMesh2`.
 
+Read more about the [mapping configuration](configuration-mapping.htmln).
```
uekerman:
```suggestion
Read more about the [mapping configuration](configuration-mapping.html).
```

pages/docs/installation/installation-overview.md
```diff
@@ -23,15 +23,15 @@ To find the right method for you, follow this decision graph, or simply read on!
 
 ### Operating systems
 
-**Linux**
+#### Linux
```
uekerman: How do such changes interfere with the TOC?

Style-wise `**...**` sometimes just looks better.

uekerman (approved):
Quickly browsed over it. Looks good overall. 
I have not built and tested locally.

pages/docs/installation/installation-overview.md
MakisH: They will appear in the TOC as level 4 headings. My motivation was that this was triggering the linter, as it is essentially a heading of hard-coded style. If this is an intended styling workaround, I can add exceptions, but we should fix it. Shall I just do this?

Here is how it looks like:

Screenshot_2021-04-23 Installing preCICE preCICE - The Coupling Library

MakisH: I suppressed this for now, reverting the previous version.

PR #60, opened 2021-05-31
Update dev-overview.md

JulianSchl:


MakisH:
Thank you for contributing, @JulianSchl! Does the current link not work for you? There was some issue with the SSL certificate a few days ago, but I think this was fixed already.

I am asking because this may be hiding larger issues or need larger updates.

JulianSchl:
Hi @MakisH,
trying to reach https://www.precice.org with Google Chrome, I am getting an NET::ERR_CERT_COMMON_NAME_INVALID

uekerman:
The "www" link neither works on chromium.

davidscn:
Still, dropping the `www` works, so any concerns before merging?

MakisH (approved):
I think we can already integrate this, there is no need to invest more work on making the doxygen page redirect from the `www.` to the non-`www.`. Unless @fsimonis has a quick fix in mind.

fsimonis:
Fixed a few more instances of this issue. Strange that we missed this PR for such a long time.

@JulianSchl Thanks!

PR #68, opened 2021-07-02
Add testimonial for volume coupling using OpenDiHu

davidscn:
@maierbn I cannot assign you as reviewer in the sidebar, but please leave a review.


uekerman:
Thanks for adding!
If time allows: If you aligned the two simulation results vertically, it would fit much better.

PR #71, opened 2021-07-14
Add note on parallel coupling for Aitken under-relaxation

BenjaminRodenberg:
* Related to https://github.com/precice/precice/pull/1042#pullrequestreview-706042347
* Reference still missing


uekerman (approved):
Thanks! Please merge directly once you added the literature.

PR #72, opened 2021-07-25
Add section on preCICE installation on MPCDF Cobra cluster

IshaanDesai:


IshaanDesai:
There is a problem linking the Boost installation to the preCICE installation and this is currently being resolved.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+
+#### Installing dependencies
+
+##### Eigen3
+
+The Eigen repository can be transferred to the cluster and extracted. The path of the extracted folder is set in `~/.bashrc`.
+
+##### Boost
+
+Boost installation needs to be done with the GCC compiler. By default the Intel compilers will be loaded in the session. After the Boost source is downloaded and transferred to the cluster, the following commands are run to install Boost:
```
MakisH: Please switch to a more active/casual language style, see https://precice.org/docs-meta-overview.html#language--style

What helps to find the right level it to think as is you are explaining the content to a close colleague/friend over coffee.

For example, I would rewrite this part as follows:

```suggestion
Transfer the Eigen sources to the cluster and set the path to your `~/.bashrc`. Read more details in the [installing dependencies](./installation-source-dependencies.html#eigen).

##### Boost

You need to [build Boost from source](https://precice.org/installation-source-dependencies.html#boost) with the GCC compiler. After uploading the Boost sources to the cluster, run the following:
```

Also: why does the user _need_ to build with GCC? Did you have issues with Intel? If a cluster prefers a compiler, we should try to stick to that, unless we document the situation.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+export Eigen3_ROOT=$HOME/eigen-<version>/
+export CPLUS_INCLUDE_PATH=$Eigen3_ROOT:$CPLUS_INCLUDE_PATH
+
+# preCICE
+export PRECICE_ROOT=$HOME/precice
+export LD_LIBRARY_PATH=$PRECICE_ROOT/build:$LD_LIBRARY_PATH
+export LIBRARY_PATH=$PRECICE_ROOT/build:$LIBRARY_PATH
+export CPLUS_INCLUDE_PATH=$PRECICE_ROOT/src:$CPLUS_INCLUDE_PATH
+export PKG_CONFIG_PATH="${PRECICE_ROOT}/build/lib/pkgconfig:${PKG_CONFIG_PATH}"
+export CMAKE_PREFIX_PATH=$HOME/precice/build
```
MakisH: Are all of these needed? Can we refer to https://precice.org/installation-source-preparation.html#installation-prefix instead of duplicating?

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+```bash
+module purge
+module load gcc/9 impi/2019.7 cmake petsc-real anaconda
+module list
+
+export Eigen3_ROOT=$HOME/eigen-3.3.9
+
+rm -rf build
+mkdir -p build && cd build
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=. -DPRECICE_PETScMapping=OFF \
```
MakisH: Why setting the install prefix to be the same as the build directory? I think this could cause some confusion if the user wants to upgrade preCICE later on.

If we already set all the paths to `build/` and `src/`, then we don't need to `make install` or to set the install prefix. So, I would either remove these steps, or set a different install prefix.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+  -DPRECICE_PythonActions=OFF -DMPI_CXX_COMPILER=mpigcc -DPYTHON_EXECUTABLE=$(which python) \
```
MakisH: We are loading `petsc-real` and `anaconda`, but then we are disabling `PRECICE_PETScMapping` and `PRECICE_PythonActions`.

- If there is a reason to disable them, we should also comment on that, but then also remove the dependencies. 
- If there is no reason to disable them, then let's keep them enabled (remove the `-D...`).

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+  -DBoost_NO_BOOST_CMAKE=TRUE -DBoost_NO_SYSTEM_PATHS=TRUE -DBOOST_ROOT:PATHNAME=$BOOST_ROOT -DBoost_LIBRARY_DIRS:FILEPATH=$BOOST_ROOT/lib ..
+make -j20
```
MakisH: Why 20? Sounds a bit arbitrary. How many threads per node can one use on this cluster?

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+
+LIB=$PWD/libprecice.so
+patchelf --set-rpath /usr/lib64:$(patchelf --print-rpath $LIB) $LIB
```
MakisH: What does this do / why do we need it? A comment would be very useful, as this is very uncommon.
I assume it is a better way compared to setting `LD_LIBRARY_PATH`, but then we should also use it for the dependencies, or use `LD_LIBRARY_PATH` also for preCICE (which we already set above - isn't it enough? should we keep that one as well?).

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+
+make test_base
```
MakisH: Is the full test suite expected to fail? A comment would be useful here.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,75 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+make install
+
+```
+
+These commands can also be found as a [build script](https://github.com/IshaanDesai/fusion-core-coupling/blob/master/build-precice.sh).
```
MakisH: Remember to update the script if you change anything here as part of the review process. :-)

MakisH (approved):
Thank you for documenting! This is definitely very useful, but there are a few unclear points that we should polish, first.

pages/docs/installation/installation-special-systems.md
fsimonis: Does the cluster provide ninja? That would make this more general.
Otherwise `make -j$(nproc)` could do the trick.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,35 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+module load gcc/9 impi/2019.7 cmake anaconda boost/1.74
+module list
+
+rm -rf build
+mkdir -p build && cd build
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=$PRECICE_ROOT -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF \
+  -DMPI_CXX_COMPILER=mpigcc -DCMAKE_BUILD_TYPE=Debug ..
+make -j20
+
+LIB=$PRECICE_ROOT/libprecice.so
```
fsimonis: `PRECICE_ROOT` is undefined at this point.

Using `PRECICE_ROOT` was heavily used in the past (internally needed for tests only). It also always pointed to the root of the repository, not at the install directory. So this could lead to confusion with more seasoned preCICE users.

pages/docs/installation/installation-special-systems.md
fsimonis: `/usr/lib64` should be visible to the dynamic linker. If not, then this is a system bug and you should report it to the admins of the cluster.
You can use `-DCMAKE_INSTALL_RPATH=/usr/lib64` if you really need this.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,35 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
 * for running: also put `module load intel` in your jobscript
 * use `network="ib0"` for sockets communication beyond one node
+
+### Max Planck Computing and Data Facility (MPCDF) Cobra cluster
+
+#### Installing dependencies
+
+##### Eigen3
+
+See the [Eigen dependency section](https://precice.org/installation-source-dependencies.html#eigen).
```
fsimonis: You could ask the admins to provide a module for this especially as it is an extremely common library.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,35 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+On the Cobra cluster preCICE is installed from source. Please set the installation prefix paths as shown [in this section](https://precice.org/installation-source-preparation.html#installation-prefix) before installing preCICE. The preCICE repository is either cloned or copied to the cluster, and then the following commands are run:
+
+```bash
+module purge
+module load gcc/9 impi/2019.7 cmake anaconda boost/1.74
+module list
+
+rm -rf build
+mkdir -p build && cd build
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=$PRECICE_ROOT -DPRECICE_PETScMapping=OFF -DPRECICE_PythonActions=OFF \
```
fsimonis: `PRECICE_ROOT` is not defined here.

pages/docs/installation/installation-special-systems.md
```diff
@@ -794,3 +794,35 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+  -DMPI_CXX_COMPILER=mpigcc -DCMAKE_BUILD_TYPE=Debug ..
```
fsimonis: Please mention somewhere that this is a debug build.

pages/docs/installation/installation-special-systems.md
IshaanDesai: `PRECICE_PythonActions` is disabled because even if `anaconda` is loaded, CMake gives the error that Numpy was not found. The `anaconda` module is also removed in a further commit

pages/docs/installation/installation-special-systems.md
IshaanDesai: I asked and there is no requirement for Eigen other than preCICE so it will not be added as a module

pages/docs/installation/installation-special-systems.md
```diff
@@ -792,3 +792,26 @@ export PRECICE_BOOST_ROOT=$HOME/software/boost_1_53_0
...
+
+#### Installing dependencies
+
+##### Eigen3
+
+See the [Eigen dependency section](https://precice.org/installation-source-dependencies.html#eigen) as Eigen is not available as a module on this cluster.
+
+#### Installing preCICE
+
+On the Cobra cluster preCICE is installed from source. Please set the installation prefix paths as shown [in this section](https://precice.org/installation-source-preparation.html#installation-prefix) before installing preCICE. The preCICE repository is either cloned or copied to the cluster, and then the following commands are run:
```
MakisH: Very nice overall, sorry that I comment again on this:
- who is installing preCICE?
- who is either cloning or copying the repository to the cluster?
- who is running the following commands?

I really want to convince you that active voice works better here, but I am no language expert. :sweat_smile: 
Note also that you are mixing passive + active + again passive right now.

```suggestion
On the Cobra cluster, you can easily install preCICE from source. Clone the repository or copy the code to the cluster, set the installation prefix paths as shown [in this section](https://precice.org/installation-source-preparation.html#installation-prefix), and then run the following commands:
```

It would be even better to also comment on what these commands do / why these commands / why these versions. But we don't need too much here.

IshaanDesai: Regarding the mix-up of active and passive voice I concede that its just how I think and hence how I write it, but a consistent voice would be better. As for who the installation guide addresses, I personally find it more elegant to write statement which explain things as they are, rather than addressing the user at all times. In my opinion this is again a matter of personal choice, I go with your choice here.

> It would be even better to also comment on what these commands do / why these commands / why these versions. But we don't need too much here.

The versions of the modules are used just because they work and because only a few versions of all software packages are available. I think it would be difficult to *justify* versions.

MakisH: Just to clarify: this is not just my choice, but a decision we took a few months ago, documented here: https://precice.org/docs-meta-overview.html#language--style

There are several good reasons we decided for the active voice and I will be more than happy to discuss this in person, in the first opportunity. :-)

IshaanDesai: Oh I did not know that we have decided on a style! Thanks for pointing me to it, then I stick to it :smile:

PR #75, opened 2021-07-26
Replace warning mechanism in Calculix-SuperMUC docs

ajaust:
This replaces the :warning: tag with new warning mechanism of the Jekyll-based homepage.


ajaust:
This was the only page I could find an old `:warning:` tag. The other pages seem to be clean.

davidscn:
Thanks, we have even more of these: cf https://precice.org/docs-meta-cheatsheet.html

ajaust:
Great! That is good to know.

PR #76, opened 2021-07-26
Add instructions for HLRS HAWK

davidscn:
Installation works fine by now, still testing some jobs before merging here.

### DIFF
--- a/pages/docs/installation/installation-special-systems.md
+++ b/pages/docs/installation/installation-special-systems.md
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
 
 ## Active systems
 
+### HAWK (HPE Apollo/AMD, Stuttgart)
+
+#### Building
+
+
+The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard MPE):
+(1) [Download Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page) and copy it to HAWK. Afterwards export the `EIGEN3_ROOT`, e.g.,
+
+```bash
+export EIGEN3_ROOT="$HOME/precice/eigen"
+```
+
+(2) Load available modules:
+
+```bash
+module load cmake boost petsc/<VERSION>-int32-shared
+```
+
+(3) Build preCICE. The library and include path for PETSc need to be defined explicitly (for an unknown reason):
+
+```bash
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
+
+make install -j 16
+```
+{% include important.html content="Running the tests for the installation will fail, even though the installation is fine, since the tests require MPI facilities which need to be enabled explicitly." %}
+
+#### Running on a single node
+
+Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise the mpi jobs are executed on the same cores, which will slow down the whole simulation significantly. In order to run the a coupled simulation on a single node, the following command can be used:
+
+```bash
+mpirun -np 4 omplace -nt 1 ./exec1 args &
+mpirun -np 4 omplace -b 4 -nt 1 ./exec2 args2
+```
+
+The `nt` argument specifies the number of threads each rank uses. Since we don't want to use multi-threading, we select just a single thread per core. The argument option `-b` specifies the starting CPU number for the effective CPU list, so that we shift the starting number of CPU list in the second participant by the cores employed for the first participant. There are further options to specify the hardware. Have a look at `man omplace` for more information.
+
+#### Notes on deal.II
+
+A variety of dependencies build on the library `METIS`, which requires an installation of `LAPACK`. In order to use `LAPACK` you can load the module `libflame`. Assuming that the preCICE modules above are loaded, you can install deal.II as usual (metis will already be loaded due to a dependency of PETSc). Additional dependecies such as TRILINOS are available through the package manager and can be loaded as well.
+
+#### Notes on OpenFOAM
+
+OpenFOAM is available on the system. You may want to call `module avail openfoam` for a complete overview.
+
 ### SuperMUC-NG (Lenovo/Intel, Munich)
 
 Login: [LRZ page](https://doku.lrz.de/display/PUBLIC/Access+and+Login+to+SuperMUC-NG)
--- a/pages/docs/installation/installation-special-systems.md
+++ b/pages/docs/installation/installation-special-systems.md
@@ -20,7 +20,7 @@ The instructions may still be valuable for unlisted systems.
 
 #### Building
 
-The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard MPE):
+The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard [HPE MPI](https://kb.hlrs.de/platforms/index.php/MPI(Hawk)) implementation):
 
 (1) [Download Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page) and copy it to HAWK. Afterwards export the `EIGEN3_ROOT`, e.g.,
 
@@ -34,7 +34,7 @@ export EIGEN3_ROOT="$HOME/precice/eigen"
 module load cmake boost petsc/<VERSION>-int32-shared
 ```
 
-(3) Build preCICE. The library and include path for PETSc need to be defined explicitly (for an unknown reason):
+(3) Build preCICE. For PETSc, the library and include path (as opposed to the directory and/or architecture, as usual) need to be defined explicitly:
 
 ```bash
 cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
@@ -42,18 +42,20 @@ cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my
 make install -j 16
 ```
 
-{% include important.html content="Running the tests for the installation will fail, even though the installation is fine, since the tests require MPI facilities which need to be enabled explicitly." %}
+Usually, both variables, `PETSc_LIBRARIES` and `PETSc_INCLUDE_DIRS` are supposed to be found by `cmake`. This might be an issue with our PETSc detection mechanism or an issue with the cluster. If you find a more native way to use the provided PETSc installation, please update this documentation. The PETSc module, where this issue occurred, was `petsc/3.12.2-int32-shared`.
+
+{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail, although the installation is fine, since the tests require MPI features that are disabled by default (Spawn Capable MPI runs). If you know how to enable this feature, please edit the page here." %}
 
 #### Running on a single node
 
-Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise, the MPI jobs are executed on the same cores, which will slow down the whole simulation significantly. In order to run the a coupled simulation on a single node, use the following:
+Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise, the MPI jobs are executed on the same cores, which will slow down the whole simulation due to migration significantly. In order to run the a coupled simulation on a single node, use the following:
 
 ```bash
 mpirun -np 4 omplace -nt 1 ./exec1 args &
 mpirun -np 4 omplace -b 4 -nt 1 ./exec2 args2
 ```
 
-The `nt` argument specifies the number of threads each rank uses. Since we don't want to use multi-threading, we select just a single thread per core. The argument option `-b` specifies the starting CPU number for the effective CPU list, so that we shift the starting number of CPU list in the second participant by the cores employed for the first participant. There are further options to specify the hardware. Have a look at `man omplace` for more information.
+The `nt` argument specifies the number of threads each rank uses. Since we don't want to use multi-threading, we select just a single thread per core. The argument option `-b` specifies the starting CPU number for the effective CPU list, so that we shift the starting number of CPU list in the second participant by the cores employed for the first participant. There are further options to specify the hardware. Have a look at the `omplace` using `man omplace` or the [hardware pinning](https://kb.hlrs.de/platforms/index.php/Batch_System_PBSPro_(Hawk)#Pinning) documentation for more information.
 
 #### Notes on deal.II
 
--- a/pages/docs/installation/installation-special-systems.md
+++ b/pages/docs/installation/installation-special-systems.md
@@ -39,14 +39,14 @@ module load cmake boost petsc/<VERSION>-int32-shared
 (3) Build preCICE. For PETSc, the library path and include path need to be defined explicitly:
 
 ```bash
-cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF /path/to/precice/source
 
 make install -j 16
 ```
 
 Usually, both variables, `PETSc_LIBRARIES` and `PETSc_INCLUDE_DIRS` are supposed to be found by `cmake`. This detection mechanism fails on Hawk and therefore we have to specify these variables on the command line. The reason for the detection mechanism to fail is unclear. It might be causes by our PETSc detection mechanism or might be an issue with the cluster. If you find a more native way to use the PETSc installation provided on Hawk, please update this documentation. The PETSc module, where this issue occurred, was `petsc/3.12.2-int32-shared`.
 
-{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail on Hawk, although the preCICE installation is fine. The tests require MPI features (Spawn Capable MPI runs) that are disabled on Hawk . If you know how to enable this feature, please edit the page here." %}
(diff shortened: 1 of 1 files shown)


pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
 
 ## Active systems
 
+### HAWK (HPE Apollo/AMD, Stuttgart)
+
+#### Building
+
+The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard MPE):
+(1) [Download Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page) and copy it to HAWK. Afterwards export the `EIGEN3_ROOT`, e.g.,
```
MakisH, line 25:
```suggestion

(1) [Download Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page) and copy it to HAWK. Afterwards export the `EIGEN3_ROOT`, e.g.,
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
...
+export EIGEN3_ROOT="$HOME/precice/eigen"
+```
+
+(2) Load available modules:
+
+```bash
+module load cmake boost petsc/<VERSION>-int32-shared
+```
+
+(3) Build preCICE. The library and include path for PETSc need to be defined explicitly (for an unknown reason):
```
MakisH: We currently also have such a note in the main documentation: https://precice.org/installation-source-dependencies.html#petsc

Do we maybe need to open an issue, then? Maybe our [`FindPETSc.cmake`](https://github.com/precice/precice/blob/develop/cmake/modules/FindPETSc.cmake) needs some update/fix.

I assume it is PETSc version specific, so it would be useful if you added here which PETSc version this refers to.

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
...
+
+```bash
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
+
+make install -j 16
+```
+
+{% include important.html content="Running the tests for the installation will fail, even though the installation is fine, since the tests require MPI facilities which need to be enabled explicitly." %}
```
MakisH: I am not sure what you mean here with "MPI facilities" (is this the right term?). It would be useful to give a short hint/link on how to enable this.

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
 
 ## Active systems
 
+### HAWK (HPE Apollo/AMD, Stuttgart)
+
+#### Building
+
+The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard MPE):
```
MakisH: What is "MPE"? Does this refer to the MPI version?

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
...
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
+
+make install -j 16
+```
+
+{% include important.html content="Running the tests for the installation will fail, even though the installation is fine, since the tests require MPI facilities which need to be enabled explicitly." %}
+
+#### Running on a single node
+
+Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise the mpi jobs are executed on the same cores, which will slow down the whole simulation significantly. In order to run the a coupled simulation on a single node, the following command can be used:
```
MakisH:
```suggestion
Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise, the MPI jobs are executed on the same cores, which will slow down the whole simulation significantly. In order to run the a coupled simulation on a single node, use the following:
```

For the last change: even though this page is a bit inconsistent, we decided at some point to follow a "casual"/"active" language style in the documentation, as if explaining to a colleague over coffee.

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,52 @@ The instructions may still be valuable for unlisted systems.
...
+
+```bash
+mpirun -np 4 omplace -nt 1 ./exec1 args &
+mpirun -np 4 omplace -b 4 -nt 1 ./exec2 args2
```
MakisH, line 57: Are the `-nt` and `-b` specific to an MPI vendor/version? If yes, we should document this (and generally document which version we are using).

MakisH (approved):
Very good, overall! Here are a few points that I imagine could need some clarification.

pages/docs/installation/installation-special-systems.md
davidscn: I already did quite some investigations of our PETSc detection mechanism and was not able to fix it by now..
The note in our documentation is different. I suspect a problem with pkg-config here. In this case, the specified variables are usually supposed to be spotted by cmake/pkg-config. I assume that this is rather cluster specific. The version used here was `3.12.2`.

pages/docs/installation/installation-special-systems.md
davidscn: I don't know it actually. The tests require to enable `MPI spawning` and ofter some trials I gave it up since I don't see any benefit from making the tests run.

pages/docs/installation/installation-special-systems.md
davidscn: Yes, it refers to the HPE MPI implementation: https://kb.hlrs.de/platforms/index.php/MPI(Hawk)

pages/docs/installation/installation-special-systems.md
davidscn: No, I think it is `omplace` specific. I already referred here to the omplace manual.

pages/docs/installation/installation-special-systems.md
MakisH: Then let's add this link to the page! :smile:

pages/docs/installation/installation-special-systems.md
MakisH: Then just add the link to the MPI version documentation above and this will be clear.

pages/docs/installation/installation-special-systems.md
MakisH: Then let's add this `3.12.2` and the `cluser-specific` in the documentation page.

pages/docs/installation/installation-special-systems.md
MakisH: Then maybe write `since the tests require MPI features that are disabled by default`.

pages/docs/installation/installation-special-systems.md
davidscn: I don't like the MPI manual, it's a mess. I added a link to the (very small but still enough) Hawk hardware pinning section.

pages/docs/installation/installation-special-systems.md
davidscn: Done

MakisH (approved):

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
 
 ## Active systems
 
+### HAWK (HPE Apollo/AMD, Stuttgart)
+
+#### Building
+
+The following steps explain how to install preCICE on HAWK with PETSc and MPI (using the system standard [HPE MPI](https://kb.hlrs.de/platforms/index.php/MPI(Hawk)) implementation):
```
ajaust: Why is this important addition "(using the system standard [HPE MPI](https://kb.hlrs.de/platforms/index.php/MPI(Hawk)) implementation)" inside parentheses? There is also one closing paranthesis too many. I would make it a normal part of the sentence. If the sentence gets too long, make it two sentences.
```suggestion
The following steps explain how to install preCICE on HAWK with PETSc and MPI. We use the system's standard [HPE MPI](https://kb.hlrs.de/platforms/index.php/MPI(Hawk) implementation:
```

pages/docs/installation/installation-special-systems.md
ajaust: Why do I need to copy it to Hawk? Can't I download it directly to/on Hawk?

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+(1) [Download Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page) and copy it to HAWK. Afterwards export the `EIGEN3_ROOT`, e.g.,
+
+```bash
+export EIGEN3_ROOT="$HOME/precice/eigen"
+```
+
+(2) Load available modules:
+
+```bash
+module load cmake boost petsc/<VERSION>-int32-shared
```
ajaust, line 34: Is libxml2 preinstalled or why do we not need to install this nor load a module?

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+```
+
+(3) Build preCICE. For PETSc, the library and include path (as opposed to the directory and/or architecture, as usual) need to be defined explicitly:
```
ajaust: I find this additional information confusing. What is usual to me? What is usual to me? I do not get what that information is supposed to tell me.
```suggestion
(3) Build preCICE. For PETSc, the library path and include path need to be defined explicitly:
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+```bash
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
+
+make install -j 16
+```
+
+Usually, both variables, `PETSc_LIBRARIES` and `PETSc_INCLUDE_DIRS` are supposed to be found by `cmake`. This might be an issue with our PETSc detection mechanism or an issue with the cluster. If you find a more native way to use the provided PETSc installation, please update this documentation. The PETSc module, where this issue occurred, was `petsc/3.12.2-int32-shared`.
```
ajaust: I find the explanation a bit vague.
```suggestion
Usually, both variables, `PETSc_LIBRARIES` and `PETSc_INCLUDE_DIRS` are supposed to be found by `cmake`. This detection mechanism fails on Hawk and therefore we have to specify these variables on the command line. The reason for the detection mechanism to fail is unclear. It might be causes by our PETSc detection mechanism or might be an issue with the cluster. If you find a more native way to use the PETSc installation provided on Hawk, please update this documentation. The PETSc module, where this issue occurred, was `petsc/3.12.2-int32-shared`.
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail, although the installation is fine, since the tests require MPI features that are disabled by default (Spawn Capable MPI runs). If you know how to enable this feature, please edit the page here." %}
```
ajaust: I like to be more explicit.
```suggestion
{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail on Hawk, although the preCICE installation is fine. The tests require MPI features (Spawn Capable MPI runs) that are disabled on Hawk . If you know how to enable this feature, please edit the page here." %}
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+#### Running on a single node
+
+Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise, the MPI jobs are executed on the same cores, which will slow down the whole simulation due to migration significantly. In order to run the a coupled simulation on a single node, use the following:
```
ajaust:
```suggestion
Simulations on a single node are possible, but you explicitly need to specify the hardware. Otherwise, the MPI jobs are executed on the same cores, which will slow down the whole simulation due to migration significantly. In order to run the a coupled simulation on a single node with 8 ranks, use the following command:
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+```bash
+mpirun -np 4 omplace -nt 1 ./exec1 args &
+mpirun -np 4 omplace -b 4 -nt 1 ./exec2 args2
+```
+
+The `nt` argument specifies the number of threads each rank uses. Since we don't want to use multi-threading, we select just a single thread per core. The argument option `-b` specifies the starting CPU number for the effective CPU list, so that we shift the starting number of CPU list in the second participant by the cores employed for the first participant. There are further options to specify the hardware. Have a look at the `omplace` using `man omplace` or the [hardware pinning](https://kb.hlrs.de/platforms/index.php/Batch_System_PBSPro_(Hawk)#Pinning) documentation for more information.
```
ajaust:
```suggestion
The `nt` argument specifies the number of threads each rank uses. Since we don't want to use multi-threading, we select just a single thread per core. The argument option `-b` specifies the starting CPU number for the effective CPU list, so that we shift the starting number of CPU list in the second participant by the cores employed for the first participant. In our case we want to use 4 ranks/cores for each participant. There are further options to specify the hardware. Have a look at `omplace` using `man omplace` or the [hardware pinning](https://kb.hlrs.de/platforms/index.php/Batch_System_PBSPro_(Hawk)#Pinning) documentation for more information.
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+#### Notes on deal.II
+
+A variety of dependencies build on the library `METIS`, which requires an installation of `LAPACK`. In order to use `LAPACK` you can load the module `libflame`. Assuming that the preCICE modules above are loaded, you can install deal.II as usual (metis will already be loaded due to a dependency of PETSc). Additional dependecies such as TRILINOS are available through the package manager and can be loaded as well.
```
ajaust: "A variety of dependencies" What does this mean? A variety of dependencies in general or dependencies of deal.II? I would like this to be more explicit. 

I find the information about `METIS` confusingly formulated. Is `METIS` preinstalled or not? You tell me to load some module for `LAPACK` because it is needed for the installation of `METIS`. On the other hand it sounds like there is `METIS` preinstalled. I just checked and `METIS` is indeed preinstalled. Maybe refer explicitly to some `METIS` module?! 

What does "are available through the package manager" mean? Which package manager? Do you mean the module system?

```suggestion
`METIS` is preinstalled and can be loaded via the module system. In case that the preCICE modules above are loaded, `METIS` will already be loaded as a dependency of PETSc. If you want to compile `METIS` yourself, you can load the module `libflame` to fulfill the `LAPACK` dependency. . In order to use `LAPACK` you can load the module `libflame`. 

Additional dependencies of deal.II, such as `TRILINOS`, are available through module system and can be loaded accordingly. You can get obtain the a full list of preinstalled software on Hawk using the command `module avail`.
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,55 @@ The instructions may still be valuable for unlisted systems.
...
+
+#### Notes on OpenFOAM
+
+OpenFOAM is available on the system. You may want to call `module avail openfoam` for a complete overview.
```
ajaust:
```suggestion
OpenFOAM is available on the system. You may want to call `module avail openfoam` for a complete overview of preinstalled OpenFOAM versions.
```

ajaust (commented):
Thanks for the nice PR. I mainly added comments and suggestions to make the information more explicit. I found some thing could be misinterpreted.

pages/docs/installation/installation-special-systems.md
davidscn: I removed the brackets and put everything into one sentence.

pages/docs/installation/installation-special-systems.md
davidscn: That's a good point, I added a 'note' entry.

pages/docs/installation/installation-special-systems.md
davidscn: I tried to clarify it. `LAPACK` is here a requirement of deal.II, not of `METIS`

davidscn:
@ajaust any further comments before merging?

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,59 @@ The instructions may still be valuable for unlisted systems.
...
+```bash
+module load cmake boost petsc/<VERSION>-int32-shared
+```
+
+{% include note.html content="libxml2 is part of the "-devel" packages, which are loaded by default on the login nodes. The compute nodes run in a diskless mode in order to save RAM. Therefore, make sure to use the login nodes for building purposes." %}
+
+(3) Build preCICE. For PETSc, the library path and include path need to be defined explicitly:
+
+```bash
+cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF .. /path/to/precice/source
```
ajaust: Why does this line contain a relative path `..` and an absolute path `/path/to/precice/source` at the end?
```suggestion
cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX="my/install/prefix" -DPRECICE_PETScMapping=ON -DPETSc_INCLUDE_DIRS="$PETSC_DIR/include" -DPETSc_LIBRARIES="$PETSC_DIR/lib/libpetsc.so" -DPRECICE_PythonActions=OFF /path/to/precice/source
```

pages/docs/installation/installation-special-systems.md
```diff
@@ -16,6 +16,59 @@ The instructions may still be valuable for unlisted systems.
...
+
+make install -j 16
+```
+
+Usually, both variables, `PETSc_LIBRARIES` and `PETSc_INCLUDE_DIRS` are supposed to be found by `cmake`. This detection mechanism fails on Hawk and therefore we have to specify these variables on the command line. The reason for the detection mechanism to fail is unclear. It might be causes by our PETSc detection mechanism or might be an issue with the cluster. If you find a more native way to use the PETSc installation provided on Hawk, please update this documentation. The PETSc module, where this issue occurred, was `petsc/3.12.2-int32-shared`.
+
+{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail on Hawk, although the preCICE installation is fine. The tests require MPI features (Spawn Capable MPI runs) that are disabled on Hawk . If you know how to enable this feature, please edit the page here." %}
```
ajaust: Small typo
```suggestion
{% include important.html content="Some tests (`acceleration`, `mpiports`, `serial` and `parallel`) are known to fail on Hawk, although the preCICE installation is fine. The tests require MPI features (Spawn Capable MPI runs) that are disabled on Hawk. If you know how to enable this feature, please edit the page here." %}
```
Should we suggest to test with the solver dummies here instead?

ajaust (changes requested):
I only found minor things. Looks good otherwise.

pages/docs/installation/installation-special-systems.md
davidscn: The preCICE solverdummies are already part of the test. I guess that the users here can decide on their own how to test this in the best way.

davidscn:
Thanks!

PR #77, opened 2021-07-26
Remove duplicated test run from GH actions

davidscn:
Our tests are currently duplicated in PRs, since the `push` event is triggered as well as the `pull_request` event. We should definitely specify the test branches here so that the `pull_request` event is triggered when opening and syncing PRs and the `push` event is triggered when we actually merge and push to master.

### DIFF
--- a/.github/workflows/check-markdown.yml
+++ b/.github/workflows/check-markdown.yml
@@ -1,5 +1,9 @@
 name: Lint docs
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 jobs:
   check_md:
     runs-on: ubuntu-latest


MakisH (approved):
I was confused whether this was normal/common. Thank you for fixing, looks good to me.

BenjaminRodenberg:
But this also changes when the test is run, isn't it? Previously we ran this test always on any branch. Now it's only for master. Is this intended?

davidscn:
Yes, it only runs if you open up a PR. This is intended.

davidscn:
Merging without PR is a bad idea anyway, isn't it?

BenjaminRodenberg:
> Merging without PR is a bad idea anyway, isn't it?

Yes. But sometimes it still happens or people just open the PR when development is done. And if failing tests pop up at such a late point in time this might be annoying. Here, I don't think it's critical, but generally I think it's good to know about all requirements (=tests) when you are developing and not only when you are submitting the PR.

PR #78, opened 2021-07-28
Improve linking to preCICE in the documentation

davidscn:
I recently added this change in the deal.II adapter and I think it is much more descriptive. The `HINT precice_DIR` is required in case someone installs preCICE at a custom location and he just wants to specify the install prefix on the command line (otherwise, the exact cmake config file location would be necessary. 

Not sure if we want to make the error message part of the code. We could also add an `info` in the wiki. However, I really like this approach, but it is just a suggestion.

### DIFF
--- a/pages/docs/installation/installation-linking.md
+++ b/pages/docs/installation/installation-linking.md
@@ -17,7 +17,16 @@ Linking to preCICE from a CMake project is simple.
 Use `find_package(precice)` and link `precice::precice` to your target:
 
 ```cmake
-find_package(precice REQUIRED CONFIG)
+find_package(precice 2.0 CONFIG HINTS ${precice_DIR})
+
+IF(NOT ${precice_FOUND})
+   MESSAGE(FATAL_ERROR "\n"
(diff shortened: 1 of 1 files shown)


MakisH:
> Not sure if we want to make the error message part of the code.

I think this is actually nice. We educate people that this is a way to handle errors and the error message itself gives good hints.

fsimonis:
The documentation states that you can either use `precice_DIR` to point to the CMake config, or use `CMAKE_PREFIX_PATH` to add an additional prefix for CMake to look for preCICE.

Adding `precice_DIR` as hint to the search procedure is weird because it doesn't mean "prefix".
You can use [`precice_ROOT`](https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT) to specify a prefix for precice in `3.12`, but I didn't add it to the docs as our base-line is `3.10`.

davidscn:
Ok, then let me close it.

PR #80, opened 2021-09-02
Remove 'www' from preCICE domain

davidscn:
... as this triggers some certificate issues. 

```
perl -pi -e 's/http:\/\/www.precice.org/https:\/\/precice.org/g;' `find _site pages imported -type f`
```


pages/docs/couple-your-code/couple-your-code-direct-access.md
```diff
@@ -12,9 +12,10 @@ This concept is required if you want to access received meshes directly. It migh
 ```cpp
     // Allocate a bounding-box vector containing lower and upper bounds per
     // space dimension
-    std::vector<double> boundingBoxes(dim * 2);
+    std::vector<double> boundingBox(dim * 2);
```
uekerman: Seems that this commit does not belong here.

davidscn: Yup

PR #81, opened 2021-09-13
Document currently supported CalculiX version

MakisH:
We currently ask the user to manually replace the version of CalculiX in every documentation instruction, e.g.:

Screenshot from 2021-09-13 10-14-13

This is problematic for two reasons. First of all, many users blindly copy-paste, without reading the context (there is  still a lot of instructions to follow before one can run their first preCICE simulation anyway). But more importantly, we don't clearly state which is the latest supported CalculiX version, leading to some confusion, as the user first needs to go to GitHub.

Now that we are using Jekyll for the documentation, we can configure such documentation-wide versions in the [`_config.yml`](https://github.com/precice/precice.github.io/blob/master/_config.yml). This adds such a configurable version.

@KyleDavisSA please check for any typos or if I need to change this anywhere else. When the default version changes (i.e., the one supported by `master`), please change the value in the `_config.yml`.


MakisH:
@boris-martin would you maybe like to give me a review here? You can then use the same option in the pages you edit.

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
```diff
@@ -12,14 +12,15 @@ After [installing preCICE](installation-overview.html) and [getting the CalculiX
 1. Download and unzip the adapter (e.g. in the `CaluliX` folder):
 
     ```bash
-    wget https://github.com/precice/calculix-adapter/archive/master.zip 
-    unzip master.zip 
-    cd calculix-adapter-master
+    wget https://github.com/precice/calculix-adapter/archive/refs/heads/v{{site.calculix_version}}.zip 
+    unzip v{{site.calculix_version}}.zip 
+    cd calculix-adapter-{{site.calculix_version}}
```
boris-martin, line 17: Unrelated to the PR, but in the long run, would'nt it be better to point to master (while writing in the doc that master is version 2.XX) instead of the branch stored in the website ?
The game then would be to ensure the website and master stay syncrhonized.

But this is just another instance of the debate of how we handle different versions I guess.

pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
```diff
@@ -126,15 +126,15 @@ After building, make sure that you make yaml-cpp discoverable by setting e.g. yo
 
 ### Get the source
 
-Once the libraries are installed, you can finally install Calculix with preCICE adapter (adapt the `VERSION` in the link, see the beginning of the [adapter's README.md](https://github.com/precice/calculix-adapter/blob/master/README.md) to find out which one you need).
+Once the libraries are installed, you can finally install Calculix with preCICE adapter. Note that the adapter version needs to be the same as the CalculiX version (replace `{{site.calculix_version}}` below).
```
boris-martin, line 139: Maybe mention that we support older versions, if necessary ?

pages/docs/adapters/calculix/adapter-calculix-overview.md
```diff
@@ -23,11 +23,12 @@ Do you Want to build on a cluster? Look at our [instructions for SuperMUC](adapt
...
 ## Versions
 
-Please check the [Calculix adapter README](https://github.com/precice/calculix-adapter/blob/master/README.md) for the newest compatible CalculiX version.
+The latest supported CalculiX version is {{site.calculix_version}}. If you already have a copy of the adapter, check the [adapter README](https://github.com/precice/calculix-adapter/blob/master/README.md) for the CalculiX version it was made for.
 
 Adapters for older versions of CalculiX and preCICE are available in various branches. Branches compatible with **preCICE v2.x:**
 
 * master
+* v2.16
 * v2.15_preCICE2.x
```
boris-martin, line 33: This is seriously outdated. But again, not really relevant to this particular PR.
We should at least add develop and 2.17. But the difference between master and develop is not always consistent...

boris-martin (approved):
That's a nice step forward. I'm not a big fan of the current way the repository is structured (one branch per CCX version), but until we change it, this is an improvement in terms of documentation, so let's go with it.

In the near future (incoming release) we should also add a paragraph about the .deb package, and we'll have to be careful to ensure that the calculix_version of the website matches the current deb package.

But in the meantime, it seems good enough to merge this.

(I assume the syntax for variables is correct as you have more experience than me, but I don't know if there's a way to check this without compiling the website)

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
MakisH: Actually, we should make adapter releases and point to them.

pages/docs/adapters/calculix/adapter-calculix-overview.md
MakisH: Then let's not treat it in this PR, but in follow-up PRs. Here we should have the main CalculiX version we support consistent across the documentation.

I understand you would add v2.17 and develop here? Then v2.17 should be the mainly supported one, right?

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
MakisH: Let's update this in a separate PR, after we have the first release.

pages/docs/adapters/calculix/adapter-calculix-overview.md
MakisH: Let's update in a separate PR.

pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
MakisH: I think this is a bigger discussion and I would like to defer it to the point we have a release strategy.

PR #89, opened 2021-09-24
Add note on CentOS 6

IshaanDesai:
Adding a note regarding CentOS 6 based on recent investigations.

### DIFF
--- a/pages/docs/installation/installation-source-dependencies.md
+++ b/pages/docs/installation/installation-source-dependencies.md
@@ -397,6 +397,10 @@ This system requires to install some tools in a fixed order.
 
 {% include important.html content="Use `cmake3` instead of `cmake` to configure preCICE!" %}
 
+### CentOS 6
+
+This system **does not** require special installation steps as for CentOS 7 or 8. Only having compatible versions of dependencies is sufficient for installing preCICE.
+
 ### OpenSUSE Leap 15.2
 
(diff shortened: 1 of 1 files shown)


MakisH (commented):
Since [CentOS 6 has reaced End-of-Life](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule), why should we add this note?

IshaanDesai:
CentOS 6 is indeed no longer maintained but recently in a project we encountered the OS on a cluster and the installation of preCICE turned out to be fairly straightforward. The note is added in case there are more users who wish to install preCICE on CentOS 6. Is it not advisable to provide documentation for older unmaintained versions of OSs?

erikscheurer:
In this project, preCICE was installed on a cluster where there also were some things preinstalled. Could it be, that the installation was only this easy because the server was already set up properly?

MakisH:
> Is it not advisable to provide documentation for older unmaintained versions of OSs?

I would consider good practice to eventually cleanup unsupported systems. For one, we help most users find relevant information and we give a more up-to-date impression. On the other hand, we (collectively, as a community) lead users to use up-to-date software. Old information can always be retrieved from old releases/archives, but I would not add old systems after their EoL. :)

IshaanDesai:
I agree with @MakisH that we should not include documentation of systems which have passed end-of-life dates and are no longer actively maintained.

PR #90, opened 2021-09-27
Add explanation for FEM coupling meshes

davidscn:


pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
 title: Dealing with FEM meshes
 permalink: couple-your-code-fem-meshes.html
 keywords: api, adapter, FEM, meshes, elements
-summary: "TODO"
+summary: "There are various options and the best one depends on your application case."
```
uekerman:
```suggestion
summary: "There are various options how to deal with FEM meshes in preCICE and the best one depends on your application."
```

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
 ---
 
-TODO
+## Common choices
+
+Finite-element discretizations come along with different options in order to define a coupling interface. Given a common (here fourth-order) finite element, we have
+
+![FEM coupling mesh](images/docs/couple-your-code-fem-meshes.png)
+
+  1. geometry-based interfaces such as vertices (black, circular contour) at the element corners or the element surface as a whole,
```
uekerman: "or the element surface as a whole" ... I don't understand this part.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
 permalink: couple-your-code-fem-meshes.html
 keywords: api, adapter, FEM, meshes, elements
-summary: "TODO"
+summary: "There are various options and the best one depends on your application case."
 ---
 
-TODO
+## Common choices
+
+Finite-element discretizations come along with different options in order to define a coupling interface. Given a common (here fourth-order) finite element, we have
```
uekerman:
```suggestion
Finite-element discretizations come along with different options in order to define a coupling interface. Given a common (here fourth-order) finite element, we have:
```

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
 title: Dealing with FEM meshes
 permalink: couple-your-code-fem-meshes.html
 keywords: api, adapter, FEM, meshes, elements
-summary: "TODO"
+summary: "There are various options and the best one depends on your application case."
 ---
 
-TODO
+## Common choices
```
uekerman: Is this headline really needed? You could remove this one and promote all below to h2.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
+  2. the finite-element support points (black, filled circles) and
+  3. the quadrature points (blue, filled circles).
+
+### Geometry-based interfaces
+
+Using the geometry in order to define a coupling interface seems like one of the most obvious choices, since the coordinate information are usually easy to access. The major drawback is, however, that the resolution of the coupling interface is independent of the discretization and you might lose information across the coupling interface. Considering our example element above, a coupling mesh based on geometric vertices would consist of four interface nodes, whereas the other choices have considerably more (25) interface nodes. The overall difference between the geometry-based interface and options two and three are related to the polynomial degree of the finite-element support points (assuming a polynomial FE basis for the moment). In case the code you want to couple uses solely linear elements, the resulting number of interface nodes coincide with the vertex based coupling mesh and the element vertices might be a perfect choice for your application case. For an increasing polynomial degree, option one becomes less attractive.
```
uekerman:
```suggestion
Using the geometry in order to define a coupling interface seems like one of the most obvious choices, since the coordinate information is usually easy to access. The major drawback is, however, that the resolution of the coupling interface is independent of the discretization and you might lose information across the coupling interface. Considering our example element above, a coupling mesh based on geometric vertices would consist of four interface nodes, whereas the other choices have considerably more (25) interface nodes. The overall difference between the geometry-based interface and both other options is related to the polynomial degree of the finite-element support points (assuming a polynomial FE basis for the moment). In case the code you want to couple uses solely linear elements, the resulting number of interface nodes coincide with the vertex based coupling mesh and the element vertices might be a perfect choice for your application case. For an increasing polynomial degree, geometry-based interfaces become less attractive.
```

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
+
+In addition to the resolution of the coupling interface, another challenge in this context is given by the required solution transfer between the FE solution space and the geometry. One has to average over an element face or map the solution within the solver to the vertex location.
+
+### Support points
+
+Finite-element support points are based on the finite-element discretization. Not all finite-element discretizations have support points at all, e.g., Legendre finite elements based on a modal decomposition. However, we focus here on polynomial finite elements with support points, which are a common choice in many applications. The major advantage of support points is that the solution values live typically on the support points and the values can be passed easily to preCICE by reading them from the solution vector. Also, increasing the polynomial degree of the discretization increases the number of coupling interface nodes in contrast to the geometry based coupling interface.
```
uekerman:
```suggestion
Finite-element support points are based on the finite-element discretization. Not all finite-element discretizations have support points, e.g., Legendre finite elements based on a modal decomposition have none. However, we focus here on polynomial finite elements with support points, which are a common choice in many applications. The major advantage of support points is that the solution values live typically on the support points and the values can be passed easily to preCICE by reading them from the solution vector. Also, increasing the polynomial degree of the discretization increases the number of coupling interface nodes in contrast to the geometry based coupling interface.
```

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
+
+However, this approach runs into trouble in case the support points live on element faces in a discontinuous solution space (Discountinuous Galerkin approach), since you define interface nodes multiple times. In particular, the duplicate definition becomes problematic if an interpolant is constructed from the coupling mesh in preCICE, which happens for the `from` mesh in a `read-consistent` setup and for the `to` mesh in a `write-conservative` setup.
```
uekerman:
```suggestion
However, this approach runs into trouble in case the support points live on element faces in a discontinuous solution space (discountinuous Galerkin approach), since you define interface nodes multiple times. In particular, the duplicate definition becomes problematic if an interpolant is constructed from the coupling mesh in preCICE, which happens for the `from` mesh in a `read-consistent` setup and for the `to` mesh in a `write-conservative` setup.
```
You could add a link to the mapping configuration here.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
+
+Another issue using higher-order support points might appear in case you run very large cases and strive for the RBF mapping of preCICE. In case the support points are non-equally distributed within the element (shown in our example element above and as it is usually the case in order to preserve a reasonable conditioning of the mass matrix), the RBF mapping becomes ill-conditioned due to the varying distance of interface nodes all over the place.
```
uekerman: You could link to the 2017 RBF paper from Florian here.

uekerman: And this is only the case if the FEM mesh is the input mesh of the RBF mapping.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,35 @@
...
+
+However, building a coupling interface based on quadrature points may lead to a decisive advantage in case the quadrature formula for data writing (assuming consistent data) can be chose independent of the problem setup. In this case, the degree of the quadrature formula can be increased without touching the discretization, i.e., it would be possible to increase the number of interface nodes artificially. If you use higher-order schemes on your finite-element participant, you could compensate the limited convergence order of the mapping algorithms to some extent. Also, you could select quadrature formulas with equally-distributed quadrature points so that no convergence issues appear in case of higher-order quadrature formulas.
```
uekerman: Another advantage is that you don't need to map within your FEM solver after reading data from preCICE. You have the data exactly where you need it.

I am not sure about the sub-sampling advantage you mention. In principle, you can do this for all three options, right?

uekerman (changes requested):
Extent and content good already. I am not sure about a few details.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
davidscn: Added a link to the restrictions for parallel applications.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
davidscn: > I am not sure about the sub-sampling advantage you mention. In principle, you can do this for all three options, right?

But not independent of the discretization. You can always refine your whole mesh/degree of the FE space, but it will increase the computational cost significantly. Here, I can just increase the number of interface nodes independent.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
davidscn: Added both points.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
davidscn: The `face` center was meant. I will clarify.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
uekerman: But you could also just sub-sample (independently of the discretization) for the other two, no?

davidscn: How would you increase the polynomial degree without changing the problem size of your system and how would you change the number of face centers without changing your triangulation? I mean you could in principle create or more refined triangulation, use it exclusively for coupling and create a mapping between the coupling triangulation and the actual triangulation, but that's in my opinion not really an alternative. 

The quadrature points (the quadrature formula) can easily be changed on individual elements and all global data structures remain the same. Do you have anything different in mind?

davidscn:
I tried to neutralize the explanation a bit @uekerman.

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,37 @@
...
+
+## Conclusion
+
+There are various options how to deal with FEM meshes in preCICE and the best choice depends on your application. In many FEM solvers, the desired write data location is typically related to the support points and the desired read data location is typically related to the quadrature points, since the data is computed/required in these locations. Therefore, you might even consider to define different meshes for reading data from and writing data to preCICE.
```
uekerman:
```suggestion
There are various options how to deal with FEM meshes in preCICE and the best choice depends on the solver you want to couple. In many FEM solvers, the desired write data location is typically related to the support points and the desired read data location is typically related to the quadrature points, since the data is computed/required in these locations. Therefore, you might even consider to define different meshes for reading data from and writing data to preCICE.
```

uekerman (approved):

pages/docs/couple-your-code/couple-your-code-fem-meshes.md
```diff
@@ -2,7 +2,37 @@
...
-summary: "TODO"
+summary: "There are various options how to deal with FEM meshes in preCICE and the best one depends on your application."
 ---
 
-TODO
+Finite-element discretizations come along with different options in order to define a coupling interface. Given a common (here fourth-order) finite element, we have:
+
+![FEM coupling mesh](images/docs/couple-your-code-fem-meshes.png)
+
+  1. geometry-based interfaces such as vertices (black, circular contour) at the element corners or the face centers,
```
MakisH, line 12: @DavidSCN Sorry for only commenting after it has been merged, but I only read this now.
The "black, circular contour" confused me a bit. Maybe rename it to "square"? If the terminology has to be like this, then maybe an "i.e." would help.

Overall, this page will be very useful! I am not sure how much I understood, but I am also not coming from the FEM world.

PR #92, opened 2021-10-01
Replace figures by in-line code in solver details

davidscn:
..which looks not that nice, though (see the direct comparison). 
Screenshot from 2021-10-01 09-00-17

TODO:
-<del> [ ] remove equation numbers from text <del/>


chlorenz:
Thanks David, it's odd the equation blocks aren't centered as they should
be cf. e.g.

https://precice.org/configuration-xml-reference.html#residual-relative-convergence-measure

could you please troubleshoot?

Kind regards
Christopher

chlorenz:
Thanks David, I believe the underlying issue is this:

> KaTeX does not support the align environment because LaTeX doesn't support align in math mode. The aligned environment offers the same functionality but in math mode, so use that instead.

https://katex.org/docs/issues.html

Could you please try with aligned instead?

davidscn:
Awesome! Thanks for pointing that out @chlorenz. It looks now very nice. Ready to go from my side.

PR #93, opened 2021-10-09
Add doc for CalculiX adapter with PaStiX

boris-martin:
Since a few months, it has been reported (https://github.com/precice/calculix-adapter/issues/57) that building the CalculiX adapter with PaStiX (which is possible since CCX 2.17) is troublesome. After a long investigation and many attempts, I managed to do it and noted all the ways it could go wrong. This PR thus includes a new page (+ links to it) explaining in detail how to build the adapter with PaStiX.

- Some fixes to the adapter must be merged before this documentation makes sense, as the current adapter code doesn't support PaStiX. This PR has a twin PR on the adapter : https://github.com/precice/calculix-adapter/pull/71 . This PR shouldn't be merged until the adapter is updated.
- I refer to the use of `ldconfig` but I don't really master this tool. Feedback on that last section is welcome !
- This is my first change to the website. Please check that I did this the correct way. (Correct links, sidebar, ...)
It does look good on my local version of the website, so I'm optimistic.

### DIFF
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -307,6 +307,10 @@ entries:
           url: /adapter-calculix-get-adapter.html
           output: web, pdf
 
+        - title: Build the adapter with PaStiX
+          url: /adapter-calculix-pastix-build.html
+          output: web, pdf
+
         - title: Configuration
           url: /adapter-calculix-config.html
           output: web, pdf
--- a/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
@@ -24,7 +24,6 @@ After [installing preCICE](installation-overview.html) and [getting the CalculiX
 4. Build with `make` (e.g. `make -j 4` for parallel).
 5. You should now have a new executable `ccx_preCICE` in the `bin/` folder of the adapter. You may move this file to a path known by your system, or [add this to your `PATH`](https://unix.stackexchange.com/a/26059/36693) (careful when doing this!).
 
-
 ### Building the adapter with PaStiX
 
 Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStix is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. To help you go thorugh this, we provide [detailed instructions on a dedicated page](adapter-calculix-pastix-build.html).
--- a/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
@@ -24,6 +24,11 @@ After [installing preCICE](installation-overview.html) and [getting the CalculiX
 4. Build with `make` (e.g. `make -j 4` for parallel).
 5. You should now have a new executable `ccx_preCICE` in the `bin/` folder of the adapter. You may move this file to a path known by your system, or [add this to your `PATH`](https://unix.stackexchange.com/a/26059/36693) (careful when doing this!).
 
+
+### Building the adapter with PaStiX
+
+Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStix is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. To help you go thorugh this, we provide [detailed instructions on a dedicated page](adapter-calculix-pastix-build.html).
+
 ### Makefile options
 
 The adapter is built using GNU Make. The `Makefile` contains a few variables on top, which need to be adapted to your system:
--- a/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
@@ -26,7 +26,7 @@ After [installing preCICE](installation-overview.html) and [getting the CalculiX
 
 ### Building the adapter with PaStiX
 
-Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStix is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. To help you go thorugh this, we provide [detailed instructions on a dedicated page](adapter-calculix-pastix-build.html).
+Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStiX is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. Check our [detailed instructions on building the adapter with PaStiX](adapter-calculix-pastix-build.html).
 
 ### Makefile options
 
--- a/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
@@ -1,0 +1,123 @@
+---
+title: Building the Calculix adapter with PaStiX (and CalculiX 2.17)
+permalink: adapter-calculix-pastix-build.html
+keywords: adapter, calculix, pastix
+summary: "Since version 2.17, CalculiX can be built with the PaStiX library to increase performance with CUDA. Building the preCICE adapter
+is somehow harder with this version. This page gives a detailed walkthrough to build the modified adapter."
+---
+
+## Overview
+
+To build the preCICE adapter for CalculiX with PaStiX, it is first necessary to build PaStiX. A manual build is necessary for most steps, as specific compilation flags must be used for the dependencies : their standard distribution cannot be used. In particular, PaStiX (and CalculiX) must be compiled with flags to use 8 bytes integer. On this page, we provide a step-by-step build of the adapter, tested on Ubuntu 20.04. (LTS)
+This should work with some modifications on other systems.
+
+## Required packages
+
+Make sur you have a working installation of preCICE. Also run these installation commands :
+
+
+
+## Downloading CalculiX
+
+Build scripts assume that CalculiX' source code is in `/usr/local`. Donwload it, extract it there and add read/write access to the folder.
+
+```
+    cd /usr/local/ && sudo wget http://www.dhondt.de/ccx_2.17.src.tar.bz2
+    sudo bunzip2 ccx_2.17.src.tar.bz2
+    sudo tar -xvf ccx_2.17.src.tar
+    sudo chmod -R a+rw /usr/local/CalculiX
+
+```
+
+This contains the source code of CalculiX, but also scripts useful for building some libraries below with the correct flags.
+
+## Building PaStiX dependencies
+
+It is assumed that all these libraries will be built on the user home folder, `~`. Minor modifications will be required otherwise.
+PaStiX requires OpenBLAS, hwloc, Scotch and PaRSEC. All of these will be built before PaStiX itself.
+
+### Building OpenBLAS
+
+Clone OpenBLAS source code and build it with 8 bytes integers option.
+
+```
+    cd ~ 
+    git clone https://github.com/xianyi/OpenBLAS.git 
+    OpenBLAS ./OpenBLAS_i8
+    cd OpenBLAS_i8 
+    make -j 4 INTERFACE64=1 
+    sudo make install
+
+```
+
+### Building hwloc
+
+This library will be put in a subfolder of the PaStiX folder.
+```
+    mkdir -p ~/PaStiX/ 
+    cd ~/PaStiX/ 
+    wget https://download.open-mpi.org/release/hwloc/v2.1/hwloc-2.1.0.tar.bz2
+    bunzip2 hwloc-2.1.0.tar.bz2 && tar -xf hwloc-2.1.0.tar
+    sudo cp /usr/local/CalculiX/ccx_2.17/src/make_hwloc.sh ~/PaStiX/hwloc-2.1.0/make_hwloc.sh
+    ./configure --prefix=$HOME/PaStiX/hwloc_i8 CC=gcc CXX=g++ &&
+    make -j8
+    make install
+
+
+```
+
+### Building PaRSEC
+
+```
+    cd ~/PaStiX && git clone -b pastix-6.0.2 --single-branch https://bitbucket.org/mfaverge/parsec.git
+    cp /usr/local/CalculiX/ccx_2.17/src/make_parsec.sh ~/PaStiX/parsec
+    /make_parsec.sh
+
+```
+
+### Building Scotch 
+
+```
+    cd ~/PaStiX
+    wget https://gforge.inria.fr/frs/download.php/file/38114/scotch_6.0.8.tar.gz
+    tar -xf scotch_6.0.8.tar.gz
+    cp /usr/local/CalculiX/ccx_2.17/src/make_scotch.sh ~/PaStiX/scotch_6.0.8 &&
+    scotch_6.0.8
+    ./make_scotch.sh
+
+```
+
+## Start here
+
+1. [Get CalculiX and the dependencies](adapter-calculix-get-calculix.html)
+2. [Build the Adapter](adapter-calculix-get-adapter.html)
+3. [Configure and run simulations](adapter-calculix-config.html)
+4. Follow a tutorial:
+   * [Tutorial for CHT with OpenFOAM and CalculiX](https://github.com/precice/precice/wiki/Tutorial-for-CHT-with-OpenFOAM-and-CalculiX): Flow in a shell-and-tube heat exchanger
+   * [Tutorial for FSI with OpenFOAM and CalculiX](https://github.com/precice/precice/wiki/Tutorial-for-FSI-with-OpenFOAM-and-CalculiX): Flow in a channel with an elastic flap, either perpendicular, or parallel to the flow and attached to a cylinder.
+   * [Tutorial on structure-structure coupling](https://github.com/precice/precice/wiki/Tutorial-for-SSI-with-CalculiX): Elastic beam artificially cut into two halves.
+
+Are you encountering an unexpected error? Have a look at our [Troubleshooting](adapter-calculix-troubleshooting.html) page.
+
+Do you Want to build on a cluster? Look at our [instructions for SuperMUC](adapter-calculix-supermuc.html) (outdated).
+
+## Versions
+
+Please check the [Calculix adapter README](https://github.com/precice/calculix-adapter/blob/master/README.md) for the newest compatible CalculiX version.
+
+Adapters for older versions of CalculiX and preCICE are available in various branches. Branches compatible with **preCICE v2.x:**
+
+* master
+* v2.15_preCICE2.x
+
+All other branches are compatible with **preCICE v1.x**.
+
+## History
+
+The adapter was initially developed for conjugate heat transfer (CHT) simulations via preCICE by Lucia Cheung in the scope of her master’s thesis [[1]](https://www5.in.tum.de/pub/Cheung2016_Thesis.pdf) in cooperation with [SimScale](https://www.simscale.com/). For running the adapter for CHT simulations refer to this thesis. The adapter was extended to fluid-structure interaction by Alexander Rusch [[2]](https://www.gacm2017.uni-stuttgart.de/registration/Upload/ExtendedAbstracts/ExtendedAbstract_0138.pdf).
+
+## References
+
+[1] Lucia Cheung Yau. Conjugate heat transfer with the multiphysics coupling library precice. Master’s thesis, Department of Informatics, Technical University of Munich, 2016.
+
+[2] Benjamin Uekermann, Hans-Joachim Bungartz, Lucia Cheung Yau, Gerasimos Chourdakis and Alexander Rusch. Official preCICE Adapters for Standard Open-Source Solvers. In Proceedings of the _7th GACM Colloquium on Computational Mechanics for Young Scientists from Academia_, 2017.
--- a/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
@@ -135,7 +135,7 @@ Once all of these are done, simply run `make lib` in the `ARPACK` folder.
 To build the adapter, clone its repository and checkout the `2.17` branch : 
 
 ```
-    git clone -b 2.17 https://github.com/precice/calculix-adapter
+    git clone -b v2.17 https://github.com/precice/calculix-adapter
     cd calculix-adapter
 ```
 
--- a/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
@@ -13,7 +13,7 @@ This should work with some modifications on other systems.
 
 ## Required packages
 
-Make sur you have a working installation of preCICE. Also run these installation commands, (after a call to `sudo apt update` and `sudo apt upgrade`) :
+Make sure you have a working installation of preCICE. Also run these installation commands, (after a call to `sudo apt update` and `sudo apt upgrade`) :
 `sudo apt install build-essential cmake git gfortran flex bison zlib1g-dev nvidia-cuda-toolkit-gcc libspooles-dev libyaml-cpp-dev`
 
 ## Downloading CalculiX source
--- a/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
@@ -14,8 +14,9 @@ This should work with some modifications on other systems.
 ## Required packages
 
 Make sur you have a working installation of preCICE. Also run these installation commands, (after a call to `sudo apt update` and `sudo apt upgrade`) :
-build-essential cmake git gfortran flex bison zlib1g-dev
+`sudo apt install build-essential cmake git gfortran flex bison zlib1g-dev nvidia-cuda-toolkit-gcc`
 
+nvidia-cuda-toolkit-gcc ? liblapacke-dev ? 64 ?
 
 ## Downloading CalculiX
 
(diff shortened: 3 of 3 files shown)


_data/sidebars/docs_sidebar.yml
```diff
@@ -307,6 +307,10 @@ entries:
           url: /adapter-calculix-get-adapter.html
           output: web, pdf
 
+        - title: Build the adapter with PaStiX
```
MakisH, line 310: We currently have "Get the adapter", as a general page. Now we add "Build the adapter with PaStiX". This "get -vs- build" sounds a bit strange to a new user, although I completely understand the reason.

Just leaving this open for a moment, to think of alternatives.

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
```diff
@@ -24,6 +24,10 @@ After [installing preCICE](installation-overview.html) and [getting the CalculiX
 4. Build with `make` (e.g. `make -j 4` for parallel).
 5. You should now have a new executable `ccx_preCICE` in the `bin/` folder of the adapter. You may move this file to a path known by your system, or [add this to your `PATH`](https://unix.stackexchange.com/a/26059/36693) (careful when doing this!).
 
+### Building the adapter with PaStiX
+
+Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStix is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. To help you go thorugh this, we provide [detailed instructions on a dedicated page](adapter-calculix-pastix-build.html).
```
MakisH: Try using self-descriptive names of links, so that the user does not need to look at the context:

```suggestion
Since version 2.17 of CalculiX, it is possible to link the PaStiX solver for increased performance, mostly through GPUs. Building the adapter with PaStiX is quite tedious, as most dependencies of PaStiX and PaStiX itslef must be built from source. Check our [detailed instructions on building the adapter with PaStiX](adapter-calculix-pastix-build.html).
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
+---
+title: Building the Calculix adapter with PaStiX (and CalculiX 2.17)
```
MakisH, line 2: Is the version really important here? We could have a disclaimer that this works with version 2.17 and later, but I guess the instructions would remain the same for later versions:

```suggestion
title: Building the Calculix adapter with PaStiX
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+permalink: adapter-calculix-pastix-build.html
+keywords: adapter, calculix, pastix
+summary: "Since version 2.17, CalculiX can be built with the PaStiX library to increase performance with CUDA. Building the preCICE adapter
```
MakisH, line 5:
```suggestion
summary: "Since version 2.17, CalculiX can be built with the PaStiX library to increase performance on some GPU-enabled systems. Building the preCICE adapter
```

It is not exactly CUDA that is accelerating, so this sentence read a bit strange for me. But I also understand that this would only work on CUDA (unless PaStiX supports other systems as well).

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+is somehow harder with this version. This page gives a detailed walkthrough to build the modified adapter."
+---
+
+## Overview
+
+To build the preCICE adapter for CalculiX with PaStiX, it is first necessary to build PaStiX. A manual build is necessary for most steps, as specific compilation flags must be used for the dependencies : their standard distribution cannot be used. In particular, PaStiX (and CalculiX) must be compiled with flags to use 8 bytes integer. On this page, we provide a step-by-step build of the adapter, tested on Ubuntu 20.04. (LTS)
```
MakisH: Just some language suggestions:

```suggestion
To build the preCICE adapter for CalculiX with PaStiX, it is first necessary to build PaStiX. A manual build is necessary for most steps, as specific compilation flags must be used for the dependencies: pre-built packages cannot be used. In particular, PaStiX (and CalculiX) must be compiled with flags to use 8-byte integers. On this page, we provide a step-by-step building guide of the adapter, tested on Ubuntu 20.04.
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+This should work with some modifications on other systems.
+
+## Required packages
+
+Make sure you have a working installation of preCICE. Also run these installation commands, (after a call to `sudo apt update` and `sudo apt upgrade`) :
+`sudo apt install build-essential cmake git gfortran flex bison zlib1g-dev nvidia-cuda-toolkit-gcc libspooles-dev libyaml-cpp-dev`
```
MakisH, line 21: I would prefer having any non-trivial code in a block:

```suggestion
```bash
sudo apt install build-essential cmake git gfortran flex bison zlib1g-dev nvidia-cuda-toolkit-gcc libspooles-dev libyaml-cpp-dev
```
```

File: pages/docs/adapters/calculix/adapter-calculix-pastix-build.md [hunk 7]
```diff
@@ -0,0 +1,166 @@
...
+
+## Downloading CalculiX source
+
+Build scripts assume that CalculiX' source code is in `/usr/local`. Donwload it, extract it there and add read/write access to the folder.
```
MakisH: Do we really need to be in a system directory? Why?
Keep in mind that this will not be possible in many HPC clusters or other shared systems.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+    cd /usr/local/ && sudo wget http://www.dhondt.de/ccx_2.17.src.tar.bz2
```
MakisH: If we need a system path, I would prefer first downloading and extracting as a normal user and then moving to the system path.

Maybe also delete the archive afterwards.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+    sudo bunzip2 ccx_2.17.src.tar.bz2
+    sudo tar -xvf ccx_2.17.src.tar
```
MakisH:
```suggestion
    sudo tar -xjvf ccx_2.17.src.tar.bz2
```

The verbatim, `-v`, you could also skip to keep the logs shorter:

```suggestion
    sudo tar -xjf ccx_2.17.src.tar.bz2
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+    sudo chmod -R a+rw /usr/local/CalculiX
```
MakisH: So, using `/usr/local`, but giving permissions to all to read and write? Some comments on the intention would be useful here.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```
+
+This contains the source code of CalculiX, but also scripts useful for building some libraries below with the correct flags.
+
+## Building PaStiX dependencies
+
+It is assumed that all these libraries will be built on the user home folder, `~`. Minor modifications will be required otherwise.
```
MakisH, line 42:
```suggestion
In the following, we assume that all these libraries will be built on the user home folder, `~`. Minor modifications will be required otherwise.
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+PaStiX requires OpenBLAS, hwloc, Scotch and PaRSEC. All of these will be built before PaStiX itself.
+
+### Building OpenBLAS
+
+Clone OpenBLAS source code and build it with 8 bytes integers option.
```
MakisH:
```suggestion
Clone OpenBLAS source code and build it with 8-byte integers.
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+    cd ~ 
+    git clone https://github.com/xianyi/OpenBLAS.git 
+    mv OpenBLAS ./OpenBLAS_i8
+    cd OpenBLAS_i8 
+    make -j 4 INTERFACE64=1 
+    sudo make install
+
```
MakisH, line 56:
```suggestion
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+    git clone https://github.com/Dhondtguido/PaStiX4CalculiX 
+    mv PaStiX4CalculiX pastix_src
+    cd pastix_src
+    rm make_pastix.sh
+    cp /usr/local/CalculiX/ccx_2.17/src/make_pastix.sh .
+    ./make_pastix.sh
+
+```
+
+The github repository contains a `make_pastix.sh` file; be sure to use the one in the CalculiX folder and not this one !
```
MakisH:
```suggestion
The repository of PaStiX contains a `make_pastix.sh` file; be sure to use the one in the CalculiX folder and not the one from PaStiX!
```

or, even better:

```suggestion
{% include warning.html content="The repository of PaStiX contains a `make_pastix.sh` file; be sure to use the one in the CalculiX folder and not the one from PaStiX!" %}
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Troubleshooting
+
+- On some occasions, a Python script called by CMake generates an incorrect Makefile because of errors in regular expressions. (The script being `cmake_modules/morse_cmake/modules/precision_generator/genDependencies.py`) This should be fixed by calling `pip install regex` (which requires installing the `python3-pip` Ubuntu package). You may also need to replace the `import re` line in that script by `import regex as re`, but the necessity seems to fluctuate among different machines.
```
MakisH:
```suggestion
- On some occasions, a Python script called by CMake (`cmake_modules/morse_cmake/modules/precision_generator/genDependencies.py`) generates an incorrect Makefile because of errors in regular expressions. This should be fixed by calling `pip install regex` (which requires installing the `python3-pip` Ubuntu package). You may also need to replace the `import re` line in that script by `import regex as re`, but the necessity seems to fluctuate among different machines.
```

Even better, use a link to the script on GitHub.

(could this be related to Python2 and Python3?)

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+- Some parts of the code require older versions of the GNU compilers. You may have to replace `gcc` by `gcc-7` and similarly for `g++` and `gfortran` in the `make_pastix.sh` script. This requires installing the relevant Ubuntu packages.
+
+## Building ARPACK, a CalculiX dependency
+
+Calculix relies on ARPACK, and when built with PaStiX, we cannot rely on standard distributions of that library, because it doesn't feature 8-bytes integers by default. We need to compile it ourself. The source code can be found [here](https://www.caam.rice.edu/software/ARPACK/), or by running these commands :
```
MakisH:
```suggestion
Calculix relies on ARPACK, and when built with PaStiX, we cannot rely on standard distributions of that library, because it doesn't feature 8-byte integers by default. We need to build the [ARPACK code](https://www.caam.rice.edu/software/ARPACK/) ourselves:
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+cd ~
+wget https://www.caam.rice.edu/software/ARPACK/SRC/arpack96.tar.gz
+wget https://www.caam.rice.edu/software/ARPACK/SRC/patch.tar.gz
+zcat arpack96.tar.gz | tar -xvf -
+zcat patch.tar.gz    | tar -xvf -
+
```
MakisH, line 130:
```suggestion
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+Before building the library, the following modifications are required :
```
MakisH:
```suggestion
Before building the library, the following modifications are required:
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+Once all of these are done, simply run `make lib` in the `ARPACK` folder.
+
+## Building the adapter
+
+To build the adapter, clone its repository and checkout the `2.17` branch :
+
+```bash
+    git clone -b v2.17 https://github.com/precice/calculix-adapter
+    cd calculix-adapter
+```
```
MakisH, line 149: To avoid duplication, we could directly refer to the normal documentation page: https://precice.org/adapter-calculix-get-adapter.html

(mainly to avoid specifying the version)

See also https://github.com/precice/precice.github.io/pull/81

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Fixing the code
+
+Due to some conflicts between CalculiX, PaStiX and the adapter (both CalculiX and PaStiX have a `pastix.h` file, and neither of them is local from the point of view of the adapter), some changes are required in the CalculiX codebase. We provide a script, `pastix_pre_build.sh` that does these changes. Run it.
```
MakisH, line 153:
```suggestion
Due to some conflicts between CalculiX, PaStiX, and the adapter (both CalculiX and PaStiX have a `pastix.h` file, and neither of them is local from the point of view of the adapter), some changes are required in the CalculiX codebase. We provide a script, `pastix_pre_build.sh` that does these changes. Run it.
```

[Oxford comma](https://en.wikipedia.org/wiki/Serial_comma)

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+    ./pastix_pre_build.sh
+```
+
+### Compilation
+
+To build the adapter, use the provided `Makefile_i8_PaStiX` : the regular Makefile would build the adapter without PaStiX. Assuming you followed the previous steps, it should be useable without modifications; otherwise, some paths updates could be required. Run this command in the `calculix-adapter` (and be sure to checkout the `2.17` branch) folder :
```
MakisH: I see quite often a space before colons. Any editor-specific reason?

```suggestion
To build the adapter, use the provided `Makefile_i8_PaStiX`: the regular Makefile would build the adapter without PaStiX. Assuming you followed the previous steps, it should be useable without modifications; otherwise, some path updates could be required. Run this command in the `calculix-adapter` folder (and be sure to checkout the `2.17` branch):
```

(I would again remove the specific version and assume that we are already in v2.17 or higher)

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+    make -f Makefile_i8_PaStiX -j 4 
+```
+
+Once the build is successful, the adapter should be in `./bin/ccx_preCICE`.
+
+### Updating shared libraries
+
+Running the adapter at this point should fail because of a missing shared library : `libparsec.so.2`. A possible fix to this is to copy it in your local library folder and run `ldconfig` :
```
MakisH:
```suggestion
Running the adapter at this point should fail because of a missing shared library: `libparsec.so.2`. A possible fix to this is to copy it in your local library folder and run `ldconfig`:
```

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+    sudo cp ~/PaStiX/parsec_i8/lib/libparsec.so.2 /usr/local/lib
+    sudo ldconfig
```
MakisH: I again don't really like the system-wide paths.

MakisH (changes requested):
Thanks a lot for the detailed documentation! It is mostly very good, I mainly have some objections on using system paths and `sudo`.

_data/sidebars/docs_sidebar.yml
boris-martin: Well in the long run user will have the choice between :
- Debian package
- Build from source (as of now)
- Build with PaStiX

That's quite a lot and a split would be welcome. But the exact way to do it is not obvious.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: CUDA is indeed required. PaStiX can be built without it, but the specific version of PaStiX tailored for CalculiX does require CUDA.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: That's instructions from Guido Dhondt but it should indeed be adapted

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: Probably one more reason to move to a local folder

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: Space before colon is standard in French, I keep forgetting it's not supposed to be done in English :|

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: What's the alternative for shared libraries ? I'm used to paste DLL locally on Windows but AFAIK there's no equivalent on Linux.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
MakisH: oh! I learned something today! :smiley:

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
MakisH: I would build in local directories and use, e.g., pkg-config. Maybe this page helps: https://precice.org/installation-linking.html

_data/sidebars/docs_sidebar.yml
MakisH: If we make the page concise enough, this should also fit in a section of the same page, actually.

boris-martin: I had the feeling it wouldn't fit on one page (or be too "distracting", the user who doesn't want PaStiX would be overwhelmed with technical details) but it's probably subjective. But the Debian package would probably only take 2-3 lines at most indeed.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: So, I investigated the `ldconfig` sutff, and I found two cleaner approaches:

- You can create a file in /etc/ld.so.conf.d/ (e.g. `parsec.conf`) where you put the path of of the lib (`~/PaStiX/parsec_i8/lib/libparsec.so.2 /usr/local/lib`, except that the ~ alias must be replaced by its true value) then run `ldconfig`. So instead of copying the library we add an index to it in the global system
- You can simply `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:path/to/lib` and it works without calling `ldconfig`. But we must make sure it's exported everytime, for instance by adding this line in a `.bashrc` file or something like this. It's probably better since no superuser call is required but I'm not sure it's the cleanest.

I read the page about pkg-config, but this is related to the issue of the makefile with absolute paths (which I've yet to improve), not to the `ldconfig` issue, right ? Not sure why you linked in a comment about `ldconfig`, unless I missed something.

MakisH: It is fine if you just use the `LD_LIBRARY_PATH` option, we anyway use it for other dependencies as well. This is very common and easy, even though not ideal. And yes, one would normally add the change in their `.bashrc`.

MakisH: > I read the page about pkg-config, but this is related to the issue of the makefile with absolute paths (which I've yet to improve), not to the ldconfig issue, right ?

One can use `pkg-config` similarly: you can make a `.pc` file describing the include and link flags, and then make this file discoverable, by extending the `PKG_CONFIG_PATH`. But this again requires changing an environment variable.

> Not sure why you linked in a comment about ldconfig, unless I missed something.

I am not sure I understand this part. My objection was mainly about `sudo`, which is needed if we want the user to install dependencies in system-wide paths.

boris-martin: `LD_LIBRARY_PATH` it is then.

> One can use pkg-config similarly: you can make a .pc file describing the include and link flags, and then make this file discoverable, by extending the PKG_CONFIG_PATH. But this again requires changing an environment variable.

I updated the makefile to use `pkg-config` to find flags for `pastix hwloc parsec` and it works. So I add in the documentation that we must export these 3 paths, right ? (It's already in the Makefile comment)

> I am not sure I understand this part.

What I mean is that I changed two things:
- I update LD_CONFIG_PATH so that at runtime, the parsec `.so` file is found
- I improved the makefile to use `pkg-config` to find correct include / link flags for dependencies of the adapter instead of hardcoding them.

But I don't see how the second point removes a `sudo`.

> My objection was mainly about sudo, which is needed if we want the user to install dependencies in system-wide paths.

With these changes the only `sudo` left are the one for installing dependencies (apt install) and for installing OpenBLAS system wide. I guess I should try to make it possible to install OpenBLAS locally, but for some reasons I had troubles the first time I tried. I'll give it another try !

MakisH: I guess this conversation was evolved for a while and we have lost synch a bit (and potentially me to myself). Let me just summarize my concerns and I believe you will know what exactly to do afterwards:

- Let's ask the user explicitly to install dependencies, instead of running a script that installs them (anyway ok)
- Dependencies from the package manager are better than installing them manually (if we can assume a system). In that case, `sudo` is perfectly fine.
- Additional information on building the dependencies from source as a fallback is welcome (for clusters etc), but optional.
- Only ask the user to build dependencies from source by default if they need some special option or any other modifications.
- Avoid running `sudo make install` or modifying system paths, even if this is `/usr/local`. This is allowed (and this is the purpose of `/usr/local`), but users often mess up their systems if they manually install software on system paths (e.g., by having duplicate installations). Asking them to install to a local prefix gives additional confidence that they can easily fix it later.

boris-martin:
I changed a few things and I think I addressed all comments above, I think we're almost good to merge. @KyleDavisSA wanted to update instructions on CUDA, but I couldn't test this thoroughly as I can't myself run CUDA (VirtualBox doesn't really support it, and my attempts on WSL wasted a lot of time)... Maybe the best would be to get some user feedback on this ?

There was also the question from @MakisH  about how to structure this, but given the complexity of the thing, I still think a dedicated page is the optimal idea.

MakisH:
@boris-martin can you please resolve any conversations you have already addressed?

Unfortunately, I don't have CUDA either.

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
```diff
@@ -0,0 +1,168 @@
...
+`sudo apt install build-essential cmake git gfortran flex bison zlib1g-dev nvidia-cuda-toolkit-gcc libspooles-dev libyaml-cpp-dev`
+
+## Downloading CalculiX source
+
+This guide assumes Calculix's source code is in the user's home folder `/home/user_name`, with the alias `~`. If you don't want to follow this convention, you may have to adapt slightly the instructions below. Donwload can be done on command line:
+
+```bash
+    cd ~ && wget http://www.dhondt.de/ccx_2.17.src.tar.bz2
+    bunzip2 ccx_2.17.src.tar.bz2
+    tar -xvf ccx_2.17.src.tar
```
MakisH, line 34: See also https://github.com/precice/precice.github.io/pull/81 (also for any other hard-coded CalculiX version here)

MakisH (approved):
Looks good! Let's not wait for someone of us to test it, I think it can be complicated to get someone with CUDA already installed. It is already much better than nothing and most probably already good enough. If a user ends up in this niche corner of the website, they will also complain. :wink:

(you could actually add a note asking for feedback, but keep in mind they are [currently being rewritten](https://github.com/precice/precice.github.io/pull/113) and should remember updating it later)

pages/docs/adapters/calculix/adapter-calculix-pastix-build.md
boris-martin: Yes, I agreee with both suggestions (notes & CCX version).
Until the notes update is merged I should probably add it as plain text then?

Also PaStiX is only available since 2.17, so using the website variable seems inappropriate as long as it refers to 2.16

MakisH: The notes update has been merged now.

MakisH (approved):
Check a minor warning in the "files changed" view (you need a newline before the code block in "Compilation"). Other than that, good to go!

Thanks for all the work!

MakisH (approved):

PR #105, opened 2021-11-30
Add ASTE Documentation in tooling section

kursatyurt:
ASTE was lack of documentation. 
This PR adds ASTE documentation to the preCICE website tooling part.


davidscn:
Ah very nice, thanks!

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Artificial Solver Testing Environment (ASTE)
+permalink: tooling-aste.html
+keywords: tooling, aste
```
uekerman, line 4: Would be great to add a summary here as well (compare similar pages).

davidscn:
Could you address the failing CI tests in a first go? (Just have a look at the 'Files Changed' view in this PR).

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,145 @@
...
+| Flag      | Explanation                                            |
+| --------- | ------------------------------------------------------ |
+| -v        | Enables verbose output                                 |
+| -c        | To specify preCICE config file (default="precice.xml") |
+| -p        | Participant name either A or B                         |
+| --runName | Name of run (default="")                               |
+| --data    | Name of data array to be mapped                        |
+| --mesh    | Mesh prefix                                            |
+| --output  | Output filename (default=output)                       |
+| --vector  | A bool switch to specify vector data (default=False)   |
```
davidscn: I guess we need to converge (and fix) on precice/aste#56 before documenting this here.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
+---
+title: Artificial Solver Testing Environment (ASTE)
+permalink: tooling-aste.html
+keywords: tooling, aste
+---
+
+## Motivation
+
+ASTE is a tool which can be used to understand the dynamics of preCICE, how the data is goes in and out. It consist of C++ and Python modules.
```
davidscn:
```suggestion
ASTE is a tool which can be used to understand the dynamics of preCICE, how the data goes in and out. It consist of C++ and Python modules.
```
I don't like this motivation. Content-wise, we should mention that aste is:
- a lightweight wrapper around preCICE  
- allows to investigate different mapping setups in terms of runtime and accuracy
- allows to investigate different mapping data
- allows to replace individual participants by aste for development purposes

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Installation
+
+Please make sure that dependencies are installed.
```
davidscn:
```suggestion
ASTE relies on the folowing requirements
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+### C++ modules
+
+#### Required
+
+- preCICE
+- MPI
+- CMake
```
davidscn: Is MPI really a requirement or is it just needed for parallel execution? We should also add a link to the preCICE installation instructions. Also mention, that `cmake` and `mpi` are requirements of preCICE itself and most probably already satisfied.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+##### Optional
+
+- METIS
```
davidscn, line 45: Please also add some explanation what METIS is and how it is optional. Which feature can be obtained when building ASTE with METIS?

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Python modules
+
+#### Required
+
+- NumPy
+- vtk (Visualization Toolkit)
```
davidscn, line 60: We should add some 'how to install' instructions here as well. Usually, python dependencies are given in a requirements file in the respective repository.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+```bash
+mkdir build
+cd build
+cmake ..
+make
+```
+
+If precice is not installed in `$PRECICE_ROOT/build` do `cmake -DCMAKE_LIBRARY_PATH=$PRECICE_INSTALL_DIR ..` with the correct installation directory.
+
+After building, add the `build/` directory to your `PATH`.
```
davidscn: We should add an install target to aste and update the documentation here.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+```bash
+preciceMap -v -p A --mesh fine_mesh/fine_mesh --data "x + y" 
+```
+
+## Python modules
+
+### vtk_calculator.py
+
+Reads a mesh as either `.vtk` or `.vtu` and evaluates a function given by `-f` on it. Using the `-d` flag, it can compute the difference between the mesh values and the values of the analytical solution (usually applied after a mapping).
```
davidscn: The flags are outdated here and in the following section.

davidscn (changes requested):

pages/docs/tooling/tooling-aste.md
kursatyurt: MPI is a requirement for ASTE but not for preCICE. IIRC, there is an option in preCICE to not using MPI just using TCP/IP communication

MakisH: If MPI is only an optional requirement, then you can also write for which feature it is needed. In any case, it should be listed, even if it is also a dependency of preCICE.

You may also want to document some minimum version requirements (not for preCICE, but for the rest). If ASTE follows the same minimum versions as preCICE (including CMake and C++ compiler), then you can also simply write that and refer to the [dependencies](https://precice.org/installation-source-dependencies.html) page of preCICE.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+- preCICE
+- MPI
+- CMake
+
+##### Optional
+
+- METIS
+
+### Python modules
```
MakisH, line 55: Why separating between C++ and Python dependencies? This should not matter for the user.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+### Building
+
+```bash
+mkdir build
+cd build
+cmake ..
+make
+```
+
+If precice is not installed in `$PRECICE_ROOT/build` do `cmake -DCMAKE_LIBRARY_PATH=$PRECICE_INSTALL_DIR ..` with the correct installation directory.
```
MakisH: Generally, we tried to stop using the `$PRECICE_ROOT` variable with preCICE v2. I suggest removing it and saying something like "if preCICE is not globally installed on your system, use cmake ...". You can also link to https://precice.org/installation-linking.html

Keep in mind that many users will overlook that `$PRECICE_INSTALL_DIR` is an example. I prefer using a more obvious placeholder, such as `/path/to/precice/build/`.

Please also always write preCICE with this capitalization (unless, of course, it is code).

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+After building, add the `build/` directory to your `PATH`.
+
+### on SuperMUC
+
+Initialize environment, e.g., put in your `.bashrc`
```
MakisH:
```suggestion
Initialize the environment, e.g., put in your `~/.bashrc`:
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+# METIS for aste
+module load metis
+export METIS_DIR=$METIS_BASE # so that cmake finds metis
+export CPLUS_INCLUDE_PATH="$METIS_BASE/include:$CPLUS_INCLUDE_PATH"
+```
+
+and execute cmake like `cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ ..` and `make`.
```
MakisH: Are the `-DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++` really needed? Why? Does METIS set some strange default?

Also, write "CMake" when referring to the name.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+After building, add the `build/` directory to your `PATH`.
+
+### on SuperMUC
+
+Initialize environment, e.g., put in your `.bashrc`
+
+```
+# METIS for aste
+module load metis
+export METIS_DIR=$METIS_BASE # so that cmake finds metis
```
MakisH: This comment is not really needed:
```suggestion
export METIS_DIR=$METIS_BASE
```
I try to avoid in-code comments in such short documentation scripts, to make the commands look shorter/simpler and easier get an overview of what is going on. These are small differences, but we can still aim for them. :-)

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+export CPLUS_INCLUDE_PATH="$METIS_BASE/include:$CPLUS_INCLUDE_PATH"
+```
+
+and execute cmake like `cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ ..` and `make`.
+
+## Usage
+
+## C++ modules
```
MakisH: Same comment about languages: does the user need to know if something is C++ or Python? The most important part should be which are the tools that they need for what (and which tools they will need more often).

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+### preciceMap
+
+The main tool used for mapping.
+
+| Flag      | Explanation                                            |
+| --------- | ------------------------------------------------------ |
+| -v        | Enables verbose output                                 |
+| -c        | To specify preCICE config file (default="precice.xml") |
```
MakisH: We should really change the default to `precice-config.xml`, as all of our tutorials use that name.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| -p        | Participant name either A or B                         |
```
MakisH:
```suggestion
| -p        | Participant name (A or B)                              |
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| --runName | Name of run (default="")                               |
+| --data    | Name of data array to be mapped                        |
+| --mesh    | Mesh prefix                                            |
```
MakisH: Any defaults here?

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+The main tool used for mapping.
+
+| Flag      | Explanation                                            |
+| --------- | ------------------------------------------------------ |
+| -v        | Enables verbose output                                 |
+| -c        | To specify preCICE config file (default="precice.xml") |
+| -p        | Participant name either A or B                         |
+| --runName | Name of run (default="")                               |
+| --data    | Name of data array to be mapped                        |
```
MakisH: Is this actually a name or an expression? In the example, I see `--data "x + y"`

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| --mesh    | Mesh prefix                                            |
+| --output  | Output filename (default=output)                       |
+| --vector  | A bool switch to specify vector data (default=False)   |
+
+Sample usage is
```
MakisH: Instead of this short generic statement, try adding a few more details, telling a story:

"For example, to map ..., you can use:"

You can then link to the ASTE tutorial for more.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+
+### vtk_calculator.py
+
+Reads a mesh as either `.vtk` or `.vtu` and evaluates a function given by `-f` on it. Using the `-d` flag, it can compute the difference between the mesh values and the values of the analytical solution (usually applied after a mapping).
+
+| Flag | Explanation                                                                         |
+| ---- | ----------------------------------------------------------------------------------- |
+| -o   | Output meshname if it is not given data appended to input mesh                      |
+| -t   | Tag for the data created (default="MyScalar")                                       |
+| -it  | The tag for input data. Used in diff mode. If not given tag is used.                |
```
MakisH:
```suggestion
| -it  | The tag for input data. Used in diff mode. If not given, the tag from -t is used.                |
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| -l   | Logging level (default="INFO")                                                      |
+| -d   | Calculate the difference to present data.                                           |
```
MakisH: What do you mean by "present data"? Clarify / rephrase.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+### join_mesh.py
+
+Reads a partitioned mesh from a given prefix (looking for `<prefix>_<#filerank>.vtu)`) and saves it to a single `.vtk` or `.vtu` file.
+Using the `-r` flag, it also recovers the connectivity information from a mesh. Notice that for recovery, partitions should contain `GlobalIDs` data.
+
+| Flag | Explanation                                                                                           |
+| ---- | ----------------------------------------------------------------------------------------------------- |
+| -o   | The output mesh. Can be VTK or VTU format. If it is not given `<inputmesh>_joined.vtk` will be used." |
```
MakisH:
```suggestion
| -o   | The output mesh. Can be VTK or VTU format. If it is not given `<inputmesh>_joined.vtk` will be used.  |
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| -r   | The path to the recovery file to fully recover it's state.                                            |
```
MakisH:
```suggestion
| -r   | The path to the recovery file to fully recover its state.                                             |
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| -n   | The number of parts to read from the input mesh. By default the entire mesh is read.                  |
```
MakisH:
```suggestion
| -n   | The number of parts to read from the input mesh. By default, the entire mesh is read.                 |
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,141 @@
...
+| -l   | Logging level (default="INFO")                                                                        |
+
+```bash
+./join_mesh.py partitoned_mesh_directory/partitioned_mesh -r partitioned_directory -o rejoined_mesh.vtk
+```
+
+### libMetisAPI
+
+This is a small C++ wrapper around METIS. It is only required if `partition_mesh.py` should use a topological algorithm.
```
MakisH: Since all previous scripts are files that the user can execute, would maybe this fit better as a short line under `partition_mesh.py`?

MakisH (commented):
Thanks a lot for adding documentation! Since this is good already, I went ahead and commented also on minor details.

pages/docs/tooling/tooling-aste.md
kursatyurt: C++ module `preciceMap` can be used without python modules, python modules just provide additional features.

pages/docs/tooling/tooling-aste.md
kursatyurt: I did not test ASTE on SuperMUC. This is from the README file, probably default compiler on SuperMUC is intel compiler and we need to specify the compiler here?

pages/docs/tooling/tooling-aste.md
kursatyurt: I changed that one. Thanks for information

pages/docs/tooling/tooling-aste.md
kursatyurt: There is no default here

pages/docs/tooling/tooling-aste.md
kursatyurt: Definitely I moved to there

pages/docs/tooling/tooling-aste.md
kursatyurt: `Calculates the difference between --diffdata and given function. ` probably more clear

pages/docs/tooling/tooling-aste.md
davidscn: Yes, the 'default' compiler on SuperMUC is an intel compiler. However, it is the responsibility of the SuperMUC user to load compilers (also the defaults). Once the compiler is loaded, cmake should pick up the correct compiler. If this is not the case, it is more of a user problem. This hint might also stem from earlier times of preCICE. In general, the intel compiler should work as well.

pages/docs/tooling/tooling-aste.md
davidscn: Is this restriction compatible with the replay mode, where the participant name is arbitrary?

pages/docs/tooling/tooling-aste.md
kursatyurt: Then I remove compiler directions from the documentation?

pages/docs/tooling/tooling-aste.md
kursatyurt: I changed this part as suggested

pages/docs/tooling/tooling-aste.md
kursatyurt: `preciceMap` always maps from participant A to participant B.

pages/docs/tooling/tooling-aste.md
davidscn: Yes. We should probably remove the whole supermuc section as we have collect all cluster documentation on this page https://precice.org/installation-special-systems.html

pages/docs/tooling/tooling-aste.md
davidscn: >  MPI is a requirement for ASTE but not for preCICE. IIRC, there is an option in preCICE to not using MPI just using TCP/IP communication

Note that the socket communication is used for the intersolver communication and in case your code (i.e. in this case aste) runs in parallel, mpi is a mandatory requirement for preCICE as well.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,161 @@
...
+### preciceMap
+
+The main tool used for mapping. It is a wrapper around preCICE interface. It can be run in serial or parallel. Mapped data can be a vector or scalar.
+
+| Flag      | Explanation                                                   |
+| --------- | ------------------------------------------------------------- |
+| -v        | Enables verbose output                                        |
+| -c        | To specify preCICE config file (default="precice-config.xml") |
+| -p        | Participant name (A or B)                                     |
+| --runName | Name of run (default="")                                      |
```
davidscn:
```suggestion
```
removed in ASTE

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,161 @@
...
+| --vector  | A bool switch to specify vector data (default=False)          |
+
+For example, mapping the data "x + y" from a mesh named fine_mesh in directory fine_mesh to anotherMesh and save into "mappedData" and "mappedMesh":
+
+```bash
+preciceMap -v -p A --mesh fine_mesh/fine_mesh --data "x + y" 
+mpirun -n 4 -v -p B --mesh anotherMesh --data "mappedData" --output mappedMesh
```
davidscn, line 110: We should also add a note that the `--mesh` argument here is different from the mesh name in the precice configuration file. Maybe refer to the tutorial xml file for references.

davidscn:
As a follow up, we should delete the content of `aste`'s readme and maybe create a similar nice image (but updated) of the workflow.

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,162 @@
...
+## Installation
+
+ASTE relies on following requirements
+
+### C++ modules
+
+#### Required dependencies for C++
+
+- C++ compiler (with support for C++14, e.g. GCC version >= 5)
+- preCICE
```
uekerman:
```suggestion
- preCICE (version >= 2.0)
```

pages/docs/tooling/tooling-aste.md
```diff
@@ -0,0 +1,162 @@
...
+- MPI
+- CMake (version >= 3.1)
+- Boost (version >= 1.65.1)
+- vtk (Visualization ToolKit)
```
uekerman, line 25: I built vtk from source. Is there any better on Ubuntu 20.04?
Could we specify a minimal vtk version?

kursatyurt: Building VTK from source would be pain. 

https://packages.ubuntu.com/search?keywords=vtk
`sudo apt install libvtk7-dev` or preferable `sudo apt install libvtk9-dev` will work.

Specifying a minimal version is a hard choice, this also depends on the VTK files. The latest version can be recommended but not a minimum requirement.

uekerman: > Specifying a minimal version is a hard choice, this also depends on the VTK files. The latest version can be recommended but not a minimum requirement.

Let's add such a statement.

and could we maybe add a brief "For example, on Ubuntu 20.04, you get all dependencies with `sudo apt install ...`"?

kursatyurt:
It is ready to review again after a long break

pages/docs/tooling/tooling-aste.md
```diff
@@ -69,16 +91,17 @@ For demo usage please check [ASTE Tutorial](https://github.com/precice/tutorials
 
 ### preciceMap
 
-The main tool used for mapping. It is a wrapper around preCICE interface. It can be run in serial or parallel. Mapped data can be a vector or scalar.
+The main tool used for mapping. It is a wrapper around the preCICE interface. It can be run in serial or parallel. There are two modes available, namely, mapper mode and replay mode. The first mode is used to assess mapping data from one mesh to another; later, one is mainly developed for diagnostic and adaptor development purposes. In Replay mode, ASTE replicates one participant of preCICE.
```
uekerman:
```suggestion
The main tool used for mapping. It is a wrapper around the preCICE interface. It can be run in serial or parallel. There are two modes available, namely, mapper mode and replay mode. The first mode is used to assess mapping data from one mesh to another; later, one is mainly developed for diagnostic and adapter development purposes. In Replay mode, ASTE replicates one participant of preCICE.
```

davidscn:
I would propose to add aste as a submodule to the repository and move the content here to the [aste docs](https://github.com/precice/aste/tree/develop/docs) similar to the tutorials.

pages/docs/tooling/tooling-aste.md
kursatyurt: Thank you.

kursatyurt:
> I would propose to add aste as a submodule to the repository and move the content here to the [aste docs](https://github.com/precice/aste/tree/develop/docs) similar to the tutorials.

I would prefer other-way around actually since this repository requires additional checks for linting and website build.

MakisH:
> > I would propose to add aste as a submodule to the repository and move the content here to the [aste docs](https://github.com/precice/aste/tree/develop/docs) similar to the tutorials.
> 
> I would prefer other-way around actually since this repository requires additional checks for linting and website build.

You could also add linting steps to the ASTE repository. The scheme that David suggested is the general direction we are going towards: the website renders content that sources from different repositories (tutorials, adapters (OpenFOAM), etc). This allows to have the content also available offline, and it links code and documentation changes in the same PR.

Let's continue the review in the ASTE repository, if that is fine with you. We would still need to update this PR to source the documentation from the ASTE repository.

davidscn:
Superseded by https://github.com/precice/aste/pull/126

PR #109, opened 2021-12-30
Documentation for waveform interpolation

BenjaminRodenberg:
Provides documentation for new feature in https://github.com/precice/precice/pull/1029.


pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
+---
+title: Waveform iteration for time interpolation of coupling data
+permalink: couple-your-code-waveform.html
+keywords: api, adapter, time, waveform, subcycling, multirate
+summary: "With Waveform iteration you can interpolate your coupling data in time for higher order time stepping and more stable subcycling."
```
```suggestion
summary: "With waveform iteration, you can interpolate coupling data in time for higher-order time stepping and more stable subcycling."
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+---
+
+{% include warning.html content="These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care." %}
+{% include note.html content="In this section, we will only discuss implicit coupling. We will, without loss of generality, only discuss the use of the API functions `readBlockVectorData` and `writeBlockVectorData` in the examples." %}
```
```suggestion
{% include note.html content="We only discuss implicit coupling. Without loss of generality, we moreover only discuss the API functions `readBlockVectorData` and `writeBlockVectorData` in the examples." %}
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+preCICE allows the participants to use subcycling - meaning: to work with different time step sizes on their respective domains. If participants are using different timestep sizes, they have to synchronize at the end of each *time window*. If you want to know how this works and what a time window is, see ["Step 5 - Non-matching time step sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In this section, we will take a closer look at the exchange of coupling data when subcycling, and advanced techniques for interpolation of coupling data inside of a time window.
```
```suggestion
preCICE allows the participants to use subcycling - meaning: to work with different time step sizes on their respective domains. If participants are using different timestep sizes, they have to synchronize at the end of each *time window*. If you want to know how this works and what a time window is, see ["Step 5 - Non-matching time step sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In this section, we take a closer look at the exchange of coupling data when subcycling, and advanced techniques for interpolation of coupling data inside of a time window.
```

Subcycling does mean different time step sizes on both sides (they could also be similar). This paragraph needs some tweaking.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Exchange of coupling data with subcycling
+
+preCICE only exchanges data and performs the coupling at the end of the last time step in the time window - the end of the time window. For the default case, preCICE only exchanges the data that was written at the very end of the time window. This approach automatically leads to discontinuities or "jumps" when going from one time windows to the next and, thus, lower accuracy (for details, see [here](https://onlinelibrary.wiley.com/doi/full/10.1002/nme.6443)).
```
```suggestion
preCICE only exchanges data at the end of the last time step in each time window – the end of the time window. By default, preCICE only exchanges data that was written at the very end of the time window. This approach automatically leads to discontinuities or "jumps" when going from one time windows to the next and, thus, lower accuracy (for details, see [here](https://onlinelibrary.wiley.com/doi/full/10.1002/nme.6443)).
```
Please don't use "here" as link text.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+{% include important.html content="If subcycling is used, data that was written before the last time step in the window will be ignored and will not be exchanged. This means that coupling data has a constant value (in time) within one coupling window." %}
```
uekerman: Isn't that what you just said? Seems redundant and a misuse of the "important" box. Also, please no "will" :grin:

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Example for subcycling without waveform iteration
+
+The figure below visualizes this situation for a single coupling window ranging from $$t_\text{ini}$$ to $$t_\text{ini}+\Delta t$$:
+
+![Coupling data exchange without interpolation](images/docs/couple-your-code/couple-your-code-waveform/WaveformConstant.png)
+
+The two participants Dirichlet $$\mathcal{D}$$ and Neumann $$\mathcal{N}$$ use their respective time step sizes $$\delta t$$ and produce coupling data $$c$$ at the end of each time step. But only the very last samples $$c_{\mathcal{N}\text{end}}$$ and $$c_{\mathcal{D}\text{end}}$$ are exchanged. If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it will always receive the same value $$c_{\mathcal{N}\text{end}}$$ from the Neumann participant $$\mathcal{N}$$, independent from the current time step.
```
```suggestion
The two participants Dirichlet $$\mathcal{D}$$ and Neumann $$\mathcal{N}$$ use their respective time step sizes $$\delta t$$ and produce coupling data $$c$$ at the end of each time step. But only the very last samples $$c_{\mathcal{N}\text{end}}$$ and $$c_{\mathcal{D}\text{end}}$$ are exchanged. If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it always receives the same value $$c_{\mathcal{N}\text{end}}$$ from the Neumann participant $$\mathcal{N}$$, independent from the current time step.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Linear interpolation in a time window
+
+A simple extension of the existing API is to apply linear interpolation inside of a time window to get smoother coupling boundary conditions. With this approach time-dependent functions (so-called *waveforms*) are exchanged between the participant. Since these waveforms are exchanged iteratively in implicit coupling, we call this procedure *waveform iteration*. Exchanging waveforms leads to a more robust subcycling and allows us to support higher order time stepping (for details, see [here](https://onlinelibrary.wiley.com/doi/full/10.1002/nme.6443)).
```
uekerman: Again, no "here" links, please.
> A simple extension of the existing API is to apply linear interpolation inside of a time window

The linear interpolation is not done by extending the API. This sentence needs tweaking.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+preCICE offers an experimental API for using linear interpolation on time windows. Here, `readBlockVectorData` accepts an additional argument `dt`. This allows us to choose the time where the interpolant should be sampled:
+
+```cpp
+// stable API with constant data in time window
+void readBlockVectorData(int dataID, int size, const int* valueIndices, double* values) const;
+
+// experimental API for waveform iteration
+void readBlockVectorData(int dataID, int size, const int* valueIndices, double dt, double* values) const;
+```
```
uekerman: Don't forget to update the variable name "dt".

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+{% include note.html content="The functionality of `writeBlockVectorData` remains unchanged. All samples but the one from the very last time step in the time window are ignored. An extension of `writeBlockVectorData` to accept multiple samples per time window is work in progress." %}
```
uekerman: Not sure if you need to make this a note. Also fits into the text without the note.
(Too many boxes make the content harder to skim)

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+The experimental API has to be activated in the configuration file via the `experimental` attribute. This allows us to define the order of the interpolant in the `read-data` tag of the corresponding `participant`. Here, `waveform-order="1"` corresponds to linear interpolation and `waveform-order="0"` to constant interpolation (this is just the same as the stable API).
```
uekerman: Sounds complicated. Just say that the default value of `waveform-order` is 0.

pages/docs/couple-your-code/couple-your-code-waveform.md
uekerman: Actually, we are not yet sure whether we really need an extension of `writeBlockVectorData`, no?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it samples the data from a time-dependent function $$c_\mathcal{D}(t)$$. This function is created from linear interpolation of the first and the last sample $$c_{\mathcal{D}0}$$ and $$c_{\mathcal{D}5}$$ created by the Neumann participant $$\mathcal{N}$$ in the current time window. This allows $$\mathcal{D}$$ to sample the coupling condition at arbitrary times $$t$$ inside the current time window.
+
+## Experimental API for waveform iteration
+
+preCICE offers an experimental API for using linear interpolation on time windows. Here, `readBlockVectorData` accepts an additional argument `dt`. This allows us to choose the time where the interpolant should be sampled:
+
+```cpp
+// stable API with constant data in time window
+void readBlockVectorData(int dataID, int size, const int* valueIndices, double* values) const;
```
uekerman, line 47: Is the data really constant within a time window? I thought we get the linear interpolation here as well? (if the read side is subcycling).

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+The experimental API has to be activated in the configuration file via the `experimental` attribute. This allows us to define the order of the interpolant in the `read-data` tag of the corresponding `participant`. Here, `waveform-order="1"` corresponds to linear interpolation and `waveform-order="0"` to constant interpolation (this is just the same as the stable API).
+
+```xml
+<solver-interface experimental="true" ... >
+...
+    <participant name="FluidSolver">
+        <use-mesh name="FluidMesh" provide="yes"/>
+        <write-data name="Forces" mesh="MyMesh"/>
+        <read-data name="Displacements" mesh="FluidMesh" waveform-order="1"/>
```
uekerman, line 65: Did we always have the new attribute here? Actually, it makes a lot of sense to have it here :+1:

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+  solveTimeStep(dt); // might be using midpoint rule for time-stepping
+  computeForces(forces);
+  precice.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
+  precice_dt = precice.advance(dt);
+  endTimeStep(); // e.g. update variables, increment time
+}
+precice.finalize();
+delete[] vertexIDs, forces, displacements;
+turnOffSolver();
+```
```
uekerman: I guess you could leave out the obvious parts here and abbreviate with `[...]`.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Use waveform iteration with subcycling
```
uekerman: Why is the code above "without" and this one "with" subcycling? They can both be used for both cases. Seems to me like one example could be enough.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+    precice.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
+  }
+  precice_dt = precice.advance(dt);
+  endTimeStep(); // e.g. update variables, increment time
+}
+```
+
+### Note on the use of `initializeData`
+
+{% include important.html content="`writeBlockVectorData` *after* `initializeData` writes data corresponding to the *end* of the window." %}
```
uekerman: This, I don't get. Seems also like a misuse of the "important" box?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+
+In the very first time window of our coupled simulation we have to provide initial data in order to be able to perform linear interpolation. We have to call `writeBlockVectorData` *before* `initializeData`. This data will be used as initial data for the interpolation. If no initial data is provided, only constant interpolation can be applied in the first window, but linear interpolation is still available in later time windows.
```
uekerman: I am again confused by this. Why don't we perform linear interpolation starting at zero if `initializeData` is not called?
Not calling `initializeData` should give the exact same behavior as calling `initializeData` with zero data.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,167 @@
...
+  if (precice.isWriteDataRequired(dt)){ // only true at the end of the time window
+    // sets forces at the end of the current time window
+    precice.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
+  }
+  precice_dt = precice.advance(dt);
+  ...
+}
+```
+
+{% include note.html content="Depending on the coupling scheme, the used interpolation scheme might differ depending on the order of the participants and depending on whether we are in the first or a later coupling iteration." %}
```
uekerman: By "order" here you mean whether a participant is first or second, right? Might be confusing.
Overall, I find this statement confusing. We should be able to explain these cases. If not, we are doing sth wrong.

uekerman (changes requested):
I think it was a good step to write this down. We clearly see where there are still open problems.

Before merging at some point, also somebody not familiar with the new functionality should review this.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I'm experimenting with footnotes for external literature in 1bb5c4e. I think it's useful that the reference is on the same page. At the sametime I found always writing "see [out paper on ...](https://onlinelibrary.wiley.com/doi/full/10.1002/nme.6443)" quite repetitive.

It's still quite redundant to our [literature guide](https://precice.org/fundamentals-literature-guide.html) and especially the [preCICE publication list](https://precice.org/publications.html). I think a general solution to avoid duplicating references would be nice, but also more complicated and as far as I know we currently just duplicate references, if this is required.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I agree. My idea here was to summarize and highlight the important part, but since the summary is not much shorter than the actual text and there are already too many boxes, I will remove this one.

BenjaminRodenberg: 6ee5743a27803117788ebd7201b4220ac36ac8e8

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: That's right. Actually I think we can completely ignore the API here and only talk about the theory. API then comes later. 4f717b932fba236ae15c8012fd1bc653ed8a962f

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: That's work in progress. Depending on what we do, we might be able to keep `writeBlockVectorData` as it is. I'm describing the steps in https://github.com/precice/precice/issues/1171. I will also add a link to this issue to the documentation here.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Changed this in 00373c511500731c2c6af2b262765b5807c51f49. I still think we should mention that `waveform-order="1"` is linear.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Yes. Data is constant, if this function call is used. From the current state of the [waveform pull request](https://github.com/BenjaminRodenberg/precice/blob/3b0a94b50f2d781019379ff285e9db8b0206e22a/src/precice/impl/SolverInterfaceImpl.hpp#L420-L436):

```
  /**
   * @brief Reads vector data values given as block from the interface mesh at the end of the time window.
   * ...
   */
  void readBlockVectorData(
      int        toDataID,
      int        size,
      const int *valueIndices,
      double *   values) const;
```

Looking [at the implementation](https://github.com/BenjaminRodenberg/precice/blob/3b0a94b50f2d781019379ff285e9db8b0206e22a/src/precice/impl/SolverInterfaceImpl.cpp#L1192-L1194) also shows this:

```
  double timeStepDt        = _couplingScheme->getThisTimeWindowRemainder(); // samples at end of time window
  bool   checkExperimental = false;
  return readBlockVectorData(dataID, size, valueIndices, timeStepDt, values, checkExperimental);
```

The reason is that when we call `read...Data(...)` without `dt`, we cannot know where the user wants to sample (even if subcycling is done). Remember `advance(dt)` defines the length of the time step. `advance` is called after `read` and therefore `read` cannot make any better assumptions than that the user want to read data at the end of the window.

BenjaminRodenberg: @uekerman does this sound reasonable? If yes, feel free to resolve this discussion.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I actually moved it here recently due to a technical problem with the implementation. I don't exactly remember the reason, but I did it here: https://github.com/BenjaminRodenberg/precice/commit/41751cf9d840a8cea487320ba9f371adbb9366b8. Previously we used
```
<data:vector name="Displacements">
  <waveform order="1" />
</data:vector>
```
Why do you think it makes a lot of sense now? I was only changing this for pragmatic reasons ;)

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Done. 07fb39cf7a8fd05ab3cf9afbb854130f57e703cf

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I think this statement is not necessary anymore, because we agreed to exclude explicit coupling. For implicit coupling the execution order of participants does not matter (besides the first iteration, but I don't think we have to mention this).

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: We discussed this issue and agreed on using zero initial data. This still has to be implemented in https://github.com/precice/precice/pull/1029.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: My idea here was to tell the reader that `initializeData` is automatically needed to set data at the beginning of the simulation, but I think this is clear and not needed, after our revision of how `initializeData` and waveform relaxation should work.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I will use `relativeReadTime` here and update https://github.com/precice/precice/pull/1029 soon. See 44df155740841b566faddcda6bf15f174be40780.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Revised this in 8ba2573eab4a20317a4cc5c247d21438d9f8f284.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Revised this in 8ba2573eab4a20317a4cc5c247d21438d9f8f284.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Revised this in 8ba2573eab4a20317a4cc5c247d21438d9f8f284.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Good point. I managed to boil the example down. Actually the initialization part is now also mostly irrelevant. See bc6702626e73b0667aeea6eceff6f499b7b74ead.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+---
+
+{% warning %}
+These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care."
+{% endwarning %}
+{% note %}
+This feature is only available for implicit coupling. Without loss of generality, we moreover only discuss the API functions `readBlockVectorData` and `writeBlockVectorData` in the examples.
+{% endnote %}
+
+preCICE allows the participants to use subcycling - meaning: to work with individual time step sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a time step or you want to know how subcycling works in detail, see ["Step 5 - Non-matching time step sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling, and advanced techniques for interpolation of coupling data inside of a time window.
```
```suggestion
preCICE allows the participants to use subcycling - meaning: to work with individual timestep sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a timestep or you want to know how subcycling works in detail, see ["Step 5 - Non-matching timestep sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling, and advanced techniques for interpolation of coupling data inside of a time window.
```

I know, I am picky here, but we have so far used "timestep" (docs and code).
More occasions below.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Exchange of coupling data with subcycling
+
+preCICE only exchanges data at the end of the last time step in each time window – the end of the time window. By default, preCICE only exchanges data that was written at the very end of the time window. This approach automatically leads to discontinuities or "jumps" when going from one time windows to the next. Coupling data has a constant value (in time) within one coupling window. This leads to lower accuracy of the overall simulation (for details, see [^1]).
```
```suggestion
preCICE only exchanges data at the end of the last time step in each time window – the end of the time window. By default, preCICE only exchanges data that was written at the very end of the time window. This approach automatically leads to discontinuities or "jumps" when going from one time window to the next. Coupling data has a constant value (in time) within one coupling window. This leads to lower accuracy of the overall simulation (for details, see [^1]).
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+![Coupling data exchange without interpolation](images/docs/couple-your-code/couple-your-code-waveform/WaveformConstant.png)
+
+The two participants Dirichlet $$\mathcal{D}$$ and Neumann $$\mathcal{N}$$ use their respective time step sizes $$\delta t$$ and produce coupling data $$c$$ at the end of each time step. But only the very last samples $$c_{\mathcal{N}\text{end}}$$ and $$c_{\mathcal{D}\text{end}}$$ are exchanged. If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it always receives the same value $$c_{\mathcal{N}\text{end}}$$ from the Neumann participant $$\mathcal{N}$$, independent from the current time step.
+
+## Linear interpolation in a time window
+
+A simple solution to reach higher accuracy is to apply linear interpolation inside of a time window to get smoother coupling boundary conditions. With this approach time-dependent functions (so-called *waveforms*) are exchanged between the participant. Since these waveforms are exchanged iteratively in implicit coupling, we call this procedure *waveform iteration*. Exchanging waveforms leads to a more robust subcycling and allows us to support higher order time stepping (for details, see [^1]).
```
```suggestion
A simple solution to reach higher accuracy is to apply linear interpolation inside of a time window to get smoother coupling boundary conditions. With this approach time-dependent functions (so-called *waveforms*) are exchanged between the participants. Since these waveforms are exchanged iteratively in implicit coupling, we call this procedure *waveform iteration*. Exchanging waveforms leads to a more robust subcycling and allows us to support higher order time stepping (for details, see [^1]).
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+![Coupling data exchange with linear interpolation](images/docs/couple-your-code/couple-your-code-waveform/WaveformLinear.png)
+
+If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it samples the data from a time-dependent function $$c_\mathcal{D}(t)$$. This function is created from linear interpolation of the first and the last sample $$c_{\mathcal{D}0}$$ and $$c_{\mathcal{D}5}$$ created by the Neumann participant $$\mathcal{N}$$ in the current time window. This allows $$\mathcal{D}$$ to sample the coupling condition at arbitrary times $$t$$ inside the current time window.
+
+## Experimental API for waveform iteration
+
+If we want to improve the accuracy by using waveforms, this requires an extension of the existing API, because we need a way to tell preCICE where we want to sample the waveform. For this purpose, preCICE offers an experimental API, which is currently only supporting a single linear interpolation along the complete time window. Here, `readBlockVectorData` accepts an additional argument `relativeReadTime`. This allows us to choose the time where the interpolant should be sampled:
```
```suggestion
If we want to improve the accuracy by using waveforms, this requires an extension of the existing API, because we need a way to tell preCICE where we want to sample the waveform. For this purpose, preCICE offers an experimental API, which is currently only supporting a linear interpolation along the complete time window. Here, `readBlockVectorData` accepts an additional argument `relativeReadTime`. This allows us to choose the time where the interpolant should be sampled:
```
What would "single" stand for here?

Also: You wet here the appetite for the new method and then you first explain the old method just below. I think it would be better to postpone the behavior of the old method when using linear interpolation till the end.

pages/docs/couple-your-code/couple-your-code-waveform.md
uekerman: > Remember advance(dt) defines the length of the time step. advance is called after read and therefore read cannot make any better assumptions than that the user want to read data at the end of the window.

You are right. That's the tricky point. We need to bring this explanation here. If I already forget it, users will struggle to understand it as well :grin: 

We might even need to print a warning if a user chooses linear interpolation, but calls "read" without "dt". There is no use case, right? It probably means that the user got it wrong.

Letting this open as a reminder.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+```cpp
+// stable API with constant data in time window
+void readBlockVectorData(int dataID, int size, const int* valueIndices, double* values) const;
+
+// experimental API for waveform iteration
+void readBlockVectorData(int dataID, int size, const int* valueIndices, double relativeReadTime, double* values) const;
+```
+
+`relativeReadTime` describes the time relatively to the beginning of the current time step. This means that `relativeReadTime = 0` gives us access to data at the beginning of the time step. Since we will call `advance(dt)` at a later point in time to finalize the time step, `relativeReadTime = dt` gives us access to data at the end of the time step.
```
uekerman: The second sentence needs tweaking. What does advance have to do with this? Maybe it could be easier to first explain `relativeReadTime = dt` saying that `dt` is the timestep size. And then, more as I reminder, that you tell preCICE in advance what `dt` you used in your solver.
Also, please no "will" here :grin:

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+{% note %}
+The functionality of `writeBlockVectorData` remains unchanged, because the data at the beginning and at the end of the window are sufficient to create a linear interpolant over the window. Therefore, all samples but the one from the very last time step in the time window are ignored. This might, however, change in the future (see [precice/#1171](https://github.com/precice/precice/issues/1171)).
+{% endnote %}
```
uekerman: I still think that this notes here confuses more than it helps for the moment. I would vote for removing it. If at some point we really introduce new write functions as well, we will need additional documentation anyway.

pages/docs/couple-your-code/couple-your-code-waveform.md
uekerman: In a way, from user perspective, waveform iteration is a pure "read data" feature. You get "better" data when reading. All the rest are implementation details for us, but not for the user.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+        <write-data name="Forces" mesh="MyMesh"/>
+        <read-data name="Displacements" mesh="FluidMesh" waveform-order="1"/>
+    </participant>
+...
+</solver-interface>
+```
+
+## Usage example
+
+We are now ready to extend the example from ["Step 6 - Implicit coupling"](couple-your-code-implicit-coupling.html) to use waveforms. Only few changes are necessary to sample the `Displacements` at the middle of the time window, which might be necessary for our specific application:
```
uekerman: "our specific application" sounds odd. Why not starting this paragraph with "Let us assume that the fluid solver uses a midpoint rule as time stepping method ...

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+```cpp
+// no relevant changes for initialization
+// See "Step 6 - Implicit coupling"
```
```suggestion
```
That's clear.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+...
+precice_dt = precice.initialize();
+while (not simulationDone()){ // time loop
+  // write checkpoint
+  ...
+  dt = beginTimeStep(); // e.g. compute adaptive dt 
+  dt = min(precice_dt, dt);
+  if (precice.isReadDataAvailable()){ // always true, because we can sample at arbitrary points
```
uekerman: I think we should explain this in the API section above already.

uekerman: And this only holds if "order=1" is configured. I don't think that this is already clear to the reader.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+  solveTimeStep(dt); // might be using midpoint rule for time-stepping
+  if (precice.isWriteDataRequired(dt)){ // only true at the end of the time window
+    computeForces(forces);
+    precice.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
+  }
+  precice_dt = precice.advance(dt);
+  // read checkpoint & endTimeStep  
+  ...
+}
+// usual finalization
```
```suggestion
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+...
+```
+
+The `precice.isReadDataAvailable()` and `precice.isWriteDataRequired(dt)` are optional, but recommended -- especially, if subcycling is used.
```
```suggestion
```
That's introduced in step 5 already. No need to repeat here.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+As described in ["Step 7 - Data initialization"](couple-your-code-initializing-coupling-data), we can also use `initializeData` to provide initial data for the interpolation. If `initializeData` is not called, preCICE uses zero initial data for constructing the interpolant.
```
uekerman: This part here might confuse more than it helps.
In a way, this is explained in step 7 and we do the logical extension here.
I would suggest either proper explanation with maybe a picture or drop the paragraph.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Literature
+
+[^1]: Rüth, B, Uekermann, B, Mehl, M, Birken, P, Monge, A, Bungartz, H-J. Quasi-Newton waveform iteration for partitioned surface-coupled multiphysics applications. Int J Numer Methods Eng. 2021; 122: 5236– 5257. [https://doi.org/10.1002/nme.6443](https://doi.org/10.1002/nme.6443)
```
uekerman, line 102: Have you checked how the "^1" works and looks? We don't use this on other pages, e.g. https://precice.org/tutorials-elastic-tube-1d.html

uekerman (changes requested):
Looking good :+1:
Still a few tweaks, but should all be easy.
Then, I would suggest that somebody not so familiar with your PRs looks over this.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Waveform iteration for time interpolation of coupling data
+permalink: couple-your-code-waveform.html
+keywords: api, adapter, time, waveform, subcycling, multirate
+summary: "With waveform iteration, you can interpolate coupling data in time for higher-order time stepping and more stable subcycling."
+---
+
+{% warning %}
+These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care."
+{% endwarning %}
```
```suggestion
{% experimental %}
These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care."
{% endexperimental %}
```

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Fixed in [](https://github.com/precice/precice.github.io/pull/109/commits/727d1f371815945209164c8629d5397a1aaf6687)[727d1f3](https://github.com/precice/precice.github.io/pull/109/commits/727d1f371815945209164c8629d5397a1aaf6687).

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Yes. It's basically creating a wikipedia-style footnote. I tweaked it a bit more in c9a944914147d376041886c38c88a822589dec73:
Screenshot from 2022-02-15 11-50-05
I think it would also be good to use this style more often. We can open an issue on this topic.

BenjaminRodenberg: Could not resist: https://github.com/precice/precice.github.io/issues/130 :smirk:

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Done. fef7da7a7a752bbd51278c62c27de387d437c8d8

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Fixed. 966be3cfb4237a2293d7eeacb466f96d48a76f51

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: No. I think `isReadDataAvailable` should always return `true`, even if `order=0` is configured. As soon as a user says "I want to use the experimental API", sampling is allowed by providing the additional argument to `readBlockScalarData` and therefore data is also available. Currently, the implementation is [as follows](https://github.com/BenjaminRodenberg/precice/blob/438de8f881fbb9b7c3bf93d612da96e2935bcc3f/src/precice/impl/SolverInterfaceImpl.cpp#L542-L550):

```
bool SolverInterfaceImpl::isReadDataAvailable() const
{
  PRECICE_TRACE();
  PRECICE_CHECK(_state != State::Constructed, "initialize() has to be called before isReadDataAvailable().");
  PRECICE_CHECK(_state != State::Finalized, "isReadDataAvailable() cannot be called after finalize().");
  bool available = _couplingScheme->hasDataBeenReceived();
  available |= (_allowsExperimental && _couplingScheme->hasInitialDataBeenReceived()); // if waveform relaxation is allowed, we always return true as soon as initialData is available.
  return available;
}
```

This is simpler and does not introduce another exception, because as soon as you are using the experimental API, `isReadDataAvailable` almost always returns true. With waveform relaxation I also think that the actual purpose of this API method becomes a bit unclear.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Understood and agree :+1: Thanks for the explanation :)

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: "single" was meant as not "piecewise linear", but I agree that this wording is not good.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: If removing it is an option, I will remove it. Shorter is better.

BenjaminRodenberg: Done 872e20bb5a4da60883d31dd6c5f0c1f1245d2489.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Done in 872e20bb5a4da60883d31dd6c5f0c1f1245d2489.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I changed this paragraph. Still, we need to explain quite a lot of things here. Subcycling is just complicated and the feature is also currently not very well documented unfortunately. I hope it will become clearer in the future, but for now I think it is good enough to know how to use the API.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Just talking about "waveforms" now. From 58281711af25dfba9b630b50058251d08e3688f7 it should be clear that there are currently only two possibilities for interpolation.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Added a clarification in 18a89043f8112ce56e8e0cf9fd73b9e5d5adb7df.

BenjaminRodenberg: Good point with the warning. Implemented it in https://github.com/precice/precice/pull/1029/commits/909520f82e8b5f6887941296b1a7dfcb117d67ea.

pages/docs/couple-your-code/couple-your-code-waveform.md
uekerman: But `experimental` includes many things, not only waveform relaxation.
The purpose of this function is to avoid unnecessary reads from the user. Imagine they have to do some costly conversion after calling read. In this sense, it should be false if `order=0`. It would be meaningless to read then, no?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+{% experimental %}
+These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care.
+{% endexperimental %}
+{% note %}
+This feature is only available for parallel implicit coupling. An extension to serial implicit coupling is planned. Without loss of generality, we moreover only discuss the API functions `readBlockVectorData` and `writeBlockVectorData` in the examples.
+{% endnote %}
+
+preCICE allows the participants to use subcycling - meaning: to work with individual timestep sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a timestep or you want to know how subcycling works in detail, see ["Step 5 - Non-matching timestep sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling, and advanced techniques for interpolation of coupling data inside of a time window.
```
```suggestion
preCICE allows the participants to use subcycling – meaning: to work with individual timestep sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a timestep or you want to know how subcycling works in detail, see ["Step 5 - Non-matching timestep sizes" of the step-by-step guide](couple-your-code-timestep-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling and advanced techniques for interpolation of coupling data inside of a time window.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Example for subcycling without waveform iteration
+
+The figure below visualizes this situation for a single coupling window ranging from $$t_\text{ini}$$ to $$t_\text{ini}+\Delta t$$:
+
+![Coupling data exchange without interpolation](images/docs/couple-your-code/couple-your-code-waveform/WaveformConstant.png)
+
+The two participants Dirichlet $$\mathcal{D}$$ and Neumann $$\mathcal{N}$$ use their respective timestep sizes $$\delta t$$ and produce coupling data $$c$$ at the end of each timestep. But only the very last samples $$c_{\mathcal{N}\text{end}}$$ and $$c_{\mathcal{D}\text{end}}$$ are exchanged. If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it always receives the same value $$c_{\mathcal{N}\text{end}}$$ from the Neumann participant $$\mathcal{N}$$, independent from the current timestep.
```
```suggestion
The two participants Dirichlet $$\mathcal{D}$$ and Neumann $$\mathcal{N}$$ use their respective timestep sizes $$\delta t$$ and produce coupling data $$c$$ at the end of each timestep. But only the very last samples $$c_{\mathcal{N}\text{end}}$$ and $$c_{\mathcal{D}\text{end}}$$ are exchanged. If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it always reads the same value $$c_{\mathcal{N}\text{end}}$$ from preCICE, independent from the current timestep.
```

Just to avoid a common misunderstanding on how preCICE works.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -0,0 +1,102 @@
...
+
+If we choose to use a smaller timestep size `dt < precice_dt`, we apply subcycling and therefore `relativeReadTime = dt` corresponds to sampling data at the end of the timestep. But we can also use smaller values for `relativeReadTime`, as shown in the usage example below. When using subcycling, it is important to note that `relativeReadTime = precice_dt` is the default behavior, if no `relativeReadTime` is provided, because preCICE cannot know the `dt` our solver wants to use. This also means that is subcycling is applied one must use the experimental API and provide `relativeReadTime` to benefit from the higher accuracy waveforms.
```
```suggestion
If we choose to use a smaller timestep size `dt < precice_dt`, we apply subcycling and therefore `relativeReadTime = dt` corresponds to sampling data at the end of the timestep. But we can also use smaller values for `relativeReadTime`, as shown in the usage example below. When using subcycling, it is important to note that `relativeReadTime = precice_dt` is the default behavior, if no `relativeReadTime` is provided, because preCICE cannot know the `dt` our solver wants to use. This also means that if subcycling is applied one must use the experimental API and provide `relativeReadTime` to benefit from the higher accuracy waveforms.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
uekerman: Let's add documentation about this later. The usage example works like it is right now.
I assume you changed the implementation in meanwhile? That one still needs fixing in my opinion.

uekerman (approved):
Minor language tweaks, otherwise good to go from my side.
Could still be helpful to let a non-expert read over this.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I added 68f0587cf4a389a1f01892d84f6b8804c5fba004. This reflects the current state as accurate as possible. I think it helps to say something (not too much) about the behavior of `isReadDataAvailable` here. We can add the detailed documentation as soon as https://github.com/precice/precice/issues/1223 is converged.

boris-martin:
I just had a quick read as a naive reader, and I wondered how we reconcile order of the waveform and numbers of subcycles. If I want a 2nd order waveform but perform 10 steps, how will the polynom be built ? Least-squares fit ? Arbirtray selection of 3 points ?
An additional paragraph (or a sentence/reference to something else) looks welcome to me.
Otherwise it's quite easy to understand.

BenjaminRodenberg:
> I just had a quick read as a naive reader, and I wondered how we reconcile order of the waveform and numbers of subcycles. 

Currently the implementation is very limited here. There is basically only the possibility to create a linear interpolant from the first and last sample in the window. If you are subcycling, all samples between will be ignored (they are not even communicated). I'm describing the current restrictions and the rough plan here: https://github.com/precice/precice/issues/1171.

PR #110, opened 2022-01-13
Move link to box out of "how to use" section

uekerman:
In a lecture, a student directly clicked on the link, downloaded the box from the website and did then not know how to go on.
By removing the link, we could avoid this misunderstanding.

Not directly related: Another student reported that they ran into a firewall problem on windows similar to [this one](https://superuser.com/questions/1013019/vagrant-up-not-working-on-windows-7-behind-firewall) maybe. I do not have the error message.
@MakisH Have you seen this one already? Should we add somewhere a statement that maybe users have to tweak their firewall?

### DIFF
--- a/pages/docs/installation/installation-vm.md
+++ b/pages/docs/installation/installation-vm.md
@@ -30,7 +30,7 @@ This contains all the solvers and adapters used in our tutorials, already built 
 
 ![Screnshot](images/docs/docs-installation-vm-screenshot.jpg)
 
-See [what is included](https://github.com/precice/vm/blob/main/README.md#what-is-included) in detail.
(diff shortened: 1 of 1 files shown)


MakisH (approved):
You are right, this is currently indeed confusing, and this is a much better formulation! :+1: Thanks for the direct edit!

MakisH:
> Not directly related: Another student reported that they ran into a firewall problem on windows similar to [this one](https://superuser.com/questions/1013019/vagrant-up-not-working-on-windows-7-behind-firewall) maybe. I do not have the error message. @MakisH Have you seen this one already? Should we add somewhere a statement that maybe users have to tweak their firewall?

I have no idea about this. My gut feeling tells me that the user may have already seen similar errors before (e.g., in Docker) and found this system-specific issue rather easily. But if this is, e.g., in a university network, we should definitely mention it. We could start a (minimized) "troubleshooting" section at the bottom.

PR #111, opened 2022-01-17
Remove early bird announcement of the workshop

davidscn:


IshaanDesai (approved):
Time to remove this yes :+1: can we also add a note that we are still accepting abstracts for talks? It would be good to explicitly mention this.

davidscn:
> can we also add a note that we are still accepting abstracts for talks?

Let's discuss tomorrow

PR #115, opened 2022-01-27
Add documentation of binprecice

### LINKED ISSUE #99: Add binprecice to docs
Add information regarding binprecice:
* `binprecice version`https://github.com/precice/precice/pull/1122
* `binprecice check` https://github.com/precice/precice/pull/1132

fsimonis:
This PR adds documentation for binprecice.
I called the page `builtin`, as we may rename the binary in the future.

Closes #99

### DIFF
--- a/pages/docs/tooling/tooling-builtin.md
+++ b/pages/docs/tooling/tooling-builtin.md
@@ -9,6 +9,7 @@ Part of a preCICE installation is the tool `binprecice`.
 It provides an easy-to-use interface to tooling API of the preCICE library.
 
 With `binprecice`, you can get the installed preCICE version, generate a reference of all available configuration options, as well as check your configuration file for basic configuration issues.
+
 ## XML reference
 
 ```bash
--- a/pages/docs/tooling/tooling-builtin.md
+++ b/pages/docs/tooling/tooling-builtin.md
@@ -72,7 +72,7 @@ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
 The `check` runs the preCICE configuration parsing and checking logic on the given configuration file.
 This will find the majority of the configuration mistakes without having to start a simulation.
 These checks include wrong tags and attribute values, and more elaborate naming checks.
-More intricate logic, such as checks if all necessary data are exchanged in a Coupling Schemes, are not covered.
+More advanced logic, such as checks if all necessary data are exchanged in a coupling scheme, are not covered.
 
 The basic usage is to simply check a configuration file:
 
--- a/pages/docs/tooling/tooling-builtin.md
+++ b/pages/docs/tooling/tooling-builtin.md
@@ -10,10 +10,10 @@ It provides an easy-to-use interface to tooling API of the preCICE library.
 
 ## XML reference
 
-```
-$ binprecice md
-$ binprecice xml
-$ binprecice dtd
+```bash
+binprecice md
+binprecice xml
+binprecice dtd
 ```
 
 This prints the XML reference to the console in various flavours.
@@ -26,8 +26,8 @@ You can find the reference of the latest release [on the website](configuration-
 It is possible to generate a local version of the reference by rendering the Markdown to HTML using the `markdown` command.
 Be aware that this version does not contain styling, LaTeX rendering, and functioning links.
 
-```
-$ binprecice md | markdown > reference.html
+```bash
+binprecice md | markdown > reference.html
 ```
 
 ### `xml`
@@ -44,8 +44,8 @@ This prints the DTD information, which can be used to validate the XML configura
 This feature is available since version 2.4.0.
 {% endversion %}
 
-```
-$ binprecice version
+```bash
+binprecice version
 ```
 
 This prints the version information of preCICE, which consists of multiple semicolon-separated parts.
@@ -57,15 +57,14 @@ This prints the version information of preCICE, which consists of multiple semic
 3. Configuration options for MPI, PETSc, Python and some more.
 4. Compilation and linker flags used to build preCICE
 
-
 ## Configuration check
 
 {% version 2.4.0 %}
 This feature is available since version 2.4.0.
 {% endversion %}
 
-```
-$ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
+```bash
+binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
 ```
 
 The `check` runs the preCICE configuration parsing and checking logic on the given configuration file.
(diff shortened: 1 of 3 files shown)


pages/docs/tooling/tooling-builtin.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+To enable more checks, additionally pass the name of one participant.
+This should be repeated for every defined participant in the simulation.
+
+```bash
+binprecice check precice.xml Fluid
+```
+
+To enable all available checks, you may additionally pass the communicator size of the Participant.
```
uekerman:
```suggestion
To enable all available checks, you may additionally pass the communicator size of the participant.
```

pages/docs/tooling/tooling-builtin.md
```diff
@@ -0,0 +1,91 @@
...
+
+{% version 2.4.0 %}
+This feature is available since version 2.4.0.
+{% endversion %}
+
+```bash
+binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
+```
+
+The `check` runs the preCICE configuration parsing and checking logic on the given configuration file.
```
uekerman, line 72: We have to be careful to not oversell the feature here. A user might think that the call also checks whether the exchange tags are set correctly or similar. It "only" validates whether the configuration follows the semantics, not the "logic" of  the configuration.

pages/docs/tooling/tooling-builtin.md
```diff
@@ -0,0 +1,91 @@
...
+
+## XML reference
+
+```bash
+binprecice md
+binprecice xml
+binprecice dtd
+```
+
+This prints the XML reference to the console in various flavours.
```
uekerman:
```suggestion
This prints the XML reference to the console in various flavors.
```
:grin:

uekerman (approved):
I like, only minor comments.

pages/docs/tooling/tooling-builtin.md
```diff
@@ -10,10 +10,10 @@ It provides an easy-to-use interface to tooling API of the preCICE library.
```
MakisH, line 10: As part of this PR, maybe also extend the introduction a bit. "an easy-to-use interface to tooling API" is a bit vague for the user.

Maybe write something like:

```suggestion

With `binprecice`, you can get the installed preCICE version, generate a reference of all available configuration options, as well as check your configuration file for basic configuration issues.
```

pages/docs/tooling/tooling-builtin.md
MakisH: Maybe also give an example of issues it can detect and an example of issues it cannot detect?

pages/docs/tooling/tooling-builtin.md
```diff
@@ -57,37 +57,35 @@ This prints the version information of preCICE, which consists of multiple semic
...
 This should be repeated for every defined participant in the simulation.
 
-```
-$ binprecice check precice.xml Fluid
+```bash
+binprecice check precice.xml Fluid
 ```
 
 To enable all available checks, you may additionally pass the communicator size of the Participant.
 This enables some niche checks, which should not be necessary in the vast majority of cases.
```
MakisH: While I do like that the "niche checks" tells me this is probably not something I need, I think that this is also a bit empty of additional information. Some examples / clarification in parentheses would help, otherwise every time I have a problem I would just enable everything without thinking.

pages/docs/tooling/tooling-builtin.md
fsimonis: True, some more information would be appropriate here.

> A user might think that the call also checks whether the exchange tags are set correctly or similar. It "only" validates whether the configuration follows the semantics, not the "logic" of the configuration.

This is a bit wishy washy at the moment and mostly due to how our configuration works.

As an example the config parser checks if the participants of the exchange tags are part of the cplscheme and whether they hold the data, but doesn't ensure that all required data is transferred.
It checks "most of the local logic", if that makes sense.

pages/docs/tooling/tooling-builtin.md
fsimonis: Oh boy

pages/docs/tooling/tooling-builtin.md
fsimonis: Good point

pages/docs/tooling/tooling-builtin.md
fsimonis: I rewrote the section for clarity.

pages/docs/tooling/tooling-builtin.md
fsimonis: I added some clarifications

pages/docs/tooling/tooling-builtin.md
```diff
@@ -69,23 +71,47 @@ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
 
 The `check` runs the preCICE configuration parsing and checking logic on the given configuration file.
 This will find the majority of the configuration mistakes without having to start a simulation.
+These checks include wrong tags and attribute values, and more elaborate naming checks.
+More intricate logic, such as checks if all necessary data are exchanged in a Coupling Schemes, are not covered.
```
MakisH:
```suggestion
More advanced logic, such as checks if all necessary data are exchanged in a coupling scheme, are not covered.
```

(we should use rather simpler vocabulary in documentation, as the goal is clarity and the target audience mostly non-native speakers)

pages/docs/tooling/tooling-builtin.md
```diff
@@ -69,23 +71,47 @@ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
...
 
-The basic usage is to simply check a configuration file.
+The basic usage is to simply check a configuration file:
 
 ```bash
 binprecice check precice.xml
```
MakisH:
```suggestion
binprecice check precice-config.xml
```
(common name)

pages/docs/tooling/tooling-builtin.md
```diff
@@ -69,23 +71,47 @@ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
...
+  ```
+
+To enable more niche checks, additionally pass the name of one participant.
+This participant is assumed to run on a single rank.
+You may additionally pass the communicator size of the participant.
+This enables some checks regarding user-defined intra-participant communication, which should not be necessary in the vast majority of cases.
 
 ```bash
+binprecice check precice.xml Fluid
 binprecice check precice.xml Fluid 2
```
MakisH:
```suggestion
binprecice check precice-config.xml Fluid
binprecice check precice-config.xml Fluid 2
```

MakisH (approved):

PR #112, opened 2022-01-17
preCICE support program

chlorenz:
A first draft of how to structure different support options to support the sustained development of preCICE:
 1. As a research project partner
 2. Through direct cost/as a consumable --> preCICE support license
 3. As a (sub-)contractor

PR #119, opened 2022-02-03
Tweak support program content

uekerman:
Follow-up of #112 

* [x] For the main landing page, we need some visual element. Could simply be an icon of a hand shake or sth similar.
* [x] It would probably be good to also have a brief link on the community landing page.
* [X] I think it's enough if we have the "call for action" (contact Benjamin ...) only once at the bottom. Also, let's please drop the "Jun. Prof." there. grin


MakisH:
Maybe also change the

> Private, on-site support and bespoke training

on the home page to

> Private, on-site support and trailored training

(it's the first time I enounter "bespoke", I usually read "tailored" or just "personalized")

I will think about a visual element.

fsimonis:
I added the prototype image now.

Feel free to update and overwrite.

fsimonis:
@chlorenz I updated the picture `images/precice-support.svg` to be round and it includes a white background. You can use it in the landing page.

MakisH (approved):
Side note: I have fixed the unrelated formatting issues on `master`, feel free to rebase or directly merge.

fsimonis:
> it's the first time I enounter "bespoke", I usually read "tailored" or just "personalized"

bespoke is quite a common word for individualised services.

PR #129, opened 2022-02-14
Add distribution v2202.0

MakisH:
This explains the different versioning systems and adds the preCICE Distribution v2202.0.

I am not sure how to handle the commit of the website itself at the moment, I guess we should also have releases here, with the same versioning system as in the VM. @uekerman your opinion?

I assume we should also add DaRUS badges and maybe something regarding citing. (related: https://github.com/precice/precice/pull/1101)


MakisH:
After discussing with @uekerman, I added a version tag similar to the vm for the website (still to generate). I also added a badge for the v2104.0 and will add another one when v2202.0 is on DaRUS.

Here is the current preview:

Screenshot from 2022-02-17 13-44-04

uekerman (approved):
Looks good.

Next steps:
- Merge this PR
- Tag website and create documentation PDF
- Publish on DaRUS (only documentation PDF, not website)
- Add DaRUS badge

PR #131, opened 2022-02-17
Add general hints for tutorials

MakisH:
This add a few hints that are relevant for multiple tutorial cases. When one wants to visualize the results of, e.g., the multiple perpendicular flaps case, should not look into the perpendicular flap case for that.

Current state:

Screenshot from 2022-02-17 16-12-59

@uekerman or @DavidSCN or @IshaanDesai do you agree with the structure? Would you add anything else here?

### DIFF
--- a/pages/tutorials/tutorials.md
+++ b/pages/tutorials/tutorials.md
@@ -50,3 +50,38 @@ In the following cases, you can explore different aspects of preCICE:
 ## Community projects
 
 Apart from these simple tutorial cases, the community has tried preCICE in different setups. Check out our new [community projects](https://precice.discourse.group/c/community-projects/11) section in our forum, or read the [community stories](community-projects.html).
+
+## General hints
+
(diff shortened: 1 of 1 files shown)


davidscn (commented):
I generally agree on the hints. However, I'm not sure if this is the right place to put them. Solver specific stuff usually belongs into the solver sections. If the hints grow in the future the page might get a bit bloated.

MakisH:
> Solver specific stuff usually belongs into the solver sections.

I understand what you mean, but we don't really have "solver sections", we mostly have "adapter sections" (for some of which we only have one page). Most of these instructions would also apply to anyone using just the uncoupled solvers. An in the tutorials they are currently more visible.

> If the hints grow in the future the page might get a bit bloated.

I hope that it will not grow much more. If it does in the future, then we could also move it to a dedicated page.

Let's leave it open until the workshop and we may collect more ideas and opinions.

IshaanDesai (approved):
I agree that such hints are very helpful and oftentimes also necessary as I myself have sometimes struggled to understand how to visualize the results. The amount of content right now seems reasonable to be placed on the tutorials main page, but if at some point it becomes twice its size then it would be too much for the main page.

MakisH:
Then I am adding this already and we can get feedback during the workshop.

PR #135, opened 2022-02-22
Update link of fracture application testimonial

ajaust:
The linked paper in the testimonial "Hybrid-dimensional coupling for flow in deformable fractures" has been published. Since the paper is available as open-access article, the link was updated to point to the published paper instead of the preprint.

Additionally, the file permissions of `18-precice-mbdyn.md` have been changed (`chmod 644`) such that the file is not executable anymore.


davidscn (approved):
Thanks for taking care.

PR #138, opened 2022-02-23
Fix typo in config visualizer

arvedes:
Removed the `merged` in the code for the full picture.


MakisH (approved):
Indeed, the current command does not correspond to the picture. Thanks!

PR #151, opened 2022-03-14
Add adapter-X redirects to overview pages

### LINKED ISSUE #79: Reserved URLs for adapter pages
We currently have an inconsistency in the adapter documentation pages: some adapters have multiple pages (including an overview page), but other adapters only have one, as they are (currently) too short on documentation.

This is not future-proof and can lead to the following situation: one day we extend, e.g., the FEniCS adapter documentation with more pages, so the main page becomes an overview page, but the URL would need to remain https://precice.org/adapter-fenics.html, as it is linked from external pages and publications.

I propose the following: let's already reserve `https://precice.org/adapter-<X>.html` URLs for every adapter, which should redirect to e.g. `https://precice.org/adapter-<X>-overview.html`. This way, we can use the top-level `/adapter-<X>.html` (or even just `/adapter-<X>`) everywhere and have larger flexibility in changing the structure in the future.

To implement this, we would need to setup redirects in Jekyll. We already have some redirections in place from the old website URLs.

fsimonis:
This adds redirects for the pages `/adapter-X` for X in { calculix, deal.ii, su2 } adapters to their overview pages.

I also added common synonyms `deal.ii` and `ccx`.

Closes #79

### DIFF
--- a/pages/docs/adapters/calculix/adapter-calculix-overview.md
+++ b/pages/docs/adapters/calculix/adapter-calculix-overview.md
@@ -1,9 +1,7 @@
 ---
 title: The CalculiX adapter
 permalink: adapter-calculix-overview.html
-redirect_from:
-  - adapter-calculix.html
-  - adapter-ccx.html
+redirect_from: adapter-calculix.html
 keywords: adapter, calculix
 summary: "The CalculiX adapter can be used to couple CalculiX to CFD solvers for FSI or CHT application or even to couple CalculiX to itself."
 ---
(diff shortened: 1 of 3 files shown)


MakisH:
Nice, thanks!

Two points:
- Can we drop the file extension (`.html`)? I guess a redirect is better, the shorter it is.
- I would not have multiple redirects, because then we lose track. I think we should have only one per page and use it everywhere. In that context, I would go with the full name, to not have to guess every time.

fsimonis:
> Can we drop the file extension (.html)? I guess a redirect is better, the shorter it is.

The client-side redirection is a html file, but you can omit the html anyhow. (Something we should probably do with all internal links)

> I would not have multiple redirects, because then we lose track.

:+1: :building_construction:

PR #66, opened 2021-06-29
Add GitHub Action to check links

MakisH:
This follows the same system used in the tutorials and openfoam-adapter to check links.

In order to avoid getting too many false negatives, I am ignoring status 429 ("too many requests") and 200 ("OK").

In later commits in this PR, I will also fix currently broken links. ~~Related to~~ Closes https://github.com/precice/precice.github.io/issues/62.

See also https://github.com/precice/tutorials/pull/206 and https://github.com/precice/openfoam-adapter/pull/169.

PR #154, opened 2022-03-15
Check links new

Scriptkiddi:
So this is a little bit awkward, I noticed the branch and the issue from @MakisH when git told me the branch `check-links` already existed... 
anyhow i figured I would create this pr even when it might be a duplicate so the work does go to waste.
related #66


Scriptkiddi:
@MakisH could you have a look

MakisH:
Thank you for pushing this forward! I don't want to delay it and we should at least already merge the fixed links.

But regarding the action:
- Question: Why did you choose `html-proofer` instead of one of the already available GitHub Actions (e.g., [markdown-link-check](https://github.com/tcort/markdown-link-check/)? Is there some advantage?
- I see that in the action you are adding, you also get a few false-negatives, but I actually don't see it reporting as many issues as markdown-link-check (#66). Could you compare if the two solutions find the same true-positives?

Scriptkiddi:
I googled link checker for jekyll and found html proofer besides that there is no real reason to it.
After checking out the markdown check links action, I'm going to switch that over from html proofer. The documentation over there looks a lot better.

I disabled the external link checking which you still have enabled, to produce consistent results, otherwise a pr could fail without the responsible person touching the broken link.

MakisH:
But if you want to change to markdown-link-check, then feel free to directly contribute to #66. I gave you write access to the repository.

MakisH:
I included these broken link fixes in #66. We can rethink a better system in a separate PR.

Thanks again for contributing, it was already helpful discovering these links!

PR #156, opened 2022-03-18
Community Story for the preCICE page

JulianSchliessus:


MakisH (approved):
Nice, thank you very much, @JulianSchliessus! :hugs:

MakisH (changes requested):
Actually, it looks like it does not render correctly at the moment:

Screenshot from 2022-03-22 14-53-58

I guess one issue is that we don't support multiple authors. @chlorenz is this something we could easily add? In any case, @JulianSchliessus see how the `Coupling mechanics and electrophysiology in skeletal muscles` is implemented.

The image also seems to not work. Is it referenced?

JulianSchliessus:
Thanks for the information :grinning:.
I have changed the filename of the image and merged the author names into one block.

MakisH (approved):
Now it looks fine, thanks a lot for the quick reaction!

Issue #161, opened 2022-03-23
Sort adapters alphabetically in adapter overview


ajaust:
Currently the sorting of adapters on the [adapter overview page](https://precice.org/adapters-overview.html) is rather random. Could we maybe sort them alphabetically? At the moment the page feels a bit "chaotic" when navigating as the list is quite long nowadays.

The list of [legacy adapter](https://precice.org/adapters-overview.html#legacy-adapters) already seems to be sorted alphabetically.

ajaust:
Closed via #162

PR #162, opened 2022-03-23
Sort adapter overview alphabetically

ajaust:
This sorts the lists of official and third-party adapters
alphabetically.

Solves #161

### DIFF
--- a/pages/docs/adapters/adapter-overview.md
+++ b/pages/docs/adapters/adapter-overview.md
@@ -12,15 +12,15 @@ There are various codes - free and proprietary ones - currently coupled with pre
 We host adapters for the following codes in the [preCICE GitHub organization](https://github.com/precice/) and we maintain them to work with the latest release of preCICE (unless stated otherwise).
 
 | Adapter for | Contact | Resources | Typical applications | Comments |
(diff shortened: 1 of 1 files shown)


uekerman (approved):
I like. Thanks!
Could you please rebase?

ajaust:
I just rebased it.

MakisH:
Thanks for contributing! Previously, it was indeed pseudo-randomly sorted, with the most actively maintained adapters on top.

Two comments:
- Now, the first adapter one sees is "ANSYS Fluent", soon followed by "COMSOL Multiphysics". We should really move them to the legacy list, because this could lead to quite some false impressions and potential frustration by users.
- Not touched by this PR but seeing this now: why "(legacy) FEniCS"? Is it already named like that everywhere? I understand that FEniCS-X is not ready and marking this "legacy" gives again a false impression (potentially for the adapter).

ajaust:
I considered "(legacy) FEniCS" still as "FEniCS" where the "legacy" is more an additional hint that it is not FEniCS-X. I agree that it is not optimal especially since the FEniCS adapter is maintened. Maybe one should remove the "legacy" and rather add a note in the comment that this refers to the "old" FEniCS versions or refer to the FEniCS-X adapter below.

MakisH:
Now I look at it again, it is fine as it is now. :-)

PR #164, opened 2022-03-29
Update installation prefix documentation

ajaust:
We had some people that were confused by parts of the documentation when installing preCICE from source (especially on MacOS). Therefore, this PR proposes the following changes.

- Moves the notes on setting the environment variables for
individual installation prefixes to the "Installation" section.
- Adds a note on the different naming of some environment
variables.
- Adds PATH environment variable to be set since `binprecice` contains
more useful features now or in the near future.


fsimonis:
Thanks! I can see how this makes the installation doc easier to understand.

fsimonis (approved):

PR #174, opened 2022-06-14
Update WR documentation for serial coupling

BenjaminRodenberg:
This PR provides documentation related to the feature https://github.com/precice/precice/pull/1029.

# Todo

- [x] Merge https://github.com/precice/precice.github.io/pull/186


pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -8,9 +8,6 @@ summary: "With waveform iteration, you can interpolate coupling data in time for
 {% experimental %}
 These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care.
 {% endexperimental %}
-{% note %}
-This feature is only available for parallel implicit coupling. An extension to serial implicit coupling is planned. Without loss of generality, we moreover only discuss the API functions `readBlockVectorData` and `writeBlockVectorData` in the examples.
-{% endnote %}
```
uekerman: Somewhere we should still mention that this is only for implicit coupling.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -97,6 +94,12 @@ while (not simulationDone()){ // time loop
 ...
 ```
 
+## Initialize Data
+
+* When using waveform iteration both participants can initialize data, even when serial implicit coupling is used.
+* Reason for this: Initialized data is used at the beginning of the window. First write data at the end of the window. This means: Also first participant can initialize data that will be used by second participant at the beginning of the window. The the same time there is already data available at the end of the window in the first iteration of a serial implict coupling scheme.
+* If no initial data is provided, the data at the beginning of the window is **always** implicitly set to zero and there is no possibility to change this.
```
```suggestion
## Initialize Data

* When using waveform iteration both participants can initialize data, even when serial implicit coupling is used.
* Reason for this: Initialized data is used at the beginning of the window. First write data at the end of the window. This means: Also first participant can initialize data that will be used by second participant at the beginning of the window. The the same time there is already data available at the end of the window in the first iteration of a serial implict coupling scheme.
* If no initial data is provided, the data at the beginning of the window is **always** implicitly set to zero and there is no possibility to change this.
```
```suggestion
## Initialize data

* When using waveform iteration, both participants can initialize data, even when serial-implicit coupling is used.
* Reason for this: Initialized data is used at the beginning of the window. First write data at the end of the window. Thus, the first participant can initialize data that is used by the second participant at the beginning of the window.
* If no initial data is provided, the data at the beginning of the window is **always** implicitly set to zero.
```
The sentence "The the same time there is already data available ...", I did not understand.

Could we add figures here? :grin:

```suggestion
## Initialize Data

* When using waveform iteration both participants can initialize data, even when serial implicit coupling is used.
* Reason for this: Initialized data is used at the beginning of the window. First write data at the end of the window. This means: Also first participant can initialize data that will be used by second participant at the beginning of the window. The the same time there is already data available at the end of the window in the first iteration of a serial implict coupling scheme.
* If no initial data is provided, the data at the beginning of the window is **always** implicitly set to zero and there is no possibility to change this.
```
```suggestion
## Initialize data

* When using waveform iteration, both participants can initialize data, even when serial-implicit coupling is used.
* Reason for this: Initialized data is used at the beginning of the window. First write data at the end of the window. Thus, the first participant can initialize data that is used by the second participant at the beginning of the window.
* If no initial data is provided, the data at the beginning of the window is **always** implicitly set to zero.
```
The sentence "The the same time there is already data available ...", I did not understand.

Could we add figures here? :grin:

BenjaminRodenberg: I will try to come up with a nice figure :+1:

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I modified the note.

BenjaminRodenberg:
Since we decided to removed `initializeData` in https://github.com/precice/precice/pull/1350 I think it might be better to revise https://precice.org/couple-your-code-initializing-coupling-data.html. This might also simplify what we have to explain here. If we drop the restriction that a serial coupling scheme cannot initialize data for both participants, we do not have to worry about this anymore. I will also open an issue in `precice/precice`.

BenjaminRodenberg:
I added an issue on the topic of data initialization: https://github.com/precice/precice/issues/1367

BenjaminRodenberg:
I also updated the documentation on data initialization: https://github.com/precice/precice.github.io/pull/186

BenjaminRodenberg:
Closing this PR, because most changes are on `precice-v3` already and rest is trivial.

PR #175, opened 2022-06-24
Add page for preCICE Workshop 2023

MakisH:
This adds a first placeholder for the workshop page, so that we have something to link to.

@uekerman / @fsimonis: one of you please do a quick sanity check and merge.


fsimonis (commented):
Let's merge and see what happens to the CI

davidscn:
Some comments:

- last time, we moved the old workshop page to the archive
- considering that we have the date fixed, you could add (replace) the banner on the community page

MakisH:
Thanks! I thought about these, but, since the 2022 workshop page is still much more useful to get an idea of what that is, I wanted to only swap them later (in August/September). This will essentially be the half a year ahead.

PR #176, opened 2022-06-27
Add comments on MPI ports vs sockets

MakisH:
It looks like users do not really understand the issue with MPI ports, and they do not understand how much of a compromise sockets are.
This adds a reference to Benjamin's dissertation (scaling results), but we may need to add even more information.

@uekerman what else could we improve here?


pages/docs/configuration/configuration-communication.md
```diff
@@ -44,11 +44,11 @@ The alternative to TCP/IP sockets is MPI ports (an MPI 2.0 feature):
 <m2n:mpi .../>
 ```
 
-As the ports functionality is not a highly used feature of MPI, it has robustness issues for several MPI implementations ([for OpenMPI, for example](https://github.com/precice/precice/issues/746)). In principle, MPI gives you faster communication roughly by a factor of 10, but, for most applications, you will not feel any difference as both are very fast. We recommend using `sockets`.
+As the ports functionality is not a highly used feature of MPI, it has robustness issues for several MPI implementations ([for OpenMPI, for example](https://github.com/precice/precice/issues/746)). In principle, MPI gives you faster communication roughly by a factor of 10 (see [Benjamin Uekermann's dissertation](https://mediatum.ub.tum.de/doc/1320661/document.pdf), section 4.2.3), but, for most applications, you will not feel any difference as both are very fast. We recommend using `sockets` by default, unless you are performing a large performance-critical simulation.
```
uekerman:
```suggestion
As the ports functionality is not a highly used feature of MPI (if used with separated `MPI_COMM_WORLD` communicators), it has robustness issues for several MPI implementations ([for OpenMPI, for example](https://github.com/precice/precice/issues/746)). In principle, MPI gives you faster communication roughly by a factor of 10 (see [Benjamin Uekermann's dissertation](https://mediatum.ub.tum.de/doc/1320661/document.pdf), section 4.2.3), but, for most applications, you will not feel any difference as both are very fast. We recommend using `sockets` by default, unless you are performing large performance-critical simulations with very large coupling meshes.
```

uekerman (approved):
Looks good. I added a bit more.

pages/docs/configuration/configuration-communication.md
MakisH: I think the "separated `MPI_COMM_WORLD` communicators" will still not be clear for many users. Can we maybe give an example or comment how probable it is that it affects a user?

uekerman: preCICE participants are typically always started with separated `MPI_COMM_WORLDS`. This first sentence here is a bit to blame MPI.
Maybe linking to https://precice.org/running-slurm.html helps?

Our main message should however always be: "use sockets, it's not really slower, but much more robust".

MakisH:
@uekerman I added a few more clarifications, based on our discussion.

Scriptkiddi:
would it be possible to add a quantification for `very large coupling meshes`?
like `very large coupling meshes ( > 1000000 vertices)` to help people decide quickly in which category they fall? Because I would guess different fields have different understandings of what a very large mesh is.

fsimonis:
@Scriptkiddi The basic guideline is to use sockets.

The size of the communicated mesh depend on many factors. Vertex count is only a part of the story. Mesh connectivity, defined mappings, defined and exchanged data, watch integrals and filtering strategies all impact this communicated size.

You will need to use [the inbuilt profiling](https://precice.org/tooling-performance-analysis), or you do your own.

The general guideline for the inbuilt profiling (which is missing in the docs) is:
1. Start using sockets
2. Enable sync mode in the configuration.
3. Run your full testcase for a few time steps.
4. Establish your baselines:
   * For the mesh transfer, checkout the events `partition.sendGlobalMesh` in initialize
   * For the data transfer, `m2n.sendData` and `m2n.receiveData` events in advance.
5. Figure out if these impact the overall simulation time. If not, then you are done.
8. Change to `<m2n:mpi >` and rerun the case.
10. Compare the events in 4 to your established baseline and select the preferred method.

uekerman:
> would it be possible to add a quantification for very large coupling meshes?

it mainly depends on how this relates to the computational cost of your solver. good point, but hard to quantify :confused:

pages/docs/configuration/configuration-communication.md
```diff
@@ -44,11 +44,11 @@ The alternative to TCP/IP sockets is MPI ports (an MPI 2.0 feature):
...
+In preCICE, we always start simulations in separated MPI communicators (remember: we start solvers in different terminals, with their own `mpirun` commands), a feature that highly improves flexibility (solvers do not need to be in the same MPI communicator at any time). As the MPI ports functionality is not a highly used feature of MPI (at least not with separated `MPI_COMM_WORLD` communicators), it has robustness issues for several MPI implementations ([for OpenMPI, for example](https://github.com/precice/precice/issues/746)). In principle, MPI gives you faster communication roughly by a factor of 10 (see [Benjamin Uekermann's dissertation](https://mediatum.ub.tum.de/doc/1320661/document.pdf), section 4.2.3), but, for most applications, you will not feel any difference as both are very fast. We recommend using `sockets` by default, unless you are performing large performance-critical simulations with very large coupling meshes.
```
fsimonis:
```suggestion
In preCICE, we always start simulations in separated MPI communicators (remember: we start solvers in different terminals, with their own `mpirun` commands), a feature that highly improves flexibility (solvers do not need to be in the same MPI communicator at any time). As the MPI ports functionality is not a highly used feature of MPI (at least not with separated `MPI_COMM_WORLD` communicators), it has robustness issues for several MPI implementations ([for OpenMPI, for example](https://github.com/precice/precice/issues/746)). In principle, MPI gives you faster communication roughly by a factor of 10 (see [Benjamin Uekermann's dissertation](https://mediatum.ub.tum.de/doc/1320661/document.pdf), section 4.2.3), but, for most applications, you will not feel any difference as both are very fast. We recommend using `sockets` by default, unless you are performing large performance-critical simulations with very large coupling meshes combined with a high total time-window count.
There are no hard limits regarding vertex or time-window count for when this change makes sense, as the the preCICE configuration and the solver-provided mesh partitioning hugely impact the required communication load. The only reliable decision process is to measure the impact of the change on the run-time of `initialize` and `advance` calls. See the section on [performance analysis](tooling-performance-analysis) for more information.
```

PR #179, opened 2022-07-04
Update system guides

MakisH:
This mainly updates the Ubuntu version from 20.04 (focal) to 22.04 (jammy), since this is now getting popular.

It also removes the (now unsupported by Canonical) 16.04 and adds a warning next to 18.04 that it was supported till preCICE v2.3.0.

Same for the CalculiX adapter page (we provide packages for that and we have a Jammy package since today).

TODO (not in this PR if a lot of work):
- [x] Check Fedora 36 (I can check)
- [x] Check which CentOS versions we still support
- [x] Check OpenSUSE Leap 15.4
- [x] Check if we still support macOS Catalina
- [x] Check newer macOS versions

When checking against older versions, main things to check are the Boost version (min 1.71.0) and CMake version (min 3.16.3).

@boris-martin did I forget to update anything? If you feel like checking any of the open points, please write here, to avoid duplicate work.


pages/docs/installation/installation-source-dependencies.md
```diff
@@ -262,6 +262,10 @@ If you want build preCICE on your own computer and you are using one of the foll
 
 Other modern versions of popular Linux distributions are also perfectly compatible, here we just list a few popular options. Since our users have tried preCICE on various distributions, you may as well ask on our [forum](https://precice.discourse.group/) for any questions.
 
+### Ubuntu 22.04 LTS Jammy Jellyfish
+
+Follow the instructions as for Ubuntu 20.04.
```
boris-martin: It seems a bit weird to have 22.04 described as "same as 20.04" instead of the other way around it.

pages/docs/installation/installation-packages.md
```diff
@@ -12,18 +12,18 @@ This section lists systems and instructions on how to install these packages.
...
 
-Alternatively, download & install it from the command line. For **Ubuntu 20.04 (focal)**:
+Alternatively, download & install it from the command line. For **Ubuntu 22.04 (jammy)**:
 
 ```bash
-wget https://github.com/precice/precice/releases/download/v{{ site.precice_version }}/libprecice2_{{ site.precice_version }}_focal.deb
-sudo apt install ./libprecice2_{{ site.precice_version }}_focal.deb
+wget https://github.com/precice/precice/releases/download/v{{ site.precice_version }}/libprecice2_{{ site.precice_version }}_jammy.deb
+sudo apt install ./libprecice2_{{ site.precice_version }}_jammy.deb
 ```
```
boris-martin, line 20: I'm starting to wonder if adding a website variable "name of the latest Ubuntu" would be interesting, or an unmaintainable overkill.

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
```diff
@@ -13,15 +13,19 @@ There are two ways to get the adapter: (a) get a binary package (Ubuntu-only), o
...
 
-Alternatively, download & install it from the command line. For Ubuntu 20.04 (focal):
+Alternatively, download & install it from the command line. For Ubuntu 22.04 (jammy):
 
 ```bash
-wget https://github.com/precice/calculix-adapter/releases/download/v{{ site.calculix_adapter_version }}/calculix-precice2_{{ site.calculix_adapter_version }}-1_amd64_focal.deb
-sudo apt install ./calculix-precice2_{{ site.calculix_adapter_version }}-1_amd64_focal.deb
+wget https://github.com/precice/calculix-adapter/releases/download/v{{ site.calculix_adapter_version }}/calculix-precice2_{{ site.calculix_adapter_version }}-1_amd64_jammy.deb
+sudo apt install ./calculix-precice2_{{ site.calculix_adapter_version }}-1_amd64_jammy.deb
 ```
```
boris-martin, line 21: Same as above

boris-martin (approved):
Current changes look good. I don't really now how to check other OSes properly but I can have a look to list pages that could require an update.

boris-martin:
I made checks on OpenSUSE and managed to configure the build with MPI and PETSc disabled.
**Required changes**
- Use Boost 1.75 instead of 1.66 (So in package names the 1_66 must become 1_75).
- Numpy must be installed. I used pip install but I assume a zypper package exists too.
- Change 15.2 into 15.4 in the title

~~However I couldn't get it to compile and have no idea why (it's a fresh VM and I'm quite unfamiliar with OpenSUSE)~~
Build successful, I just didn't give enough RAM to the VM

MakisH:
> I made checks on OpenSUSE and managed to configure the build with MPI and PETSc disabled. **Required changes**
> 
>     * Use Boost 1.75 instead of 1.66 (So in package names the 1_66 must become 1_75).
> 
>     * Numpy must be installed. I used pip install but I assume a zypper package exists too.
> 
>     * Change 15.2 into 15.4 in the title
> 
> 
> ~However I couldn't get it to compile and have no idea why (it's a fresh VM and I'm quite unfamiliar with OpenSUSE)~ Build successful, I just didn't give enough RAM to the VM

Nice! Can you please push your changes for OpenSUSE here, as well? I guess we can remove the 15.2, then.

pages/docs/installation/installation-packages.md
MakisH: I had similar feelings, both for the Ubuntu and for the OpenFOAM version. Some weak arguments in favor and against:

- In favor: No need to update many files just to change the version
- Against: Such files (especially the `README.md` of the Quickstart) will not make much sense if read locally and not rendered on the website.
- Against: This is also a regular reminder to check for other outdated parts, such as removing old Ubuntu versions, or updating the preCICE version.

boris-martin:
I tried with the zypper package for Numpy and it didn't work, so I pushed instructions recommending pip.

MakisH:
> I tried with the zypper package for Numpy and it didn't work, so I pushed instructions recommending pip.

Thanks! No `pip install --user`? We usually avoid pointing to system-wide installs for pip.
Also, no `pip3`?

boris-martin:
Right, --user should probably used. Bad practice coming from my typical usage.
For pip3, I guess so. pip alone works, but I tried on a VM with only python3 and no python2 so the risk of conflcit is low. Better safe than sorry though.

boris-martin:
I tested Fedora an instructions seem up to data, except that I the module command doesn't seem installed by default. Successful configuration of preCICE with MPI and PETSc disabled otherwise.
(If someone wants to debug the module issue it'd be nice, my computer seems to really hate Fedora which crashes every 5 mins on my VM)

MakisH:
> I tested Fedora an instructions seem up to data, except that I the module command doesn't seem installed by default.

I tried on a rather new Fedora installation on my laptop. I think that you need to log out and in again before `module` becomes available. Can you please confirm?

Added the respective documentation.

MakisH:
@fsimonis on Fedora 36, CMake behaves weirdly regarding PETSc. Does this log look familiar?

[precice-cmake.log](https://github.com/precice/precice.github.io/files/9047395/precice-cmake.log)

boris-martin:
> > I tested Fedora an instructions seem up to data, except that I the module command doesn't seem installed by default.
> 
> I tried on a rather new Fedora installation on my laptop. I think that you need to log out and in again before `module` becomes available. Can you please confirm?
> 
> Added the respective documentation.

Rebooted a couple of times and still no sign of it. I wouldn't be surprised if my Fedora installation was broken though

fsimonis:
@MakisH looks like it can find PETSc, but cannot resolve hdf5 as a dependency.

fsimonis:
@MakisH Just checked the Fedora 64 issue again. The PETSc detection requires installing `hdf5-openmp-devel` in addition to the automatically installed `hdf5-openmpi`.
Currently, `dnf install hdf5-openmpi-devel` is the solution.

All tests succeed in the docker container.

pages/docs/installation/installation-source-dependencies.md
MakisH: Good point, I swapped them. Thanks!

pages/docs/installation/installation-packages.md
MakisH: Let's not implement this with this PR and reconsider it if updating becomes a burden or if we continuously run into inconsistencies.

pages/docs/adapters/calculix/adapter-calculix-get-adapter.md
MakisH: Replied above

MakisH:
Regarding macOS:
- @solids4foam could build and run on macOS 11.4 via homebrew ([details](https://github.com/precice/tutorials/pull/286#issuecomment-1206491298))
- Our CI builds and tests preCICE on `macos-latest`, which currently is macOS 11.6.8 ([details](https://github.com/precice/precice/actions/workflows/build-and-test-mac.yml)). It installs the same packages from Homebrew.

I added a note regarding versions, asking for contributions.

MakisH:
Regarding CentOS, I understand that we now don't support either CentOS 7 or CentOS 8, since they both provide very old Boost packages (while we need 1.71 since preCICE v2.4.0).

I added a warning that the last supported version there was preCICE v2.3.0. I will eventually check if Rocky 9 works.

MakisH:
Short dump of updates on Rocky 9:
- It provides Boost 1.75: https://pkgs.org/search/?q=boost-devel
- It has dropped the `powertools` repository and introduced the `crb` repository: https://wiki.rockylinux.org/rocky/repo/
- `eigen3-devel` seems to only be available from `crb`: https://pkgs.org/search/?q=eigen3-devel
- Explicitly installing numpy from pip3 does not seem to be needed anymore.
- PETSc does not seems to be available at all in Rocky: https://pkgs.org/search/?q=petsc (neither with `petsc` and derivatives, nor with `libpetsc`)

With disabled PETSc, preCICE builds successfully, and all tests run.

boris-martin:
Just for my culture: what is the policy on what systems are worth a specific section? Is Rocky added because it's the future of CentOS?

MakisH:
> Just for my culture: what is the policy on what systems are worth a specific section? Is Rocky added because it's the future of CentOS?

Exactly. CentOS switched to CentOS Stream, which some administrators seem to be upset about. This led to the Rocky Linux fork, which seems to be turning into a major server distribution. So, you can read this as if it was "CentOS 9".

PR #186, opened 2022-07-16
Update documentation for initialize

BenjaminRodenberg:
Follows changes in implementation introduced in https://github.com/precice/precice/pull/1368.

# Todo

- [x] merge https://github.com/precice/precice.github.io/pull/254


pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
```diff
@@ -5,19 +5,17 @@ keywords: api, adapter, initialization, coupling scheme, restart
 summary: "As default values, preCICE assumes that all coupling variables are zero initially. For fluid-structure interaction, for example, this means that the structure is in its reference state. Sometimes, you want to change this behavior – for instance, you may want to restart your simulation."
 ---
 
-For initializing coupling data, you can add the following **optional** method:
+{% version 3.0.0 %}
+`initializeData()` will be removed in preCICE version 3.0.0. You can do data initialization now by calling `initialize()`.
```
uekerman:
```suggestion
`initializeData()` only exists for preCICE versions < 3.0. For versions >= 3.0, data initialization is directly handled in `initialize()`.
```

uekerman (commented):
I am not sure yet, what the best option for the page `couple-your-code-initializing-coupling-data` is.

Options:

* Explain both variants (old and new)
* Remove page completely
* Explain only new version (but is there much to explain?) 

We should also write a porting guide, to which we should link from here.

BenjaminRodenberg:
> I am not sure yet, what the best option for the page `couple-your-code-initializing-coupling-data` is.
> 
> Options:
> 
>     * Explain both variants (old and new)
I would prefer only new, because this is less confusing and less work. If somebody is interested in the old version, one can refer to the old version of the docs.
> 
>     * Remove page completely
I would keep it (see below).
> 
>     * Explain only new version (but is there much to explain?)
There are some features that need explanation: 1) we need a new action. 2) we need to write data. 3) we have to specify this in the configuration. I'm not really sure where we could explain this instead without making another step in the step-by-step guide too complicated. And: data initialization is important. I think we should also give it some visibility.
> 
> We should also write a porting guide, to which we should link from here.

Agree. I'm currently a bit short on time. I guess the old porting guide (see https://precice.org/couple-your-code-porting-adapters.html) will act as a template here. Quick and dirty porting guide from my side:

1. move `initialize` to the place where `initializeData` is called.
2. remove calls of `initializeData`.

The diff of the tests in https://github.com/precice/precice/pull/1350/files#diff-4032663f1dad054c37ca6b17856ee821140a58a05378aae3e219adf266436be6 shows this workflow quite nicely.

BenjaminRodenberg:
This PR has been lying around for quite long. @uekerman I would suggest to merge it. We can still improve it later depending on user feedback, but I think the most important changes to keep the documentation up-to-date with version 3 are here.

Note: Porting guide has been extended already a while ago.

uekerman:
I am not sure we should merge already. So far, we did not break the documentation already. I would actually only break it once we release v3.

BenjaminRodenberg:
@uekerman should be ready to merge now that `precice-v3` exists. Right?

pages/docs/configuration/configuration-action.md
uekerman, line 1: I would recommend to first merge #254 
Then, the changes to this page will be obsolete.

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
```diff
@@ -5,19 +5,17 @@ keywords: api, adapter, initialization, coupling scheme, restart
...
-For initializing coupling data, you can add the following **optional** method:
+{% version 3.0.0 %}
+`initializeData()` only exists for preCICE versions < 3.0. For versions >= 3.0, data initialization is directly handled in `initialize()`.
+{% endversion %}
+
+By default preCICE assumes that all coupling variables are zero initially. If you want to provide non-zero initial values, you can write data before calling `initialize()`. This data will then be used as initial data. The preCICE action interface that was introduced in [Step 6](couple-your-code-implicit-coupling) provides the following action to check whether initial data has to be written by a participant:
 
 ```cpp
-void initializeData();
+const std::string& constants::actionWriteInitialData()
```
uekerman: Needs updates. "action interface" is gone.

uekerman (changes requested):

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
BenjaminRodenberg: Done cdd2062 : **Important** we still need to update the main section about the actions. I don't want to update all the action things in this PR, because this is unrelated.

BenjaminRodenberg:
I also updated the initialization docs here w.r.t the removal of mesh and data IDs.

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
```diff
@@ -5,45 +5,27 @@ keywords: api, adapter, initialization, coupling scheme, restart
...
-[...]
-
-const std::string& cowid = precice::constants::actionWriteInitialData();
-
-[...]
-
-int displID = precice.getDataID("Displacements", meshID);
-int forceID = precice.getDataID("Forces", meshID);
-double* forces = new double[vertexSize*dim];
-double* displacements = new double[vertexSize*dim];
```
BenjaminRodenberg, line 32: This removal does not really belong to this documentation update, because it is triggered by the update of mesh and data ids. But I did it here anyway.

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
```diff
@@ -5,45 +5,27 @@ keywords: api, adapter, initialization, coupling scheme, restart
 summary: "As default values, preCICE assumes that all coupling variables are zero initially. For fluid-structure interaction, for example, this means that the structure is in its reference state. Sometimes, you want to change this behavior – for instance, you may want to restart your simulation."
 ---
 
-For initializing coupling data, you can add the following **optional** method:
+By default preCICE assumes that all coupling variables are zero initially. If you want to provide non-zero initial values, you can write data before calling `initialize()`. This data will then be used as initial data. To check whether initial data is required, you can use the following function:
 
 ```cpp
-void initializeData();
+bool requiresInitialData()
```
BenjaminRodenberg, line 11: This removal does not really belong to this documentation update, because it is triggered by the update of the action interface. But I did it here anyway.

BenjaminRodenberg (commented):
We are currently missing documentation updates for changes this PR depends on. I'm providing a review to give an overview:

pages/docs/configuration/configuration-action.md
uekerman: Merged, please rebase

BenjaminRodenberg: Done.

_includes/xmlreference.md
uekerman, line 1: no need to change things here, this will happen automatically

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
```diff
@@ -54,3 +36,9 @@ Now, you can specify at runtime if you want to initialize coupling data. For exa
 <exchange data="Displacements" mesh="StructureMesh" from="SolidSolver" to="FluidSolver" initialize="yes"/>
 [...]
 ```
+
+If you are using a serial coupling scheme, only initial data that is written from the second to the first participant is actually used. Initial data written from the first to the second participant will be ignored, because in a serial coupling scheme the first participant already computes its first result before the second participant even starts.
+
+{% note %}
+preCICE still supports data initialization for both participants, even if a serial coupling scheme is used. You can read our section on [time interpolation](couple-your-code-waveform) to learn more about the reasons.
+{% endnote %}
```
uekerman: I had to read this part three times before understanding it. Could we improve it?
Instead of "written from A to B", I would always say "written to A and then communicated by preCICE from A to B".
Either all in a note or non. Remember that waveform will be the norm with v3 eventually.

uekerman (changes requested):
Sry for playing this back another time.

_includes/xmlreference.md
BenjaminRodenberg: Yes. I think merging also cleaned this up.

pages/docs/couple-your-code/couple-your-code-initializing-coupling-data.md
BenjaminRodenberg: No so sure about this whole paragraph anymore, because it makes things more complicated and actually we are now planning to use initial data for every participant for every coupling scheme. Also the second participant in a serial coupling scheme uses the initial data it receives from the first. Even for waveform order = 0 we use the initial data for the relative time = 0. This is a very narrow edge case, but it leads to a consistent treatment across windows. See also the following figure:

Screenshot from 2023-05-28 14-11-10

How about removing this paragraph and opening the whole topic later when we introduce time interpolation?

BenjaminRodenberg: We could also use the figure from above later in the waveform relaxation section, if it is useful. For the section on data initialization, I think it contains too much information that has not been introduced yet.

BenjaminRodenberg: See [ec217b5](https://github.com/precice/precice.github.io/pull/186/commits/ec217b58ee30129a1d940147c7b2299b934af163) for some steps into this direction.

_includes/xmlreference.md
uekerman, line 1: No need to modify this page as it is auto-generated.

uekerman (approved):
Let's try to move forward here. We, anyway, should have a closer look at all of these pages before releasing.
Please remove the edits in the `xmlreference` and merge.

_includes/xmlreference.md
BenjaminRodenberg: Done. [c27eda8](https://github.com/precice/precice.github.io/pull/186/commits/c27eda8159aa976b8d2e2154eda0044b4ec5273c)

PR #191, opened 2022-08-02
Update clang-tidy docs

davidscn:
Related to precice/precice#999

### DIFF
--- a/pages/docs/dev-docs/dev-tooling.md
+++ b/pages/docs/dev-docs/dev-tooling.md
@@ -72,7 +72,7 @@ cd path/to/precice
 tools/linting/run_clang_tidy.sh
 ```
 
-which will report potential errors on the console. Executing the script can be done using a `make` target called `make tidy` as well. Some errors can be fixed by `clang-tidy` itself. In order to let `clang-tidy` fix errors, add the `-fix` option [in the script](https://github.com/precice/precice/blob/f84a41da2ea4c5b2ae8cc8fe4b38bde707a1c4c1/tools/linting/run_clang_tidy.sh#L50-L52). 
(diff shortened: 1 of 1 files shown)


pages/docs/dev-docs/dev-tooling.md
```diff
@@ -62,26 +62,17 @@ void formatted_code_yet_again;
...
+cd path/to/precice
+tools/linting/run_clang_tidy.sh
+```
 
-- Inspect a single file:
-  `clang-tidy -p ~/tmp/precice/ src/precice/impl/SolverInterfaceImpl.cpp` ( -p sets the dir where to find the file compile_commands.json)
-- Inspect the entire source tree or apply fixes:  
-  Use the `run-clang-tidy.py` to prevent header overlap etc.
-  Installed as `/usr/share/clang/run-clang-tidy.py` or from the [mirror](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py).
+which will report potential errors on the console. Executing the script can be done using a `make` target called `make tidy` as well. Some errors can be fixed by `clang-tidy` itself. In order to let `clang-tidy` fix errors, add the `-fix` option [in the script](https://github.com/precice/precice/blob/f84a41da2ea4c5b2ae8cc8fe4b38bde707a1c4c1/tools/linting/run_clang_tidy.sh#L50-L52).
```
MakisH: Do we really need to refer to a specific commit and lines here? Just referring to the file should be enough (but maybe even that is not needed).

MakisH (approved):

pages/docs/dev-docs/dev-tooling.md
davidscn: Removed it.

PR #193, opened 2022-08-08
Clarify FEniCS legacy

MakisH:
I spotted at least one person that got confused by our `(legacy) FEniCS` statement, thinking that our adapter is now legacy, unmaintained software (which is not the case, is it?).

As a solution, I removed the "(legacy)" part and added a comment pointing to FEniCS-X. This should give the same information, without sounding in any way abandoned.

@IshaanDesai / @BenjaminRodenberg please check my understanding and merge.


IshaanDesai (approved):
Makes sense to not say **legacy** FEniCS right in the adapter overview page and that too right at the start.

PR #196, opened 2022-08-09
Update link to DuMuX adapter repository

ajaust:
This updates the link of the DuMuX adapter repository from the previously used IWS
GitLab repository to the new repository on GitHub.

Do we want to change the maintainer as well?


MakisH (approved):
Thanks! As long as the current maintainer is still responding on GitHub, I don't think we need to change anything. :smile:

PR #197, opened 2022-08-10
Added "Flow over heated plate with two meshes" tutorial to doc.

boris-martin:
Updated the tutorial list & sidebar.
(Should be merged after that tutorial gets on the master branch of tutorials)


_data/sidebars/tutorial_sidebar.yml
```diff
@@ -49,6 +49,10 @@ entries:
     - title: Flow over heated plate steady state
       url: /tutorials-flow-over-heated-plate-steady-state.html
       output: web
+    
+    - title: Flow over heated plate with two meshes
```
MakisH: Shorter:

```suggestion
    - title: Flow over heated plate (two meshes)
```

But only it we also apply the same style to the "steady state" and "nearest projection" tutorials, actually.

MakisH (approved):
Nice, thanks!

I was wondering a bit about the order, but I guess the "nearest projection" and the "two meshes" are similarly complicated.

Please "squash and merge".

_data/sidebars/tutorial_sidebar.yml
boris-martin: Probably best yes, "Flow over heated plate steady state" without parentheses is ugly.

boris-martin:
My intuition would be that two meshes is more complex than NP but simpler ("less different") than steady state. But do we want ascending order? Descending ?

(Yes I'll squash. No idea how all these old commits came up here)

MakisH (approved):
Looks good as it is! :+1:

PR #202, opened 2022-08-16
Add the Julia bindings to the architecture picture

MakisH:
Also add missing source file for the image.

I assumed that `PreCICE.jl` is directly linking to `libprecice.so`, and that a Julia solver is including (importing) `PreCICE.jl`.

@IshaanDesai / @erikscheurer please check and merge if my undersdanding looks ok.


IshaanDesai (approved):
Figure looks correct :+1:

PR #207, opened 2022-09-24
Add hint for pkg-config from the build directory

MakisH:
While preCICE v2.5.0 introduced a `libprecice.pc` file specific to using preCICE from the build directory, this was not documented clearly, at least not next to where I would expect to find it.

@fsimonis please check for sanity and let me know if this is already documented somewhere else.

### DIFF
--- a/pages/docs/installation/installation-source-installation.md
+++ b/pages/docs/installation/installation-source-installation.md
@@ -31,6 +31,10 @@ export PKG_CONFIG_PATH=$PRECICE_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH
 export CMAKE_PREFIX_PATH=$PRECICE_PREFIX:$CMAKE_PREFIX_PATH
 ```
 
+{% tip %}
+Do you want to use preCICE from the build directory, without installing it to the prefix? Set both `PRECICE_PREFIX` and `PKG_CONFIG_PATH` to the build directory. This setting will let png-config discover a `libprecice.pc` configuration for using preCICE from the build directory.
+{% endtip %}
+
 After adding these variables, please start a new session (open a new terminal or logout and login again).
 
 {% note %}


fsimonis:
We currently also don't mention how to use precice from the build directory using CMake.

This note is important, but doesn't conceptually fit in the "building from source" guide. The goal here is to guide the user to having a functioning installation in a defined prefix.

We should probably add a page on using precice from the build directory in the developer docs, though.

MakisH:
> This note is important, but doesn't conceptually fit in the "building from source" guide. The goal here is to guide the user to having a functioning installation in a defined prefix.

We already partially have this here, so this is why I extended the note.

> We should probably add a page on using precice from the build directory in the developer docs, though.

I am afraid that this would be too far from the actual installation instructions, making it a perfect candidate for forgetting. We already have similarly outdated pages for special systems.

I would either make it a short section here, or make a complete "onboarding" page in the developers section. If this note is not fine, of course.

fsimonis:
> We already partially have this here, so this is why I extended the note.

No. Using preCICE from the build directory isn't mentioned in the "building from source" section, which is deliberate.

The only mention we have regarding using preCICE from the build directory is a single sentence in the CMake section of "linking to preCICE".

> I am afraid that this would be too far from the actual installation instructions, making it a perfect candidate for forgetting.

The discussion in https://github.com/precice/precice/issues/674 showed that the benefit for the user is not 100% clear. This feature is definitely useful for preCICE developers, maybe useful for adapter developer and potentially confusing for pure users.

> We already have similarly outdated pages for special systems.

Not sure what you mean. The "special systems" page is designed to be updated by people that use these systems.

### Suggestion

Let's create tailing paragraphs in the CMake and [Make and scripts in general](https://precice.org/installation-linking.html#make-and-scripts-in-general) which explain 
* that one can use preCICE from the build directory
* that this is primarily meant for preCICE developers
* this is discouraged unless you have a reason to do so
* build support for adapters requires an installed version of preCICE
* how to do this for CMake/pkg-config

MakisH:
> > We already partially have this here, so this is why I extended the note.
> 
> No. Using preCICE from the build directory isn't mentioned in the "building from source" section, which is deliberate.

Indeed, the build directory is not mentioned there, but we have the environment variables to set. I am fine with moving this information to a different place, but it will make sense to link to that place from this one. Otherwise, I am not ever going to find it again, and I will keep guessing how to do it next time that after several months/years I need to do a similar setup.

> The only mention we have regarding using preCICE from the build directory is a single sentence in the CMake section of "linking to preCICE".

Or at least from here, please? 🥺

> > I am afraid that this would be too far from the actual installation instructions, making it a perfect candidate for forgetting.
> 
> The discussion in [precice/precice#674](https://github.com/precice/precice/issues/674) showed that the benefit for the user is not 100% clear. This feature is definitely useful for preCICE developers, maybe useful for adapter developer and potentially confusing for pure users.

:+1:

> > We already have similarly outdated pages for special systems.
> 
> Not sure what you mean. The "special systems" page is designed to be updated by people that use these systems.

Fair enough. But if we duplicate information to some less frequently-accessed page, information will get outdated. This is the main reason we want developers and users to use the same source of truth. We should at least link from the one side to the other.

> ### Suggestion
> 
> Let's create tailing paragraphs in the CMake and [Make and scripts in general](https://precice.org/installation-linking.html#make-and-scripts-in-general) which explain
> 
>     * that one can use preCICE from the build directory
> 
>     * that this is primarily meant for preCICE developers
> 
>     * this is discouraged unless you have a reason to do so
> 
>     * build support for adapters requires an installed version of preCICE
> 
>     * how to do this for CMake/pkg-config

This sounds more complicated that what I am proposing here. How about we make a dedicated section with all these arguments and warnings in either of the [Linking to preCICE](https://precice.org/installation-linking.html) or [Installation](https://precice.org/installation-source-installation.html)? I still see the Installation page as more natural, since the environment variables are already set there.

I think that your perspective is "build directory vs prefix", while my perspective is "setting environment variables vs installation to a system-known prefix".

I understand that the proper way is to go step-by-step through the complete documentation (at least the installation guide) and that the installation prefix has a different meaning and structure than the build directory, but the user (or developer) is most probably going to land on one of these two pages from a search or guess.

fsimonis:
> This sounds more complicated that what I am proposing here. How about we make a dedicated section with all these arguments and warnings in either of the [Linking to preCICE](https://precice.org/installation-linking.html) or [Installation](https://precice.org/installation-source-installation.html)? I still see the Installation page as more natural, since the environment variables are already set there.

Conceptually, it makes no sense to include a section on how to use preCICE without installing it in the installation page.

A possibility is to move the section "Make the preCICE installation findable" of the installation page to its own separate page.
There we could mention how to use preCICE from the build directory.

> I think that your perspective is "build directory vs prefix", while my perspective is "setting environment variables vs installation to a system-known prefix".

These perspectives are very similar. There are three cases: system prefix, non-system prefix and build directory. The latter two require environment variables, the former doesn't.
Problem is that a complete env setup for the build directory differs from the custom prefix as we have to deal with different directories (`CPATH` + `LD_LIBRARY_PATH` currently contains 3 different directories).

The "building from source" docs are designed to always install preCICE in some prefix. Meaning, in case of a non-system prefix, we have to explain only a single simple case.

If we mention how to use preCICE from the build directory, then we have to always check if `CPATH` is correct as we deal with auto-generated headers that are located in different directories.

We could of course rely on pkg-config to extend `CPATH` and `LD_LIBRARY_PATH`, but that looks error-prone to me. Especially if pkg-config fails in the `~.bashrc`.

```bash
export PKG_CONFIG_PATH...
export CPATH=$CPATH$(pkg-config --cflags-only-I libprecice | sed "s/\W*-I/:/g")
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH$(pkg-config --libs-only-L libprecice | sed "s/\W*-L/:/g")
```

MakisH:
> A possibility is to move the section "Make the preCICE installation findable" of the installation page to its own separate page.
There we could mention how to use preCICE from the build directory.

I like this suggestion, as it seems to solve the root cause of our disagreement.

> Conceptually, it makes no sense to include a section on how to use preCICE without installing it in the installation page.

While I agree, I am not sure if the user understands the difference and can easily find what she needs. "Installation" is a term that is often used with different meanings. If we decouple the `make install` part from the "ok, I installed it, what now?" part (without stepping too much into the "linking to preCICE" territory), then we educate the user.

I can then imagine a tip in the installation page that says "Do you prefer to use preCICE without installing it? Check out *that page*.".

Thank you for the review, I will revise and ask you again when I am ready.

MakisH:
Closing in favor of #227.

PR #208, opened 2022-09-28
Update petsc git and website

erikscheurer:
I got an error during PETSc compilation and found that the links on the website were not up to date. On [BitBucket](https://bitbucket.org/petsc/petsc/src/maint/) where the current guide pulls from, it says:
> PETSc repository is moved to https://gitlab.com/petsc/petsc The bitbucket archive is at https://bitbucket.org/petsc/petsc-moved-to-gitlab This repository is primarily to preserve old URLs.

Since I don't get errors when following the guide on the new [PETSc website](https://petsc.org/release/download) which pulls from [GitLab](https://gitlab.com/petsc/petsc.git), it probably is time to update the links.

There is also a mention of the BitBucket adress in the [special-systems guide](https://precice.org/installation-special-systems.html#building-with-cmake) but I'm not sure if I should touch those.

### DIFF
--- a/pages/docs/installation/installation-source-dependencies.md
+++ b/pages/docs/installation/installation-source-dependencies.md
@@ -213,11 +213,11 @@ We require at least version 3.12. For preCICE versions earlier than v2.1.0, PETS
 
 If you prefer to install the most recent version from source, do the following:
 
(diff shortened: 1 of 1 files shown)


davidscn (commented):
Thanks! Let's not update the CooLMuc docs, it's anyway a bit outdated and PETSc is available as a module there.

davidscn (approved):

PR #212, opened 2022-10-24
Simply avoid simply

fsimonis:
This PR simply rewrites the documentation to avoid using simply.
Many things simply are more complicated than they seem.

Imported repos still need adjusting.


pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
```diff
@@ -10,7 +10,7 @@ The CalculiX adapter for preCICE directly modifies the source code of CalculiX a
 [CalculiX](http://www.dhondt.de) consists of the solver, called "CCX" and a pre- and postprocessing software with graphical user interface "CGX".
 
 - The installation procedure of CCX is described in its `src/README.INSTALL` files, but we also give a summary here.
-- We don't modify CGX, so you can simply get a binary package (if needed, e.g. as a preprocessor in our FSI tutorials)
+- We don't modify CGX, so you can get a binary package (if needed, e.g. as a preprocessor in our FSI tutorials)
```
MakisH, line 13: Here another term would be helpful to emphasize that a binary package is the simpler approach compared to building from source. Maybe "directly"?

pages/docs/adapters/fenics/adapter-fenics.md
```diff
@@ -30,7 +30,7 @@ The adapter requires FEniCS and preCICE version 2.0 or greater and the preCICE l
 
 ### Use `pip`
 
-The adapter is [published on PyPI](https://pypi.org/project/fenicsprecice/). After installing preCICE and the python language bindings one can simply run `pip3 install --user fenicsprecice` to install the adapter via your Python package manager.
+The adapter is [published on PyPI](https://pypi.org/project/fenicsprecice/). After installing preCICE and the python language bindings run `pip3 install --user fenicsprecice` to install the adapter via your Python package manager.
```
MakisH:
```suggestion
The adapter is [published on PyPI](https://pypi.org/project/fenicsprecice/). After installing preCICE and the python language bindings, run `pip3 install --user fenicsprecice` to install the adapter via your Python package manager.
```

pages/docs/configuration/configuration-coupling-multi.md
```diff
@@ -7,7 +7,7 @@ summary: "If you want to couple more than two participants, there are two option
 
 ## Composition of bi-coupling schemes
 
-To combine multiple coupling schemes, simply add them one after the other in the configuration:
+To combine multiple coupling schemes, add them one after the other in the configuration:
```
MakisH: This simplicity is a selling feature of preCICE. Maybe we can write it in a way that still implies that. How about:

```suggestion
preCICE allows us to combine multiple coupling schemes by only specifying them one after the other in the configuration:
```

I also like the "boils down to" that you used somewhere else.

MakisH (approved):
Thank you for simply editing these simply awkward sentences!

For context: https://twitter.com/dave_universetf/status/1580734944909070339?t=IxD0MF6CX71ifLjrdTli3Q&s=19

(refers to the [Kubernetes documentation style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding))

pages/docs/adapters/calculix/adapter-calculix-get-calculix.md
fsimonis: Why not simply:

> The adapter is fully compativle with CGX. You can use it from your package repository or compile it from source.

MakisH: Ah right, that was about CGX. It is simply fine both as you changed it, as well as you suggested in your last comment.

PR #216, opened 2022-11-13
Specify guidelines for meshes in multi-coupling

MakisH:
Documents the naming conventions documented in https://github.com/precice/tutorials/pull/303, adding:

> - Mesh names of participants with multiple interfaces contain the interface in the mesh name, e.g. `Fluid-Upstream-Mesh`. For meshes on which it is important to distinguish between face centers and face nodes, the modifier comes at the end, e.g. `Fluid-Upstream-Mesh-Centers`.

I believe that `Fluid-Upstream-Centers-Mesh` would make more sense, but this is closer to our current practice.


uekerman (approved):
LGTM

PR #221, opened 2022-11-15
Add heat-exchanger-simplified to sidebar

MakisH:
To merge just before we update master **and** https://github.com/precice/tutorials/pull/301 has been merged.

### DIFF
--- a/_data/sidebars/tutorial_sidebar.yml
+++ b/_data/sidebars/tutorial_sidebar.yml
@@ -57,6 +57,10 @@ entries:
     - title: Heat exchanger
       url: /tutorials-heat-exchanger.html
       output: web
+    
+    - title: Heat exchanger simplified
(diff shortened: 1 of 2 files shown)


IshaanDesai (commented):
Don't we want to add this also to the [Further Cases](https://precice.org/tutorials.html#further-cases) section of the tutorials page?

MakisH:
> Don't we want to add this also to the [Further Cases](https://precice.org/tutorials.html#further-cases) section of the tutorials page?

Good point, thanks! I completely forgot about that section. I added it.

IshaanDesai (commented):
Sorry I just noticed one more thing, I looked up an older pull request when I added a tutorial case and I also added it to the `subprojects` part of `_config.yml`. Do we need this?

PR #222, opened 2022-11-16
Add heat-exchanger-simplified to _config.yml

IshaanDesai:
In https://github.com/precice/precice.github.io/pull/221, the case heat-exchange-simplified was not added to `_config.yml`, which led to a broken link. The case is now added.


MakisH (approved):
Thanks! The previous comment was very close to the moment I pressed "merge", thus I missed it. :see_no_evil:

PR #226, opened 2022-11-18
Add oscillator tutorial to the lists

MakisH:
Following https://github.com/precice/tutorials/pull/297

@IshaanDesai: Are the technicalities ok?
@uekerman: Does the position in the list make sense?


uekerman (approved):
Looks good, the order is anyway screwed up a bit. Maybe we should go alphabetically eventually.

PR #227, opened 2022-11-21
Add discoverability section

fsimonis:
This PR adds a discoverability section in response to #207.

I didn't pick the following names:
* Findability just sounds silly
* Detectability sounds like a process needs to be run to detect it.
* Integration looks too similar to Installation (as both are directly next to each other)

### DIFF
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -85,8 +85,8 @@ entries:
             url: /installation-source-installation.html
             output: web, pdf
 
-          - title: Discoverability
-            url: /installation-source-discoverability.html
+          - title: Finding
+            url: /installation-source-finding.html
             output: web, pdf
 
           - title: Advanced
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -85,6 +85,10 @@ entries:
             url: /installation-source-installation.html
             output: web, pdf
 
+          - title: Discoverability
+            url: /installation-source-discoverability.html
+            output: web, pdf
+
           - title: Advanced
             url: /installation-source-advanced.html
             output: web, pdf
--- a/pages/docs/installation/installation-source-discoverability.md
+++ b/pages/docs/installation/installation-source-discoverability.md
@@ -26,11 +26,12 @@ However, they also need to find preCICE in some way:
 
 If you are using a Unix-like system, you are using a shell, which offers an easy and straight forward way to make preCICE discoverable.
 
-Let `PRECICE_PREFIX` be the chosen installation prefix.
+Let `PRECICE_PREFIX` be the installation prefix chosen during the [preparation step](installation-source-preparation#installation-prefix).
 Then add the following to your `.profile` (for bash) or `.zshrc` (for zsh).
 
 ```bash
-PRECICE_PREFIX=~/software/prefix # set this to your selected prefix
+# set this to your selected installation prefix
+PRECICE_PREFIX=~/software/prefix
 export PATH=$PRECICE_PREFIX/bin:$PATH
 export LD_LIBRARY_PATH=$PRECICE_PREFIX/lib:$LD_LIBRARY_PATH
 export CPATH=$PRECICE_PREFIX/include:$CPATH
@@ -51,10 +52,11 @@ Create the directory if it does not yet exist:
 mkdir -p ~/.config/environment.d/
 ```
 
+Let `PRECICE_PREFIX` be the installation prefix chosen during the [preparation step](installation-source-preparation#installation-prefix).
 Then create a file `~/.config/environment.d/99-precice.conf` with the content:
 
 ```conf
-# set this to your selected prefix
+# set this to your selected installation prefix
 PRECICE_PREFIX=$HOME/software/prefix
 PATH=${PRECICE_PREFIX}/bin:${PATH}
 LD_LIBRARY_PATH=${PRECICE_PREFIX}/lib:${LD_LIBRARY_PATH}
--- a/pages/docs/installation/installation-source-discoverability.md
+++ b/pages/docs/installation/installation-source-discoverability.md
@@ -18,7 +18,7 @@ The preCICE library needs to be discoverable in various ways:
 As this setup can become tedious to maintain, there are some tools that expose the above details.
 However, they also need to find preCICE in some way:
 
-* `pkg-config` and `pkgconfig` use `PKG_CONFIG_PATH` to search for additional `.pc` files.
+* `pkg-config` and `pkgconf` use `PKG_CONFIG_PATH` to search for additional `.pc` files.
 * `CMake` uses `CMAKE_PREFIX_PATH` for additional installation prefixes.
   Alternatively, one can specify the location of the preCICE CMake configuration file using `precice_DIR`.
 
--- a/pages/docs/installation/installation-source-discoverability.md
+++ b/pages/docs/installation/installation-source-discoverability.md
@@ -1,0 +1,85 @@
+---
+title: Building from source - Discoverability
+permalink: installation-source-discoverability.html
+keywords: configuration, basics, cmake, installation, building, source, bash, profile
+toc: false
+---
+
+If you installed preCICE in a custom prefix, then one still needs to make it discoverable by the system.
+
+The preCICE library needs to be discoverable in various ways:
+
+1. The location of headers needs to be available during compilation.
+   Compilers use hints from `CPATH` if no extra compilation flags were passed to them using `-I`.
+2. the shared library needs to be available during linking and execution.
+   The dynamic linker uses hints from `LD_LIBRARY_PATH` (`DYLD_LIBRARY_PATH` on MacOS X).
+3. an optional step is to make the preCICE executables available using `PATH`.
+
+As this setup can become tedious to maintain, there are some tools that expose the above details.
+However, they also need to find preCICE in some way:
+
+* `pkg-config` and `pkgconfig` use `PKG_CONFIG_PATH` to search for additional `.pc` files.
+* `CMake` uses `CMAKE_PREFIX_PATH` for additional installation prefixes.
+  Alternatively one can specify the location of the preCICE configuration file using `precice_DIR`.
+
+## Using the shell
+
+If you are using a Unix-like system, you are using a shell, which is an easy and straight forward way of making preCICE discoverable.
+
+Let `PRECICE_PREFIX` be the chosen installation prefix. 
+Then add the following to your `.profile` (for bash) or `.zshrc` (for zsh).
+
+```bash
+PRECICE_PREFIX=~/software/prefix # set this to your selected prefix
+export PATH=$PRECICE_PREFIX/bin:$PATH
+export LD_LIBRARY_PATH=$PRECICE_PREFIX/lib:$LD_LIBRARY_PATH
+export CPATH=$PRECICE_PREFIX/include:$CPATH
+# Enable detection with pkg-config and CMake
+export PKG_CONFIG_PATH=$PRECICE_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH
+export CMAKE_PREFIX_PATH=$PRECICE_PREFIX:$CMAKE_PREFIX_PATH
+```
+
+After adding these variables, please start a new session (open a new terminal or logout and login again).
+
+## Using systemd environment.d
+
+An alternative to the above system is to set these variables in the shell-agnostic [`environment.d`](https://www.man7.org/linux/man-pages/man5/environment.d.5.html).
+
+Make sure the directory exists:
+
+```terminal
+mkdir -p ~/.config/environment.d/
+```
+
+Then create a file `~/.config/environment.d/99-precice.conf` with the content:
+
+```conf
+# set this to your selected prefix
+PRECICE_PREFIX=$HOME/software/prefix
+PATH=${PRECICE_PREFIX}/bin:${PATH}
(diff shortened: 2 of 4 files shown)


pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+If you installed preCICE in a custom prefix, then one still needs to make it discoverable by the system.
+
+The preCICE library needs to be discoverable in various ways:
+
+1. The location of headers needs to be available during compilation.
+   Compilers use hints from `CPATH` if no extra compilation flags were passed to them using `-I`.
+2. the shared library needs to be available during linking and execution.
+   The dynamic linker uses hints from `LD_LIBRARY_PATH` (`DYLD_LIBRARY_PATH` on MacOS X).
+3. an optional step is to make the preCICE executables available using `PATH`.
```
MakisH:
```suggestion
1. The location of headers needs to be available during compilation.
   Compilers use hints from `CPATH` if no extra compilation flags were passed to them using `-I`.
2. The shared library needs to be available during linking and execution.
   The dynamic linker uses hints from `LD_LIBRARY_PATH` (`DYLD_LIBRARY_PATH` on MacOS X).
3. The preCICE executables need to be in `PATH` (optional, only for additional tools).
```

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+As this setup can become tedious to maintain, there are some tools that expose the above details.
+However, they also need to find preCICE in some way:
+
+* `pkg-config` and `pkgconfig` use `PKG_CONFIG_PATH` to search for additional `.pc` files.
```
MakisH:
```suggestion
* `pkg-config` uses `PKG_CONFIG_PATH` to search for additional `.pc` files.
```

Is there a **command/binary** `pkgconfig`? I thought this is only the name of the project.

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+* `CMake` uses `CMAKE_PREFIX_PATH` for additional installation prefixes.
+  Alternatively one can specify the location of the preCICE configuration file using `precice_DIR`.
```
MakisH:
```suggestion
  Alternatively, one can specify the location of the preCICE CMake configuration file using `precice_DIR`.
```

(we already use "preCICE configuration file" too much with the other meaning. Yes, I know, the context is clear to someone that reads the page, but people search and think in unexpected ways)

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Using the shell
+
+If you are using a Unix-like system, you are using a shell, which is an easy and straight forward way of making preCICE discoverable.
```
MakisH:
```suggestion
If you are using a Unix-like system, you are using a shell, which offers an easy and straight forward way to make preCICE discoverable.
```

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+Let `PRECICE_PREFIX` be the chosen installation prefix.
+Then add the following to your `.profile` (for bash) or `.zshrc` (for zsh).
```
MakisH: Why `.profile` for bash but `.zshrc` for zsh? The corresponding file for bash is `.bashrc`, no?

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+```bash
+PRECICE_PREFIX=~/software/prefix # set this to your selected prefix
+export PATH=$PRECICE_PREFIX/bin:$PATH
```
MakisH: I know that this was already like this before, but try `shellcheck` on this. I would at least add double quotes to treat any spaces in the paths.

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+export CMAKE_PREFIX_PATH=$PRECICE_PREFIX:$CMAKE_PREFIX_PATH
+```
+
+After adding these variables, please start a new session (open a new terminal or logout and login again).
+
+## Using systemd environment.d
+
+An alternative to the above system is to set these variables in the shell-agnostic [`environment.d`](https://www.man7.org/linux/man-pages/man5/environment.d.5.html).
+
+Make sure the directory exists:
```
MakisH:
```suggestion
Create the directory if it does not yet exist:
```

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+PRECICE_PREFIX=$HOME/software/prefix
+PATH=${PRECICE_PREFIX}/bin:${PATH}
+LD_LIBRARY_PATH=${PRECICE_PREFIX}/lib:${LD_LIBRARY_PATH}
+CPATH=${PRECICE_PREFIX}/include:${CPATH}
+# Enable detection with pkg-config and CMake
+PKG_CONFIG_PATH=${PRECICE_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH}
+CMAKE_PREFIX_PATH=${PRECICE_PREFIX}:${CMAKE_PREFIX_PATH}
+```
+
+After adding the file, please logout and login again.
```
MakisH: I don't believe I am the one writing this, but please don't use `please`, as we don't use it elsewhere in the docs :see_no_evil:

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Using directly from the build directory
+
+It may not always be practical to install preCICE repeatedly.
+This is especially the case for simultaneous development of preCICE and an adapter.
+
+This method is discouraged as file layouts are fundamentally different and we make guarantees on their locations.
+The supported methods of using preCICE from the build directory are using `pkg-config` and `CMake`.
+
+For `pkg-config`, extend `PKG_CONFIG_PATH` with the binary directory of preCICE.
```
MakisH: Add example of this path.

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+If you are using a Unix-like system, you are using a shell, which is an easy and straight forward way of making preCICE discoverable.
+
+Let `PRECICE_PREFIX` be the chosen installation prefix.
+Then add the following to your `.profile` (for bash) or `.zshrc` (for zsh).
+
+```bash
+PRECICE_PREFIX=~/software/prefix # set this to your selected prefix
+export PATH=$PRECICE_PREFIX/bin:$PATH
+export LD_LIBRARY_PATH=$PRECICE_PREFIX/lib:$LD_LIBRARY_PATH
+export CPATH=$PRECICE_PREFIX/include:$CPATH
```
MakisH: We also need the C bindings from `$PRECICE_PREFIX/../extras/bindings/c/include`

```suggestion
export CPATH=$PRECICE_PREFIX/include:$PRECICE_PREFIX/../extras/bindings/c/include:$CPATH
```

pages/docs/installation/installation-source-discoverability.md
```diff
@@ -0,0 +1,84 @@
...
+
+It may not always be practical to install preCICE repeatedly.
+This is especially the case for simultaneous development of preCICE and an adapter.
+
+This method is discouraged as file layouts are fundamentally different and we make guarantees on their locations.
+The supported methods of using preCICE from the build directory are using `pkg-config` and `CMake`.
+
+For `pkg-config`, extend `PKG_CONFIG_PATH` with the binary directory of preCICE.
+For `CMake`, set `precice_ROOT` to the binary directory of preCICE.
+
```
MakisH: I think not all adapters support pkgconfig at the moment.

This is my current setup, parts of which are not covered here (even in the "unsafe - don't try this at home" section):

```bash
PRECICE_PREFIX="/home/gc/repos/precice/precice/build"
export LD_LIBRARY_PATH="$PRECICE_PREFIX:$LD_LIBRARY_PATH"
export LIBRARY_PATH="$PRECICE_PREFIX:$LD_LIBRARY_PATH"
export CPATH="$PRECICE_PREFIX/src:$PRECICE_PREFIX/../extras/bindings/c/include/:$CPATH"
export INCLUDE_PATH="$PRECICE_PREFIX/../extras/bindings/c/include/:$INCLUDE_PATH"
export PKG_CONFIG_PATH="$PRECICE_PREFIX:$PKG_CONFIG_PATH"
export CMAKE_PREFIX_PATH="$PRECICE_PREFIX:$CMAKE_PREFIX_PATH"
```

_data/sidebars/docs_sidebar.yml
```diff
@@ -85,6 +85,10 @@ entries:
             url: /installation-source-installation.html
             output: web, pdf
 
+          - title: Discoverability
```
MakisH: I don't really like the name, but I see the difficulty of choosing a good one... :sweat_smile: No better suggestion at the moment.

pages/docs/installation/installation-source-discoverability.md
fsimonis: There are two implementations, which behave slightly differently.

pages/docs/installation/installation-source-discoverability.md
fsimonis: Implementing pkg-config in the adapters is quicker than tracing typos in these paths. :see_no_evil:

pages/docs/installation/installation-source-discoverability.md
MakisH: Ah, I didn't know. But then it is pkg-config and pkgconf, right? Or PkgConf. At least according to Wikipedia: https://en.wikipedia.org/wiki/Pkg-config#Alternative_implementations

_data/sidebars/docs_sidebar.yml
uekerman: What about "Finding preCICE"? Style-wise this would fit to "Buidling from source" and "Linking to preCICE".
And it would reflect the CMake terminology `find_package` etc.

fsimonis: This section is a subpage of "installing from source" and they all use single word names.
I would prefer to stay consistent here.

pages/docs/installation/installation-source-discoverability.md
fsimonis: `pkg-config` and `pkgconf` are correct.
Thanks for pointing that out!

pages/docs/installation/installation-source-discoverability.md
fsimonis: We install all headers into `<prefix>/include/precice/`, including the C header.

pages/docs/installation/installation-source-discoverability.md
fsimonis: If the user (or dev) doesn't know what a binary directory is, then this method isn't for them.

pages/docs/installation/installation-source-discoverability.md
fsimonis: I reworded the section. Using CPATH etc is not worth it.
Use pkg-config or install to a prefix.

pages/docs/installation/installation-source-discoverability.md
fsimonis: I just used shellcheck on this. No issues detected.

I expect that spaces aren't an issue as the paths are colon-separated.

_data/sidebars/docs_sidebar.yml
MakisH: How about just "Finding", then?
We already mix e.g., "Instal**lation**" and "Build**ing**"
"Discoverability" is anyway a property (getting tricky to explain what I mean).

pages/docs/installation/installation-source-discoverability.md
MakisH: I thought clarity is always nice, but ok :grimacing:

_data/sidebars/docs_sidebar.yml
fsimonis: Both are not great as they can easily be confused with a guide on finding the dependencies.

Why is naming so difficult...

MakisH: I would never think that any of the names proposed so far, and at that place in the structure, would mean finding the dependencies, especially since they have their own prominent page above. I think it makes sense if we look at the order:

```
Preparation
Dependencies
Configuration
Building
Testing
Installation
Finding
Advanced
```

I would maybe even bring consistency to all (only on the sidebar, to not break any links):

```
Preparation
Dependencies
Configuring
Building
Testing
Installing
Finding
Advanced
```

pages/docs/installation/installation-source-discoverability.md
MakisH: Interestingly, this depends on the shell. With `/bin/sh`, shellcheck 0.8.0 complains about the missing double quotes. With `/bin/bash`, it does not.

Since in other places in the documentation (e.g., in [Dependencies](https://precice.org/installation-source-dependencies.html) and in [Installation](https://precice.org/installation-source-installation.html)) we use the variant without quotes, let's stick with that for consistency.

_data/sidebars/docs_sidebar.yml
MakisH: Looking at all the arguments raised so far, I would go with `Finding` (fits with `Building` and `Testing`, is one word, fits with the CMake `find_package`, pkg-config also uses the term "to find a library").

In a follow-up, we could change the sidebar entries for consistency (but not the permalinks, please).

I will change the name and merge.

MakisH (approved):

PR #230, opened 2022-11-23
Distribution: Add tools section (ASTE, config-visualizer, ci-images)

MakisH:
Previously, we did not mention any tools in the distribution page, so I overlooked that we wanted to include ASTE, the config-visualizer, and the ci-images this time.

This adds a "tools" section underneath preCICE, with these three components.

@fsimonis please confirm the versions. There is nothing else we wanted to include this time, right?

@uekerman FYI

### DIFF
--- a/pages/docs/installation/installation-distribution.md
+++ b/pages/docs/installation/installation-distribution.md
@@ -23,8 +23,12 @@ This is the last release that works with preCICE v2.x.
 It comprises the following components:
 
 - preCICE: [v2.5.0](https://github.com/precice/precice/releases/tag/v2.5.0)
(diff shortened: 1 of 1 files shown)


fsimonis:
Do we really want to include the ci images? They aren't useful to the normal user.

aste and config visualizer look fine.

MakisH:
> Do we really want to include the ci images? They aren't useful to the normal user.
> 
> aste and config visualizer look fine.

They are in the pad as a new component, without any comments regarding not including them. Whatever you prefer. I would say they are helpful to users and developer-users relying on Docker.

PR #233, opened 2022-12-08
Add implementation hints for mesh connectivity

fsimonis:
This PR adds hints for mesh connectivity implementation in adapters.

Feedback is highly appreciated.

### DIFF
--- a/pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
+++ b/pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
@@ -141,6 +141,7 @@ It does provide a functionality to attach some kind of customised attribute to a
 This attribute is also available when iterating over faces.
 
 Examples of such attributes:
+
 * custom tags or labels: `vertex.label = myinfo` and `myinfo = vertex.label`
 * custom key-value dictionaries: `vertex.attributes[mykey] = myvalue` and `myvalue = vertex.attributes[mykey]`
 
--- a/pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
+++ b/pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
@@ -104,70 +104,79 @@ In this section, we handle some common cases and how to implement them.
 
 ### Solver provides IDs
 
-You solver gives each vertex or node a unique identifier.
+Your solver gives each vertex a unique identifier.
 This identifier is also available when iterating over faces.
 
 In this case you can save a mapping from Solver ID to preCICE Vertex ID, after defining the vertices.
 When iterating over the faces, get the point identifier, map them to preCICE IDs and use those to define your connectivity.
 
 
-### Solver supports labels or tags
-
-You solver doesn't provide a unique identifier for each vertex or node.
-It does provide a functionality to attach some kind of label or tag to nodes.
-This tag/label is also available when iterating over faces.
-
-In this case you can save preCICE Vertex ID as labels directly in the solver vertices.
-Define the vertices using the preCICE API, then iterate over them and apply the preCICE vertex IDs as labels.
-When iterating over the faces, get the preCICE vertex IDs from the point labels, and use those to define your connectivity.
-
-
-### Solver supports coordinates only
-
-Your solver provides coordinates only or labels are already used for another purpose.
-
-In this case, you need to generate a mapping from coordinates to preCICE vertex ids.
-Due to rounding error, a C++ `std::map` or python `dict` may not be sufficient.
-
-An alternative would be to use a spacial index as a data-structure to store this information for you.
-
 ```cpp
-#include <boost/geometry.hpp>
-#include <cassert>
+SolverInterface participant(...);
+auto meshID = participant.getMesh(...);
 
-class MiniLookup {
-   public:
-    // The type the preCICE ID
-    using ID = int;
-    // Point type, here 3D double in a cartesian space
-    using Point = boost::geometry::model::point<double, 3,
-                                                boost::geometry::cs::cartesian>;
-    using Value = std::pair<Point, ID>;
+// Define the map from the solver to the preCICE vertex ID
+std::map<Solver::VertexID, precice::VertexID> idMap;
+for (auto& vertex: solver.vertices) {
+  auto vertexID = participant.setMeshVertex(meshID, vertex.coords);
+  // set the vertexID as label
+  idMap.emplace(vertex.id, vertexID);
+}
 
-    void insert(const Point& location, ID id) {
-        _tree.insert(std::make_pair(location, id));
-    }
-
-    ID lookup(const Point& location) const {
-        assert(!_tree.empty());
-        Value result;
-        _tree.query(boost::geometry::index::nearest(location, 1), &result);
-        return result.second;
-    }
-
-   private:
-    boost::geometry::index::rtree<Value, boost::geometry::index::linear<32>>
-        _tree;
-};
+for (auto& tri: solver.triangularFaces) {
+  // Lookup the preCICE vertex ID using the solver vertex ID
+  auto a = idMap.at(tri.a.id);
+  auto b = idMap.at(tri.b.id);
+  auto c = idMap.at(tri.c.id);
+  // Then define the connectivity
+  participant.setMeshTriangle(meshID, a, b, c);
+}
 ```
 
-Fill this tree, after setting preCICE vertices:
+### Solver supports labels
+
+You solver doesn't provide a unique identifier for each vertex.
+It does provide a functionality to attach some kind of label to a vertex.
+This label is also available when iterating over faces.
+
+In this case you can save preCICE Vertex IDs as labels directly in the solver vertices.
+Define the vertices using the preCICE API, then iterate over them and apply the preCICE vertex IDs as labels.
+When iterating over faces, get the preCICE vertex IDs from the point labels, and use those to define your connectivity.
 
 ```cpp
 SolverInterface participant(...);
 auto meshID = participant.getMesh(...);
 
-MiniLookup lookup;
+for (auto& vertex: solver.vertices) {
+  auto vertexID = participant.setMeshVertex(meshID, vertex.coords);
+  vertex.label = vertexID; // set the vertexID as label
+}
+
+for (auto& tri: solver.triangularFaces) {
+  // Extract the vertex IDs from the vertex labels
+  auto a = tri.a.label;
+  auto b = tri.b.label;
+  auto c = tri.c.label;
+  // Then define the connectivity
+  participant.setMeshTriangle(meshID, a, b, c);
+}
+```
+
+### Solver supports coordinates only
+
+Your solver provides coordinates only or labels are already used for another purpose.
+
+In this case, you need to generate a mapping from coordinates to preCICE vertex IDs.
+Depending on your solver, coordinates available at various stages may be subject to rounding error.
+Hence, a C++ `std::map` without custom comparator, or python `dict` may not be sufficient.
+
+An alternative would be to use a spatial index as a data structure to store this information.
+
+```cpp
+SolverInterface participant(...);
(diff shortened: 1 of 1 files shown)


pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+
+## Putting it all together
+
+Solvers may give you a range of information regarding vertices and faces.
+Some may give you unique ids for vertices, some only coordinates.
+In this section, we handle some common cases and how to implement them.
+
+### Solver provides IDs
+
+You solver gives each vertex or node a unique identifier.
```
MakisH:
```suggestion
Your solver gives each vertex or node a unique identifier.
```

also: why distinguishing between vertices and nodes? I understand the same thing. This also applied below.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+This identifier is also available when iterating over faces.
+
+In this case you can save a mapping from Solver ID to preCICE Vertex ID, after defining the vertices.
```
MakisH, line 110:
```suggestion
In this case, you can save a mapping from solver vertex ID to preCICE vertex ID, after defining the vertices.
```

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+When iterating over the faces, get the point identifier, map them to preCICE IDs and use those to define your connectivity.
+
+
+### Solver supports labels or tags
+
+You solver doesn't provide a unique identifier for each vertex or node.
+It does provide a functionality to attach some kind of label or tag to nodes.
```
MakisH: Do you have any particular example in mind for this case?

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+This tag/label is also available when iterating over faces.
+
+In this case you can save preCICE Vertex ID as labels directly in the solver vertices.
```
MakisH: How to get the vertex ID? You probably want to refer to the `getMeshVertexIDsFromPositions`.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+Define the vertices using the preCICE API, then iterate over them and apply the preCICE vertex IDs as labels.
+When iterating over the faces, get the preCICE vertex IDs from the point labels, and use those to define your connectivity.
+
+
+### Solver supports coordinates only
+
+Your solver provides coordinates only or labels are already used for another purpose.
+
+In this case, you need to generate a mapping from coordinates to preCICE vertex ids.
+Due to rounding error, a C++ `std::map` or python `dict` may not be sufficient.
```
MakisH: How are rounding errors related to containers? Linking to some reference in the text (and/or adding more context) would be helpful.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+
+An alternative would be to use a spacial index as a data-structure to store this information for you.
```
MakisH:
```suggestion
An alternative would be to use a spatial index as a data structure to store this information.
```

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+        Value result;
+        _tree.query(boost::geometry::index::nearest(location, 1), &result);
+        return result.second;
+    }
+
+   private:
+    boost::geometry::index::rtree<Value, boost::geometry::index::linear<32>>
+        _tree;
+};
+```
```
MakisH: This class probably needs some more documentation. I understand what every method does, but some summary would have helped me understand it faster.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,108 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+
+Fill this tree, after setting preCICE vertices:
```
MakisH: Maybe it would be easier to first show the usage and then how `MiniLookup` looks like.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Hmm, I don't destinguish between the two. Both are common names, so I mention both. Thinking about it, face centers are also common...

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Doesn't OpenFoam allow to do this?

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
MakisH: But, everywhere in this page (and maybe elsewhere in the docs) we are using "vertices". It is a bit strange to use both suddenly, and I understood `each vertex or node` as something special that I overlooked (e.g., like `each vertex or edge`).

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
MakisH: I was trying to understand which case OpenFOAM falls into, but I think it is the last one. I don't think every cell has a label, but every patch has faces associated with it. But I am not sure at all.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Not really, this will be removed after all :joy:

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Very good point.

fsimonis:
I completely modified the new section and added examples for every case.
Does this look more understandable?

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,150 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+Some may give you unique ids for vertices, some only coordinates.
+In this section, we handle some common cases and how to implement them.
+
+### Solver provides IDs
+
+Your solver gives each vertex a unique identifier.
+This identifier is also available when iterating over faces.
+
+In this case you can save a mapping from Solver ID to preCICE Vertex ID, after defining the vertices.
+When iterating over the faces, get the point identifier, map them to preCICE IDs and use those to define your connectivity.
```
MakisH: what does "point identifier" mean?
(you probably need plural as well)

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,150 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+  auto c = idMap.at(tri.c.id);
+  // Then define the connectivity
+  participant.setMeshTriangle(meshID, a, b, c);
+}
+```
+
+### Solver supports labels
+
+You solver doesn't provide a unique identifier for each vertex.
+It does provide a functionality to attach some kind of label to a vertex.
```
MakisH: Can you give an example of such labels? I understand something like "inner cell / boundary cell / coupled cell", but it looks like you mean again something like a unique identifier for each cell. In that case, I see the second example as a different form of the first example.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,150 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+
+An alternative would be to use a spatial index as a data structure to store this information.
+
+```cpp
+SolverInterface participant(...);
+auto meshID = participant.getMesh(...);
+
+IDLookup lookup;
+for (auto& vertex: solver.vertices) {
+  auto vid = participant.setMeshVertex(meshID, vertex.coords);
```
MakisH, line 187: We probably need to mention that, contrary to what we typically recommend, we need the singular version of `setMeshVertex` here.

MakisH (commented):
This indeed reads more clearly. I still don't really understand the labels case, though.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: I changed the wording to `vertex` only

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Tried to clarify

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: Done

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: We don't need to.
We could also first gather all coordinates in a `std::vector`, then call `setMeshVertices` and gather all ids in a `std::vector`, and finally build the lookup for them coordinates and ids.
It's just more verbose, which is why I used the singular version.

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -95,3 +95,156 @@ void setMeshTriangles(int meshID, int size, int* vertices);
...
+Define the vertices using the preCICE API, then iterate over them and apply the preCICE vertex IDs as labels.
+When iterating over faces, get the preCICE vertex IDs from the point labels, and use those to define your connectivity.
+
+```cpp
+SolverInterface participant(...);
+auto meshID = participant.getMesh(...);
+
+for (auto& vertex: solver.vertices) {
+  auto vertexID = participant.setMeshVertex(meshID, vertex.coords);
+  vertex.label = vertexID; // set the vertexID as label
```
MakisH, line 158: We probably want to define separate labels, or extend the current ones, right? I assume that the solver would still need those labels (e.g., on which vertices lie on a coupled boundary). I am not sure I fully understand this, but it also looks fine as-is.

MakisH (approved):
I went through everything again (the complete page), it makes total sense. :+1:

pages/docs/couple-your-code/couple-your-code-defining-mesh-connectivity.md
fsimonis: I suggest leaving it as is until we have a concrete example of such a case.

PR #234, opened 2022-12-15
Make news global

fsimonis:
This PR moves the news banner html to its own `_include/news-banner.html`.

It is included in every page, use the boolean `include.onlanding` inside to determine where to show the banner.
I needed to add a `.banner { padding: 15px 0; }`  to the `customstyles.css` to make this look decent on non-landing pages.

### DIFF
--- a/_includes/news_banner.html
+++ b/_includes/news_banner.html
@@ -10,7 +10,7 @@ For news visible on all pages use:
 
 {% if true %}
 
-<div class="background-light">
+<div class="background-light banner-container">
   <div class="container">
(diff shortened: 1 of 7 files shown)


MakisH (approved):
Thank you, this looks really nice, as we now have a specific file to change and this will be included by Jekyll on the landing or every page.

I added a class `banner-container` and adjusted the margins to fix the spacing on non-landing pages. On Firefox, this seems to work on all three cases (`true`/`false`/`include.onlanding`).

PR #235, opened 2022-12-15
Document format_precice_config.py

MakisH:
Related to https://github.com/precice/precice/pull/1501 and https://github.com/precice/precice-pre-commit-hooks/pull/1


MakisH:
This is for formatting configuration files, not for the code itself. It is also useful for users contributing tutorials and for maintainers of other repositories that contain configs. It is currently only briefly documented in https://precice.org/community-contribute-to-precice.html#optional-help-us-with-some-checks

While it is included in the pre-commit hook of preCICE, information on formatting the code itself is also documented in the same page.

PR #236, opened 2023-01-05
Update adapter-calculix-get-calculix.md

hojjatnaderi:
At the time of writing this, {{site.calculix_version}} points to 2.2 which is wrong and makes the links broken. If it points to 2.20 then there will be no problem. For the time being I change {{site.calculix_version}} to 2.20 which is not the smartest thing to do but solves the issue anyway.  Also there is typo in "Summary". I changed "challenges" to "challenge".

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -218,9 +218,10 @@ subprojects:
 
 # We use these versions to centrally update links
 # and other version-specific information.
+# Note that Jekyll cuts off trailing zeros: https://stackoverflow.com/questions/37862167/trailing-zeros-in-jekyll-liquid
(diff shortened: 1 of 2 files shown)


MakisH (approved):
Thank you for reporting and trying to fix! :smile: 
The issue was not a hard-coded `2.2`, but a more sneaky Jekyll feature: https://stackoverflow.com/questions/37862167/trailing-zeros-in-jekyll-liquid

I fixed it by converting the respective version to a string. I did not touch the `precice_version`, as I am not sure if it could have any side-effects (and it does not seem to be affected at the moment).

PR #241, opened 2023-01-25
Add notes on porting rbf mapping configurations

davidscn:
... as required by https://github.com/precice/precice/pull/1550


MakisH:
I have not followed the discussion on this feature, but are we now going back from the simple `<mapping:rbf ... >` with automatic algorithm selection to `<mapping::rbf-global-direct ...>` and `<<mapping:rbf-global-iterative ...>`?

Have you already considered keeping the `<mapping:rbf ...>` with a default option?

Remember that we used to have `rbf` and `petrbf`, which was always difficult to explain.

davidscn:
> I have not followed the discussion on this feature, but are we now going back from the simple <mapping:rbf ... > with automatic algorithm selection to <mapping::rbf-global-direct ...> and <<mapping:rbf-global-iterative ...>?

No. Did you read careful enough or do I need to be more explicit?

MakisH:
> > I have not followed the discussion on this feature, but are we now going back from the simple <mapping:rbf ... > with automatic algorithm selection to <mapping::rbf-global-direct ...> and <<mapping:rbf-global-iterative ...>?
> 
> No. Did you read careful enough or do I need to be more explicit?

It's complicated. I think I read it in the same speed as a user would read it, but I got confused. Part of the reason was that I had the (wrong) impression that we already had `<mapping:rbf>` as a tag, or that we at least discussed it at some point (it would be great to have it to auto-generate default configs).

Reading it again, the subtags are great, but this documentation entry could largely benefit from a diff code block with a concrete example. Right now, there is simply too much to follow in inline text.

PR #245, opened 2023-02-20
Clearly differentiate between `participant_dt` and `dt`

BenjaminRodenberg:
I think we should be more clear here, because otherwise the double meaning of `dt` can lead to bugs that are difficult to find. During the preCICE workshop 2023 I found a bug in the code of a user where the statement `dt = beginTimeStep()` was left out, because `dt` was fixed to a constant before the timeloop started `dt = subcycling_dt`. This is totally understandable from my perspective, if you do not use adaptivity. Why should I redefine `dt` in every time step, if it stays the same?

The goal of the user was to use subcycling. However, this led to the situation that `dt` was getting smaller than `subcycling_dt` when getting close to the end of the first window: `min(precice_dt, subcycling_dt)` returns `precice_dt < subcycling_dt` and we overwrite `dt` leading to `dt < subcycling_dt` for the remainder of the simulation time.

I think this is a situation that we can avoid easily and at the same time we can show clearly that there are basically three different `dt`s:

1) `dt` which is the `dt` that we actually use for the time step
2) `precice_dt` which is the maximum allowed `dt` defined by preCICE
3) `participant_dt` (or `solver_dt`?) which is the `dt` desired by the solver (fixed or computed), but might conflict with `precice_dt`

This is still a draft, because we still have the following todos:

1) Do we want this change?
2) Apply it consistently to figures, solverdummies etc.

### DIFF
--- a/pages/docs/couple-your-code/couple-your-code-implicit-coupling.md
+++ b/pages/docs/couple-your-code/couple-your-code-implicit-coupling.md
@@ -43,8 +43,9 @@ int forceID = precice.getDataID("Forces", meshID);
 double* forces = new double[vertexSize*dim];
 double* displacements = new double[vertexSize*dim];
 
-double dt; // solver timestep size
+double solver_dt; // solver timestep size
 double precice_dt; // maximum precice timestep size
+double dt; // actual time step size
 ```
 
 ```cpp
@@ -56,8 +57,8 @@ while (precice.isCouplingOngoing()){
   }
   precice.readBlockVectorData(displID, vertexSize, vertexIDs, displacements);
   setDisplacements(displacements);
-  dt = beginTimeStep(); // e.g. compute adaptive dt 
-  dt = min(precice_dt, dt);
+  solver_dt = beginTimeStep(); // e.g. compute adaptive dt
+  dt = min(precice_dt, solver_dt);
   solveTimeStep(dt);
   computeForces(forces);
   precice.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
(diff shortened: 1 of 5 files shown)


uekerman:
Yes, why not. I am only afraid that this will be "high effort – low gain".

BenjaminRodenberg:
@uekerman Any preference `participant_dt` or `solver_dt`? I think both is fine, but would use `solver_dt`, because it's shorter and we also write `double dt; // solver timestep size` in the step-by-step guide.

uekerman:
> Any preference participant_dt or solver_dt?

No, not really. We explain that both are valid options here: https://precice.org/fundamentals-terminology.html#solver-and-participant

BenjaminRodenberg:
@uekerman I hope I did not oversee any necessary changes.

One thing I just stumbled over: `precice_dt` or `dt_precice`? Same for `solver_dt`. The text in https://precice.org/couple-your-code-timestep-sizes.html sometimes mentions `dt_precice`. Thinking about the mathematical notation $\Delta t_\text{preCICE}$, `dt_precice` would probably make more sense. Same for $\delta t_\mathcal{D}$ in https://precice.org/couple-your-code-waveform.html.

Note: Our tutorials currently use `precice_dt` or `my_dt` (e.g. [here](https://github.com/precice/tutorials/blob/0482e5b4d3e1604dd031d2b113a2aaa63733c9a8/oscillator/python/oscillator.py#L102-L103))

Pragmatic suggestion: Let's keep `precice_dt`, because I think this is really only about consistency (and means a lot of work) and it does not really influence readability.

fsimonis:
Just for completeness: Once we rename the `SolverInterface` to `Participant` https://github.com/precice/precice/issues/1458, then using `participant_dt` for the solver would be highly confusing.

uekerman (approved):
Looks good

PR #252, opened 2023-04-05
Updated description for modal dynamic analysis

AugustoRiedinger:
Specified changes needed to run modal dynamic simulations. The main difference from regular simulations is the need for **two separate input files**, due to a peculiarity of the adapter. This should be emphasized since it differs significantly from CalculiX. 

- Changed the section on `Modal dynamic simulations`.
- Expanded the description of how the system is solved.
- Added example workflow with input file examples.<

The input files examples given are consistent with the [calculix case input file](https://precice.org/adapter-calculix-config.html#calculix-case-input-file) given earlier in the documentation. We just changed the type of analysis on the case input file to a `MODAL DYNAMIC` one.


MakisH (approved):
Thanks a lot for the additional details and the very polished text! :smile: :hugs:

PR #253, opened 2023-04-11
Add OpenFOAM Journal article to literature guide

MakisH:
This updates the [literature guide](https://precice.org/fundamentals-literature-guide.html) page with our new OpenFOAM-preCICE paper:

Screenshot from 2023-04-11 16-14-25

___

This includes a BibTeX file, which, when used in the OFJ template renders as:

Screenshot from 2023-04-11 16-20-54

___

It also adds a news banner for the landing page (only):

Screenshot from 2023-04-11 16-14-43

___

@uekerman @davidscn anything else we need to update in the website repository? Does everything look as expected?

### DIFF
--- a/assets/openfoam-precice.bib
+++ b/assets/openfoam-precice.bib
@@ -1,0 +1,11 @@
+@article{OpenFOAMpreCICE,
+  author = {Chourdakis, Gerasimos and Schneider, David and Uekermann, Benjamin},
+  title = {{OpenFOAM-preCICE}: Coupling {OpenFOAM} with external solvers for multi-physics simulations},
+  journal = {OpenFOAM® Journal},
+  volume = {3},
+  year = {2023},
+  month = {Feb},
+  pages = {1-25},
+  doi = {10.51560/ofj.v3.88},
+  url = {https://doi.org/10.51560/ofj.v3.88}
+}
--- a/_includes/news_banner.html
+++ b/_includes/news_banner.html
@@ -8,14 +8,14 @@ For news visible on all pages use:
 
 {% endraw %}{% endcomment %}
 
-{% if false %}
+{% if include.onlanding %}
 
 <div class="background-light banner-container">
   <div class="container">
     <div class="row no-margin">
       <div class="col-lg-12 banner">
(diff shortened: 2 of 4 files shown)


assets/openfoam-precice.bib
```diff
@@ -0,0 +1,11 @@
+@article{OpenFOAMpreCICE,
+  author = {Chourdakis, Gerasimos and Schneider, David and Uekermann, Benjamin},
+  title = {{OpenFOAM-preCICE}: Coupling {OpenFOAM} with external solvers for multi-physics simulations},
+  journal = {OpenFOAM® Journal},
```
davidscn, line 4: Do we want to hard-copy the `textregistered` symbol here? I would assume that almost all tex engines won't compile it.

davidscn (commented):
I would expect some information here: https://precice.org/adapter-openfoam-overview.html#cite

assets/openfoam-precice.bib
MakisH: pdflatex seems to compile it (see image above). But good point. From my view, this is an important argument for dropping the symbol from the journal's name, but I understand that that would be complicated now.

With `journal = {OpenFOAM\textregistered Journal},` there seems to be a space missing before the "Journal":

Screenshot from 2023-04-11 17-00-16

MakisH:
> I would expect some information here: https://precice.org/adapter-openfoam-overview.html#cite

This originates from the adapter repository, there is a separate PR: https://github.com/precice/openfoam-adapter/pull/287

assets/openfoam-precice.bib
davidscn: Shouldn't be too much of an issue: Potential remedies: using `OpenFOAM\textsuperscript{\textregistered}` or add a space explicitly (e.g.) `\;`

Anyway, I'm fine with either version. Feel free to modify or merge the changes here.

MakisH: I don't want to insert too many non-content elements in there, as I don't know how different templates will behave. Let this be a problem of the journal.

PR #257, opened 2023-04-21
v3: Make dt mandatory in documentation

BenjaminRodenberg:
- [x] Merge https://github.com/precice/precice.github.io/pull/299


BenjaminRodenberg:
This implements the changes from https://github.com/precice/precice/pull/1622 in the documentation.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -40,17 +40,13 @@ If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it sam
 
 ## Experimental API for waveform iteration
 
-If we want to improve the accuracy by using waveforms, this requires an extension of the existing API, because we need a way to tell preCICE where we want to sample the waveform. For this purpose, preCICE offers an experimental API. Here, `readBlockVectorData` accepts an additional argument `relativeReadTime`. This allows us to choose the time where the waveform should be sampled:
+If we want to improve the accuracy by using waveforms, this requires you to tell preCICE where we want to sample the waveform. For this purpose, preCICE offers the argument `relativeReadTime` for all read data functions of the API:
```
fsimonis: Passing relative read time is no longer a choice.
This entire paragraph needs rewording to reflect this.

A visual example could be helpful here.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -86,7 +86,7 @@ while (not simulationDone()){ // time loop
   solverDt = beginTimeStep(); // e.g. compute adaptive dt
   dt = min(preciceDt, solverDt);
   if (precice.isReadDataAvailable()){
-    precice.readBlockVectorData(displID, vertexSize, vertexIDs, displacements);
+    precice.readBlockVectorData(displID, vertexSize, vertexIDs, preciceDt, displacements);
```
fsimonis: The surrounding code needs to be updated too. Especially regarding time step size

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -40,17 +40,13 @@ If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it sam
...
+`relativeReadTime` describes the time relatively to the beginning of the current time step. This means that `relativeReadTime = 0` gives us access to data at the beginning of the time step. By choosing `relativeReadTime > 0` we can sample data at later points. The maximum allowed `relativeReadTime` corresponds to the remaining time until the end of the current time window. Remember that the remaining time until the end of the time window is always returned when calling `preciceDt = precice.advance(dt)` as `preciceDt`. So `relativeReadTime = preciceDt` corresponds to sampling data at the end of the current time window (This is also what we do in the step by step guide to keep things simple).
```
fsimonis: We removed the return values of advance and initialize

fsimonis (commented):
There are some contextual changes that are missing. Especially the way we handle the max time step size has changed.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
BenjaminRodenberg: Took care of this via #258. In this PR I mainly wanted to take care of the part with the relative time. Another thing we need to update is the change of `readBlockVectorData` to `readData`. But this is again a todo for another PR or commit (and maybe somebody else :smirk:)

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: See #258. I think this should be resolved.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: I reformulated a bit via [d280ec6](https://github.com/precice/precice.github.io/pull/257/commits/d280ec6866ba1a1720cd6b0bb406c88a8517ba3a). 

One figure I could offer is the following:

Screenshot from 2023-05-28 15-13-58

Does this go into the right direction? I can tweak it a bit. If you have something different in mind: Let's do this in a follow up PR and get this PR merged, because there are some more dependencies (read data functions...).

fsimonis: That looks very good. The picture looks useful. Could you add it along it's sources?

BenjaminRodenberg: Done. [680eee7](https://github.com/precice/precice.github.io/pull/257/commits/680eee7bf841f44de0d6c9494372099af78d1115)

images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.pdf
uekerman, line 1: do we need the PDF as well?

pages/docs/couple-your-code/couple-your-code-gradient-data.md
```diff
@@ -75,7 +75,7 @@ double* stressGradient = new double[vertexSize * dim * dim]
 precice.initialize();
 
 while (not simulationDone()){ // time loop
-  precice.readBlockVectorData(displID, vertexSize, vertexIDs, displacements);
+  precice.readBlockVectorData(displID, vertexSize, vertexIDs, preciceDt, displacements);
```
uekerman: This code example has no `dt`, maybe you could remove the complete reading here.

pages/docs/couple-your-code/couple-your-code-implicit-coupling.md
```diff
@@ -58,7 +58,7 @@ while (precice.isCouplingOngoing()){
   preciceDt = precice.getMaxTimeStepSize();
   solverDt = beginTimeStep(); // e.g. compute adaptive dt
   dt = min(preciceDt, solverDt);
-  precice.readBlockVectorData(displID, vertexSize, vertexIDs, displacements);
+  precice.readBlockVectorData(displID, vertexSize, vertexIDs, preciceDt, displacements);
```
uekerman:
```suggestion
  precice.readBlockVectorData(displID, vertexSize, vertexIDs, dt, displacements);
```
no?
and maybe we could add a comment that we are reading at the end of the time step?

pages/docs/couple-your-code/couple-your-code-mesh-and-data-access.md
```diff
@@ -69,7 +69,7 @@ while (not simulationDone()){ // time loop
   preciceDt = precice.getMaxTimeStepSize();
   solverDt = beginTimeStep(); // e.g. compute adaptive dt
   dt = min(preciceDt, solverDt);
-  precice.readBlockVectorData(displID, vertexSize, vertexIDs, displacements);
+  precice.readBlockVectorData(displID, vertexSize, vertexIDs, preciceDt, displacements);
```
uekerman: dt?

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
uekerman: dt?
is there still a rebasing missing?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -40,19 +37,19 @@ If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it sam
...
+In the previous sections of the step-by-step guide we always used `relativeReadTime = preciceDt` where `preciceDt = precice.getMaxTimeStepSize()` points to the end of the current time window (see, for example ["Step 5 - Non-matching time step sizes"](couple-your-code-time-step-sizes.html)). However, the original purpose of `relativeReadTime` is exactly to offer the user an interface for sampling from waveforms. The figure below illustrates how providing different values `dt` for `relativeReadTime` allows to sample interpolated values at different points in time:
```
uekerman, line 47: I don't think that this is a good strategy (see also above): To tell users first how to do it the wrong way, and then the right way. Maybe we need to re-order our steps?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -40,19 +37,19 @@ If the Dirichlet participant $$\mathcal{D}$$ calls `readBlockVectorData`, it sam
...
+
+![API for relativeReadTime](images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.png)
+
+`relativeReadTime` describes the time relatively to the beginning of the current time step starting at $$\tau_n$$. This means that `dt = 0` gives us access to data at the beginning of the time step. By choosing `dt > 0` we can sample data at points in time after $$\tau_n$$. The maximum allowed `dt = preciceDt` corresponds to $$\tau_n$$ plus the remaining time until the end of the current time window (i.e. `preciceDt = precice.getMaxTimeStepSize()`).
```
uekerman, line 51: In the picture, you don't use "tau".

uekerman (changes requested):

images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.pdf
BenjaminRodenberg: I would see it as a nice to have, but sure: We can also remove it.

pages/docs/couple-your-code/couple-your-code-gradient-data.md
BenjaminRodenberg: Completely removing the reading here is a different issue. Let's use `preciceDt = precice.getMaxTimeStepSize()` here, because this is also what we tell our users, if they don't want to worry about the `dt`

BenjaminRodenberg: I would actually try to make this example more consistent with the other examples. Before, the whole topic of picking the right time step size was completely missing. Let's keep it consistent and just do what we always do:

```
preciceDt = precice.getMaxTimeStepSize();
solverDt = beginTimeStep(); // e.g. compute adaptive dt
dt = min(preciceDt, solverDt);
```

We already introduce it this way much earlier (Step 2) and I think if we suddenly leave it out here, it does not really help.

pages/docs/couple-your-code/couple-your-code-implicit-coupling.md
BenjaminRodenberg: Fixed. 3635024f76cd3bd36cddd12e046db2de949cffc4

pages/docs/couple-your-code/couple-your-code-mesh-and-data-access.md
BenjaminRodenberg: Fixed 3635024f76cd3bd36cddd12e046db2de949cffc4.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
BenjaminRodenberg: Fixed. 3635024f76cd3bd36cddd12e046db2de949cffc4

images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.png
BenjaminRodenberg, line 1: There is a $\tau_n$ here. The point about the $\tau$ is the following: Actually it can refer to any point in time. So $\tau$ could be $t_n$, if we are at the beginning of the window, but also any arbitrary point in time between $t_n$ and $t_{n+1}$.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: There is a $\tau_n$. See comment for the picture. But maybe I'm misunderstanding something?

images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.pdf
BenjaminRodenberg: Removed. 080925108965efdf5547fe854b1b2140e43ba73e

images/docs/couple-your-code/couple-your-code-waveform/APIRelativeReadTime.png
uekerman: I didn't get this from the picture, seems like you use $t_n$ and $\tau_n$ for the same thing.
Above you write  

> ... relatively to the beginning of the current time step starting at $\tau_n$

This should be $t_n$, no?

BenjaminRodenberg:
To be able to continue with this PR, we first need to take care of https://github.com/precice/precice.github.io/pull/299.

pages/docs/couple-your-code/couple-your-code-waveform.md
BenjaminRodenberg: Now that the waveform API is not experimental anymore but the default, let's put some information from here into "Step 5 - Non-matching time step sizes". I think this section needs some updates and refinement anyway. Additionally, I think that the topics that we currently explain in "Step 5 - Non-matching time step sizes" are definitely more advanced and exotic than some of the basic topics that we explain here. So yes: let's reorder some things.

We should still keep the more advanced background information (interpolation degree, higher-order time stepping) here to not provide too much information too early.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,45 @@ This procedure is independent of whether a serial or a parallel coupling scheme
 For parallel coupling, both solvers run together and everything happens simultaneously in both participants, while for serial coupling, the first participant needs reach the end of the window before the second one can start.
 {% endnote %}
 
+### Possible subcycling pitfall
+
+If you are using very small many time steps in one window, you might see the following error message:
+```
+preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows.
```
BenjaminRodenberg:
```suggestion
ERROR: preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows.
```

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,45 @@ This procedure is independent of whether a serial or a parallel coupling scheme
 For parallel coupling, both solvers run together and everything happens simultaneously in both participants, while for serial coupling, the first participant needs reach the end of the window before the second one can start.
 {% endnote %}
 
+### Possible subcycling pitfall
+
+If you are using very small many time steps in one window, you might see the following error message:
+```
+ERROR: preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows.
+```
+preCICE ignores differences smaller than machine precision when comparing the time where a coupling window ends and the participant time it knows from subsequent `advance(dt)` calls. Such small differences are usually negligible. They, however, might add up over multiple windows and become relevant. If preCICE detects this, it raises the error message shown above as a safeguard.
```
uekerman: Our numerical zero differences != machine precision. It is on purpose significantly larger.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,45 @@ This procedure is independent of whether a serial or a parallel coupling scheme
...
+precice.readData("FluidMesh", "Displacements", vertexIDs, dt, displacements);
+setDisplacements(displacements);
+solveTimeStep(dt);
+computeForces(forces);
+precice.writeData("FluidMesh", "Forces", vertexIDs, forces);
+// if dt = preciceDt, we will exactly reach the end of the window when calling advance
+precice.advance(dt);
+...
+```
+Here, a participant accepts `preciceDt`, even if it is bigger than `solverDt`, if the difference is not bigger than a certain tolerance `tol`. From the perspective of the participant this tolerance should be negligible with respect to the requirements (e.g. numerical stability)  that led to `solverDt`. With this modification the participant ensures that there is absolutely no difference between its internal time and the time used by preCICE and this eliminates the root cause for the error message from above.
```
uekerman: Not sure if we can really promise this. One might still end up with an diff of the order of the machine precision.
I have the feeling we are mixing two things here: avoiding too small time steps sizes and the drift.
Does the drift really need documentation?

pages/docs/couple-your-code/couple-your-code-mesh-and-data-access.md
```diff
@@ -33,8 +33,18 @@ void Participant::writeData(
...
+```cpp
+void readData(
+    precice::string_view          meshName,
+    precice::string_view          dataName,
+    precice::span<const VertexID> vertices,
+    double                        relativeReadTime,
+    precice::span<double>         values) const;
+```
+
+We will talk about the additional argument `relativeReadTime` in detail in [the section on time interpolation](couple-your-code-waveform.html).
```
fsimonis: At least a minimum amount of information would be useful here. Otherwise, we force the reader to switch pages in order to get an idea what the argument of this fundamental function even means.

This way we give a basic explanation and a hint on where to find more information later, which shouldn't interrupt the flow of reading the page.

```suggestion
The relative read time can be anthing from the current point in time `0` and the end of the time window `getMaxTimeStepSize()`. We will talk about the additional argument `relativeReadTime` in detail in [the section on time interpolation](couple-your-code-waveform.html).
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -35,27 +35,26 @@ Linear interpolation between coupling boundary conditions of the previous and th
...
 
-{% note %}
-As soon as time interpolation is used to obtain data for a `relativeReadTime` that does not refer to the end of the current window, the data from the end of the previous window might be used to compute the interpolated value. For the first time window, this means one should provide appropriate initial data, since otherwise preCICE will use zero-valued initial data. There are, however, situations where initial data is not necessarily needed. For example, if one uses zeroth order time interpolation without accessing the data value at the beginning of the time window. It is still recommended to always provide initial data, if available. See ["Step 7 - Data initialization"](couple-your-code-initializing-coupling-data.md) for details on data initialization.
-{% endnote %}
+## API for waveform iteration
```
fsimonis: This sounds like there is a dedicated API for the waveform iteration, which isn't the case.

Maybe `Using waveform iteration` or `In the adapter` or `In practise` would be easier to understand.

fsimonis (commented):
Some of these pages also still refer to `waveform-order` instead of `waveform-degree`.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,45 @@ This procedure is independent of whether a serial or a parallel coupling scheme
 For parallel coupling, both solvers run together and everything happens simultaneously in both participants, while for serial coupling, the first participant needs reach the end of the window before the second one can start.
 {% endnote %}
 
+### Possible subcycling pitfall
```
BenjaminRodenberg: I will remove this section from this PR. It's not really the same topic and it's I distracting from the main issue. We can deal with this elsewhere.

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: `Time interpolation in practise` :wink:

pages/docs/couple-your-code/couple-your-code-mesh-and-data-access.md
```diff
@@ -33,8 +33,18 @@ void Participant::writeData(
...
+```cpp
+void readData(
+    precice::string_view          meshName,
+    precice::string_view          dataName,
+    precice::span<const VertexID> vertices,
+    double                        relativeReadTime,
+    precice::span<double>         values) const;
+```
+
+The relative read time can be anthing from the current point in time `0` and the end of the time window `getMaxTimeStepSize()`. We will talk about the additional argument `relativeReadTime` in detail in [the section on time interpolation](couple-your-code-waveform.html).
```
uekerman:
```suggestion
The relative read time can be anything from the current point in time (`0`) to the end of the time window (`getMaxTimeStepSize())`. We will talk about the additional argument `relativeReadTime` in detail in [the section on time interpolation](couple-your-code-waveform.html).
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -5,10 +5,6 @@ keywords: api, adapter, time, waveform, subcycling, multirate
 summary: "With waveform iteration, you can interpolate coupling data in time for higher-order time stepping and more stable subcycling."
 ---
 
-{% experimental %}
-These API functions are work in progress, experimental, and are not yet released. The API might change during the ongoing development process. Use with care.
-{% endexperimental %}
-
 preCICE allows the participants to use subcycling – meaning: to work with individual time step sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a time step or you want to know how subcycling works in detail, see ["Step 5 - Non-matching time step sizes" of the step-by-step guide](couple-your-code-time-step-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling and advanced techniques for interpolation of coupling data inside of a time window.
```
```suggestion
preCICE allows the participants to use subcycling – meaning: to work with individual time step sizes smaller than the time window size. Note that participants always have to synchronize at the end of each *time window*. If you are not sure about the difference between a time window and a time step or you want to know how subcycling works in detail, see ["Step 5 - Non-matching time step sizes" of the step-by-step guide](couple-your-code-time-step-sizes.html). In the following section, we take a closer look at the exchange of coupling data when subcycling and advanced techniques for interpolation of coupling data are used inside of a time window.
```

uekerman (approved):
I would suggest to merge here and improve later.

PR #259, opened 2023-04-21
Use time step, not timestep.

BenjaminRodenberg:


BenjaminRodenberg:
@fsimonis for the review it would be great, if you could check with a fresh pair of eyes before merging. I hope I did not miss any leftover *timesteps*.

uekerman (approved):
OK for me, but I had only had a very quick look.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: Step 5 – Non-matching timestep sizes
-permalink: couple-your-code-timestep-sizes.html
+title: Step 5 – Non-matching time step sizes
+permalink: couple-your-code-time-step-sizes.html
```
MakisH, line 3: @BenjaminRodenberg we should also have a `redirect-from` for the old URL here.
This seems to work at the moment, but I assume only temporarily due to caching.

BenjaminRodenberg: @MakisH can you add this directly on master? Not sure how this works. Or is it just an additional line ?
```
redirect-from: couple-your-code-timestep-sizes.html
```

MakisH: Done: https://github.com/precice/precice.github.io/commit/39373f002a771df4eb5736e5fef0c91ec3b8752d

Issue #46, opened 2021-04-13
Improve footer on landing page and docs pages


chlorenz:
- [ ] split into two footer, one for landing page, one for docs pages
- [ ] check that licenses are correct (for preCICE on landing page)
- [x] replace icon from licensebuttons.com

MakisH:
I don't remember anymore why this issue was assigned also to me.
I am also not sure what we meant with "split into two footer, one for landing page, one for docs pages".
The CC BY-SA license at the footer looks correct.

chlorenz:
@MakisH You were added as the license expert. 
I might mis-remember this: There is preCICE the software, and there is the website - do they fall under the same license?

MakisH:
> I might mis-remember this: There is preCICE the software, and there is the website - do they fall under the same license?

The preCICE library is LGPL-3. The website content (and previous wiki) is Creative Commons Attribution 4.0 International. There are also more software components, each with a different license.

chlorenz:
Thanks. Got it, in that case I will only clarify the footer to reflect that "this work" means the website.

MakisH:
Looking at different issues, I see that many mention adding something to the footer. Could we prioritize making a basic skeleton for the footer?

Something like a 3-column list of items would already be very nice to extend.

MakisH:
As we are now reshuffling technical issues on the website, I would again like to prioritize this, as it would open some possibilities to make some pages and other information easier to find. I could extend the footer if there was a general skeleton.

chlorenz:
@MakisH I have implemented a 3 column layout in the footer, and populated it with some defaults. The final styling, e.g. the width of the columns, will depend on the content, do you already have some ideas what to display there?

MakisH:
Quick thoughts (definitely not perfect):

## Column 1

- [Cite](https://precice.org/fundamentals-literature-guide.html)
- [Contribute](https://precice.org/community-contribute-to-precice.html)
- [GitHub](https://github.com/precice)
- [Forum](https://precice.discourse.group/)

## Column 2

- License (I will create a new page)
- [Support](https://precice.org/community-support-precice.html)
- [Training](https://precice.org/community-training.html)
- [Impressum](https://precice.org/about.html)

## Column 3

Current CC-BY logo and last site last generated

chlorenz:
That sounds good, I'll implement three equally spaced columns. Maybe we can arrange the content like this:

[...]

PR #263, opened 2023-04-24
Add a page for previous versions

MakisH:
This adds a short page `Previous versions` below the `Roadmap`, which looks like this:

Screenshot from 2023-04-24 13-16-50

Motivation for some decisions:

- It's not a button below the sidebar, as this would be more complicated to implement, especially in a responsive way. It would, however, be more discoverable.
- It is under "Fundamentals" and next to the Roadmap, because is it the past (vs future of roadmap) and it shows up when first going to the docs.
- It is a complete page, aiming to later have entries for v3, v4, etc.
- It does not point to every feature release, as we don't have a PDF for every feature release and we generally keep the docs applicable to all feature releases of a major version.
- Linking to a previous revision of the GitHub Wiki does not really help, as every URL from the home page will lead again to the latest version.

Ideally, I would put links to each major version to the footer, which we don't have yet. #46

### DIFF
--- a/pages/docs/fundamentals/fundamentals-previous-versions.md
+++ b/pages/docs/fundamentals/fundamentals-previous-versions.md
@@ -7,6 +7,6 @@ summary: "Archives of this documentation"
 
 In regular intervals, we archive PDF exports of this documentation. Refer to those when you are looking for information regarding previous versions. These are part of the [preCICE Distribution](./installation-distribution.html). Alternatively, you can build [previous states of this website](https://github.com/precice/precice.github.io).
 
-- [preCICE v2.x](https://github.com/precice/precice.github.io/releases/download/v202211.0.0/libprecice2_2.2.5_docs_v202211.0.0.pdf)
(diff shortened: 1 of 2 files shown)


pages/docs/fundamentals/fundamentals-previous-versions.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Previous versions
+permalink: fundamentals-previous-versions.html
+keywords: pdf, export, v1, v2
+summary: "Archives of this documentation"
+---
+
+In regular intervals, we archive PDF exports of this documentation. Refer to those when you are looking for information regarding previous versions. These are part of the [preCICE Distribution](./installation-distribution.html). Alternatively, you can build [previous states of this website](https://github.com/precice/precice.github.io).
+
+- [preCICE v2.x](https://github.com/precice/precice.github.io/releases/download/v202211.0.0/libprecice2_2.2.5_docs_v202211.0.0.pdf)
```
uekerman:
```suggestion
- For preCICE v2.x: [libprecice2_2.2.5_docs_v202211.0.0.pdf](https://github.com/precice/precice.github.io/releases/download/v202211.0.0/libprecice2_2.2.5_docs_v202211.0.0.pdf)
```

At least the "pdf" should be somewhere in the link text.

uekerman (approved):
I like

Issue #268, opened 2023-05-05
Add section on tools on landing page


uekerman:
Two important tools around preCICE have reached a certain level of maturity such that we should advertise them more prominently. They might attract new user communities:
- Micro Manager
- FMI-preCICE Runner

We could add a dedicated section on such tools on the landing page, maybe between "adapters" and "what our community is saying". With short explanations on what users can do with these tools:
- "Perform many-to-one macro-micro simulations with the preCICE Micro Manager"
- "Couple FMUs to PDEs with the FMI-preCICE Runner"

Each tool could become a card with a visual (with a similar style as the features on the landing page), such that we could also extend this list in the future.

Links should go to dedicated pages under `docs/tooling` (WIP for both tools).

I am not sure whether to add more of our tools here (config visualizer, ASTE, ...). These are hard to understand without knowing preCICE already.

MakisH:
> Each tool could become a card with a visual (with a similar style as the features on the landing page)

Actually, why not simply extend the "features" section? Considering these tools as part of the preCICE ecosystem, these are features of what the user sees as "preCICE".

As we need more and more sections on the landing page, I am trying to think of ways to reduce their number, to not make the page too long.

uekerman:
Could be an alternative, yes.

MakisH:
We could have a section "Versatile coupling" between "prepared ..." and "ready-to-use"

- Surface and volume coupling (needs documentation entrypoint #473)
- Coupling to FMUs
- Coupling to particles
- Many-to-one coupling
- Mixed dimensions

The "Learn more" could link to a new section in https://precice.org/docs.html explaining how to approach the documentation depending on the coupling type you need.

I could then design fitting pictures.

What do you think?

uekerman:
> We could have a section "Versatile coupling" between "prepared ..." and "ready-to-use"

Sounds good.

And then a card for each of your items? Sounds good to me.

> The "Learn more" could link to a new section in https://precice.org/docs.html explaining how to approach the documentation depending on the coupling type you need.

This point I don't understand yet. It would be one "learn more" for each card, right? For some of your items, we would already have good landing pages.

MakisH:
> > The "Learn more" could link to a new section in https://precice.org/docs.html explaining how to approach the documentation depending on the coupling type you need.
> 
> This point I don't understand yet. It would be one "learn more" for each card, right? For some of your items, we would already have good landing pages.

Yes, one for each card. Leftover from my first draft, where I was thinking of one feature card for all.

PR #269, opened 2023-05-05
Add hint to GPUs on landing page

uekerman:
What do you think?

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -125,6 +125,7 @@ layout: landing_page
             <div class="panel-body">
               <ul class="usp">
                 <li class="usp">Pure peer-to-peer approach</li>
+                <li class="usp">Support of heterogenous hardware (CPU/GPU)</li>
                 <li class="usp">Efficient also on a laptop</li>
               </ul>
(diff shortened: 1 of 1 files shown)


davidscn:
Not sure if that's a bit too much for what we have right now.

uekerman:
Is it? We already could support solvers on GPU before. Now, parts of preCICE can also run on GPUs, so one could exploit a heterogeneous cluster. It's not that we promise here a lot.

MakisH:
Similarly to #268, we could add another feature card, where we could also explain the state.

Maybe something like "GPU-accelerated mappings".

One could even extend the `Radial-basis function data mapping` to `Radial-basis function data mapping (CPU and GPU)`.

davidscn:
If we interpret the sentence more from the user perspective, it's of course true that people relying on GPUs can use preCICE. However, coupling is still carried out on a CPU.
We can also add the sentence and explain in another section more in detail what's meant here.

uekerman:
Let's merge as discussed in the telco.

PR #270, opened 2023-05-08
Add documentation for micro-manager installation

erikscheurer:
The first release of the micro-manager is coming up, adding information on installation on the website would be ideal. 
However, as the micro manager is not directly a language binding or an adapter, I am not sure where to put the information. The two options I see are with the provided adapters or in the installation tab under the language bindings.


uekerman:
IMO, this should go under "tools" and only be linked. The actual content should be in the micro-manager repo. Similarly to e.g. the OpenFOAM adapter.

_data/sidebars/docs_sidebar.yml
```diff
@@ -241,6 +241,10 @@ entries:
       url: /tooling-rbf-shape.html
       output: web, pdf
 
+    - title: Micro-Manager
```
uekerman:
```suggestion
    - title: Micro Manager
```

maybe even:

```suggestion
    - title: Micro manager
```

IshaanDesai: So far we have stuck to **Micro Manager** as the name of the software. I would suggest sticking to that here too.

_data/sidebars/docs_sidebar.yml
```diff
@@ -241,9 +241,31 @@ entries:
...
+          url: /tooling-micro-manager-usage-code-changes.html
+          output: web, pdf
+
+        - title: Configuration
+          url: /tooling-micro-manager-usage-configuration.html
+          output: web, pdf
+
+        - title: Running
+          url: /tooling-micro-manager-usage-running.html
+          output: web, pdf
```
uekerman: Do we really need this second hierarchy level `Usage`? Even large packages such as the OpenFOAM adapter don't have this.

IshaanDesai: `Usage` seems unnecessary as there are only a few section in it. Lets remove it @erikscheurer

erikscheurer: We do not have the second hierachy level anymore, but the Usage page was supposed to give the user an overview over what has to be done to adapt their code, do you think the other pages are expressive enough by themselves?

IshaanDesai: It should be pretty clear from the order that the sections are instructions to be run one after the other. If in the future we get feedback about the section not being clear enough, we can introduce the Usage level again.

.gitmodules
```diff
@@ -10,3 +10,6 @@
 	path = imported/aste
 	url = https://github.com/precice/aste.git
 	branch = develop
+[submodule "imported/micro-manager"]
+	path = imported/micro-manager
+	url = https://github.com/precice/micro-manager.git
```
MakisH, line 15: Why no `branch` for the `micro-manager` here?
I guess the workflow is failing because the pages don't yet exist in the default branch (main): https://github.com/precice/micro-manager/tree/main/docs

```
       Subproject: imported/openfoam-adapter/docs
      Adding pages: README.md, openfoam-support.md, config.md, get.md, extend.md
         Tutorials: No subproject found at imported/micro-manager/docs
        Subproject: imported/micro-manager/docs
                    ------------------------------------------------
      Jekyll 4.2.2   Please append `--trace` to the `build` command 
                     for any additional information or backtrace. 
                    ------------------------------------------------
/home/runner/work/precice.github.io/precice.github.io/website/_plugins/subprojects.rb:22:in `block in generate': No subproject found at imported/micro-manager/docs (RuntimeError)
```

IshaanDesai: Thanks for noticing this! The website building still fails, but this was definitely an issue.

erikscheurer:
The pipeline fails because the micro-manager submodule is imported at [this](https://github.com/precice/micro-manager/commit/7e9c3e78ae226023e3e0ed5fcf7663e5d133a859) commit for some reason and not at develop as specified in `.gitmodules`.
Any ideas why this is the case?

.gitmodules
MakisH: I don't know, but I am confused by this line:

```
Tutorials: No subproject found at imported/micro-manager/docs
```

Where should the `Tutorials:` come from?

IshaanDesai:
> The pipeline fails because the micro-manager submodule is imported at [this](https://github.com/precice/micro-manager/commit/7e9c3e78ae226023e3e0ed5fcf7663e5d133a859) commit for some reason and not at develop as specified in `.gitmodules`. Any ideas why this is the case?

Perhaps @MakisH has a clue.

erikscheurer:
Ah it was already described [in the README](https://github.com/precice/precice.github.io#update-submodules) :)
So as I understand, git submodules can't follow a branch but only save a current commit. And therefore every time we update the documentation, we have to open a PR here right?

MakisH:
> So as I understand, git submodules can't follow a branch but only save a current commit.

Correct.

> And therefore every time we update the documentation, we have to open a PR here right?

Almost... The website points to specific submodule commits, but triggering [this workflow on the website repository](https://github.com/precice/precice.github.io/blob/master/.github/workflows/update-submodules.yml) updates the submodules. This is done automatically by workflows in each of the source repositories, with [such a workflow](https://github.com/precice/openfoam-adapter/blob/develop/.github/workflows/update-website.yml). You actually should add a similar file here as well.

PR #266, opened 2023-05-03
Update profiling section

fsimonis:
This PR updates the profiling section according to https://github.com/precice/precice/pull/1419 

TODO:

* [x] recreate output after the preCICE PR is merged
* [x] add example of a supemuc run
* [x] add comparison outputs between all and fundamental modes

PR #272, opened 2023-06-02
Fix minor issues in tooling/performance-analysis

MakisH:
Following up on https://github.com/precice/precice.github.io/pull/266 (sorry for the late review).

@fsimonis did I understand the file naming scheme correctly?

I have not tested it yet, as I am not sure which components work with the current develop at the moment. I may end up updating the elastic-tube-1d in the process.

### DIFF
--- a/pages/docs/tooling/tooling-performance-analysis.md
+++ b/pages/docs/tooling/tooling-performance-analysis.md
@@ -98,7 +98,7 @@ The general workflow looks as follows:
 
 1. Run the simulation
 2. Merge the event files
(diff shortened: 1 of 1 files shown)


MakisH:
Checked the website build, and the image is now loaded correctly.
@fsimonis in case I misunderstood the naming scheme (`participant-rank-file_number`), fix in a follow-up.

fsimonis (approved):

fsimonis:
File nameing scheme is correct, but shouldn't matter much to the user. Commands allow filtering ranks

PR #275, opened 2023-06-14
Make new FF tutorials discoverable

MakisH:
Makes the tutorials introduced in https://github.com/precice/tutorials/pull/326 discoverable.

@thesamriel FYI

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -221,6 +221,8 @@ subprojects:
   - imported/tutorials/perpendicular-flap
   - imported/tutorials/turek-hron-fsi3
   - imported/tutorials/partitioned-pipe
+  - imported/tutorials/partitioned-backwards-facing-step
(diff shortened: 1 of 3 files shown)


MakisH:
Checked locally, it looks as expected.

However, the note _"Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/partitioned-backwards-facing-step)."_ points, by default, to master, which will only provide the files once we update the tutorials master branch (e.g., with v3).

uekerman (approved):
> However, the note "Get the [case files of this tutorial](https://github.com/precice/tutorials/tree/master/partitioned-backwards-facing-step)." points, by default, to master, which will only provide the files once we update the tutorials master branch (e.g., with v3).

Yes, let's live with this for the moment. Eventually, we need a better solution.
Probably, we could / should only merge such (feature) website updates once we release the tutorials.

PR #277, opened 2023-06-16
Include FMI runner in Tooling

LeonardWilleke:
Add instructions on the FMI runner on precice.org in the Tooling section

In combination with changes in the FMI runner repo: https://github.com/precice/fmi-runner/pull/6

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,3 +14,7 @@
 	path = imported/micro-manager
 	url = https://github.com/precice/micro-manager.git
 	branch = develop
+[submodule "imported/fmi-runner"]
+	path = imported/fmi-runner
+	url = https://github.com/precice/fmi-runner.git
+	branch = develop
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,5 +16,5 @@
 	branch = develop
 [submodule "imported/fmi-runner"]
 	path = imported/fmi-runner
-	url = https://github.com/precice/fmi-runner.git
+	url = https://github.com/precice/fmi-runner
 	branch = develop
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -232,6 +232,10 @@ entries:
     - title: Config visualization
       url: /tooling-config-visualization.html
       output: web, pdf
+      
+    - title: FMI runner
+      url: /tooling-fmi-runner.html
+      output: web, pdf
 
       subfolders:
       - title: Micro Manager
@@ -257,10 +261,6 @@ entries:
         - title: Running
           url: /tooling-micro-manager-running.html
           output: web, pdf
-
-    - title: FMI runner
-      url: /tooling-fmi-runner.html
-      output: web, pdf
     
     - title: Performance analysis
       url: /tooling-performance-analysis.html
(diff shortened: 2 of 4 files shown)


_data/sidebars/docs_sidebar.yml
```diff
@@ -258,6 +258,10 @@ entries:
           url: /tooling-micro-manager-running.html
           output: web, pdf
 
+    - title: FMI runner
+      url: /tooling-fmi-runner.html
+      output: web, pdf
+
```
uekerman: I would go alphabetically here, meaning, move above Micro Manager

uekerman (approved):
Let's also add a line here: https://precice.org/tooling-overview.html

.gitmodules
```diff
@@ -14,3 +14,7 @@
 	path = imported/micro-manager
 	url = https://github.com/precice/micro-manager.git
 	branch = develop
+[submodule "imported/fmi-runner"]
+	path = imported/fmi-runner
+	url = https://github.com/precice/fmi-runner.git
+	branch = develop
```
MakisH: Apart from the fact that `develop` does not yet include a `docs/` folder, specifying this in the `.gitmodules` does not seem to be enough. We probably need to:

- remove these lines
- run `git submodule add https://github.com/precice/fmi-runner imported/fmi-runner` to properly add the submodule
- run `git submodule set-branch --default update-docu imported/fmi-runner` to set the branch (switch to `develop` once the upstream PR is merged).
- commit the result and push

If this works out, doing `git submodule update --remote --merge` will report that:

```
Submodule path 'imported/fmi-runner': checked out '09f4697fb601e63942771252928d83688963e2ff'
```

and running `ls imported/fmi-runner` will show a `docs` folder.

Seems to be running for me locally, but I let you do it here for practice and cross-checking.

If you push with the `update-docu` branch in the submodule, the CI should pass already.

LeonardWilleke: Thank you for the instructions. It mostly worked out. I merged the upstream PR to use the `develop` branch and used it here.

The report from `git submodule update --remote --merge` was a bit different:
```
Submodule path 'imported/fmi-runner': merged in '680ce077d4349c43d2181d9f34316887df71af7e'

```
But I see the `docs` folder in `imported/fmi-runner` and an entry was added to the `.gitmodules` file

MakisH: Now the CI seems to be fetching the submodule, but Jekyll still complains that `No subproject found at imported/fmi-runner/docs`.

The submodule (as I see in the diff view of this PR) seems to be pointing to https://github.com/precice/fmi-runner/tree/4c93437febe0399575396513efe0ea7bcdb66023

I think that, now that you merged the upstream, you need to update the submodule and it should work.

LeonardWilleke: Does the latest commit solve the problem?

MakisH: The module seems to be pointing to the latest commit, so everything should be fine.

But the CI has not been executed on the last commit, so I can only guess that.

Note that there are some merge conflicts you could resolve. This would also trigger the CI.

LeonardWilleke: Thank you, I missed that.

MakisH: The CI succeeds now! Should I merge, or wait for anything else?

LeonardWilleke: Please merge, thanks!

PR #296, opened 2023-10-23
Add recommendation for `imvj-restart-mode` in configuration of acceleration

Fujikawas:
To make users be aware of that the `IQN-IMVJ` method would result in quadratic storage and runtime complexity without proper `restart-mode`, which is quite expensive for large-scale simulations.


pages/docs/configuration/configuration-acceleration.md
```diff
@@ -87,5 +87,6 @@ For quasi-Newton, the configuration is more involved and requires some attention
...
+* While the `IQN-IMVJ` method exhibits excellent convergence properties as a result of implicit reuse of prior information, it also causes quadratic storage and runtime complexity because of the explicit computation of Jacobian matrix. To make this method practical for large-scale simulations, we introduced a set of `imvj-restart-mode` variants, which reduce the method's complexity to a linear one. In particular, the `RS-SVD` mode gives good performance comparing to other options while retaining the original MV convergence properties, although its performance slightly depends on the choose of the truncation parameter.
```
uekerman:
```suggestion
* While the `IQN-IMVJ` method exhibits excellent convergence properties as a result of implicit reuse of prior information, it also causes quadratic storage and runtime complexity because of the explicit computation of Jacobian matrix. To make this method practical for large-scale simulations, we introduced a set of `imvj-restart-mode` variants, which reduce the method's complexity to a linear one. In particular, the `RS-SVD` mode gives good performance comparing to other options while retaining the original multi-vector convergence properties, although its performance slightly depends on the choice of the truncation parameter.
```

uekerman (approved):

PR #308, opened 2023-11-21
Rewrite mapping docs

davidscn:
...according to preCICE v3

Closes https://github.com/precice/precice/issues/1608


images/docs/configuration/doc-mapping-options.svg
uekerman, line 1: Why is "iterative solving" and "direct solving" in v3? Since there were GPU updates? Is this clear?

pages/docs/configuration/configuration-mapping.md
```diff
@@ -23,14 +23,20 @@ A participant needs to `use` at least two meshes to define a mapping between bot
 
 ![Mapping configuration](images/docs/configuration/doc-mapping.png)
 
-Mappings can be defined in two `directions`, either `read` or `write`:
+The `provide` mesh and `receive` mesh are then assigned to the `from` and `to` slot in the mapping configuration to indicate the source mesh, on which data is written, and the target mesh, on which data is read. In addition to source and target mesh, each mapping defines a `direction`, which can either be `read` or `write`:
```
uekerman: Do we need the additional terms "source" and "target"?

pages/docs/configuration/configuration-mapping.md
```diff
@@ -23,14 +23,20 @@ A participant needs to `use` at least two meshes to define a mapping between bot
...
 * `write` mappings are executed _after_ you have written data.
 
-Furthermore, mappings come int two types of `constraint`: `consistent` and `conservative`:
+The `direction` indicates, how the defined meshes are used from the participant perspective: for a `read` mapping, the participant reads data from the `provide` mesh (`MyMesh1`), for a `write` mapping, the participant writes data to the `provide` mesh. Therefore, the `direction` is related to the `exchange` tag of the `coupling-scheme`, as the remote mesh defined by the 'other' participant (e.g. `MyMesh2`) needs to be the mesh used in `<exchange mesh="MyMesh2" ...`. In principle, each `read` mapping can be transformed into a `write` mapping (ignoring the [restrictions for parallel participants for now](configuration-mapping.html#restrictions-for-parallel-participants)) by shifting the mapping tag on the 'other' involved participant (e.g. `MySolver2`). Depending on the configuration, data mapping might be computationally demanding. In preCICE the mapping is executed on the participant, where the mapping tag is defined in the configuration file.
```
uekerman: I have the feeling that talking about how this is connected to coupling schemes here already could confuse more than it helps. Maybe moving this further down into a note box could be better?

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+Methods marked in orange have been added more recently in preCICE and are available as annotated by the corresponding preCICE version.
+
+{% note %}
+Chapter 3.2 (Data mapping) of the preCICE version 2 [reference paper](https://doi.org/10.12688/openreseurope.14445.2) explains and compares a selection of projection-based methods and kernel methods.
+{% endnote %}
+
+### Projection-based methods
+
+Projection-based data mapping methods are typically cheap to compute as they don't involve solving expensive linear systems as opposed to the kernel methods. The basic variants, which operates solely on vertex data is `nearest-neighbor` mapping. All other variants require, additional information from the user, as stated in the overview figure.
```
uekerman:
```suggestion
Projection-based data mapping methods are typically cheap to compute as they don't involve solving expensive linear systems as opposed to the kernel methods. The basic variant, which operates solely on vertex data, is `nearest-neighbor` mapping. All other variants require additional information from the user, as shown in the overview figure above.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
 
 ![different mapping variants visualised](images/docs/configuration-mapping-variants.png)
 
-## Radial-basis function mapping
+The available methods are:
```
uekerman:
```suggestion
Available methods are:
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+* `nearest-neighbor`: A first-order method, which is fast, easy to use, but, of course, has numerical deficiencies.
+* `nearest-projection`: A (depending on how well the geometries match at the coupling interface) second-order method, which first projects onto surface mesh elements (first-order step) and, then, uses linear interpolation within each element (second-order step) as illustrated in the figure below. The method is still relatively fast and numerically clearly superior to `nearest-neighbor`. The downside is that mesh connectivity information needs to be defined, i.e. in the adapter, the participant needs to tell preCICE about edges in 2D and edges, triangles, or quadrilaterals in 3D. On the [mesh connectivity page](couple-your-code-defining-mesh-connectivity.html), we explain how to define connectivity. If no connectivity information is provided, `nearest-projection` falls back to a `nearest-neighbor` mapping.
```
uekerman:
```suggestion
* `nearest-projection`: A (depending on how well the geometries match at the coupling interface) second-order method, which first projects onto surface mesh elements (first-order step) and, then, uses linear interpolation within each element (second-order step) as illustrated in the figure above. The method is still relatively fast and numerically clearly superior to `nearest-neighbor`. The downside is that mesh connectivity information needs to be defined, i.e. in the adapter, the participant needs to tell preCICE about edges in 2D and edges, triangles, or quadrilaterals in 3D. On the [mesh connectivity page](couple-your-code-defining-mesh-connectivity.html), we explain how to define connectivity. If no connectivity information is provided, `nearest-projection` falls back to a `nearest-neighbor` mapping.
```
?

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+* `linear-cell-interpolation`: Instead of mapping to surface-elements as the `nearest-projection`, `linear-cell-interpolation` uses volumetric elements, i.e., it is designed for volumetric coupling, where the coupling mesh is a region of space and not a domain boundary. It interpolates on triangles in 2D and on tetrahedra in 3D. Hence, connectivity information for volumetric elements needs to be defined. If none are found, it falls back on `nearest-projection` or `nearest-neighbor` (depending on the available connectivity information). The method was developed in the [Master's thesis of Boris Martin](https://mediatum.ub.tum.de/doc/1685618/1685618.pdf), where more detailed information are available.
```
uekerman:
```suggestion
* `linear-cell-interpolation`: Instead of mapping to surface-elements as the `nearest-projection`, `linear-cell-interpolation` uses volumetric elements, i.e., it is designed for volumetric coupling, where the coupling mesh is a region of space and not a domain boundary. It interpolates on triangles in 2D and on tetrahedra in 3D. Hence, connectivity information for volumetric elements needs to be defined. If none are found, it falls back on `nearest-projection` or `nearest-neighbor` (depending on the available connectivity information). The method was developed in the [Master's thesis of Boris Martin](https://mediatum.ub.tum.de/doc/1685618/1685618.pdf), where more detailed information is available.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+* `nearest-neighbor-gradient`: A second-order method, which uses the same algorithm as nearest-neighbor with an additional linear approximation using gradient data. This method requires additional gradient data information. On the [gradient data page](couple-your-code-gradient-data.html), we explain how to add gradient data to the mesh. This method is only applicable with the `consistent` constraint. The method was developed [Master's thesis of Boshra Ariguib](http://dx.doi.org/10.18419/opus-12128), where more detailed information are available.
```
uekerman:
```suggestion
* `nearest-neighbor-gradient`: A second-order method, which uses the same algorithm as nearest-neighbor with an additional linear approximation using gradient data. This method requires additional gradient data information. On the [gradient data page](couple-your-code-gradient-data.html), we explain how to add gradient data to the mesh. This method is only applicable with the `consistent` constraint. The method was developed [Master's thesis of Boshra Ariguib](http://dx.doi.org/10.18419/opus-12128), where more detailed information is available.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+{% endnote %}
 
-```xml
-<mapping:rbf-thin-plate-splines direction="read" from="MyMesh2" to="MyMesh1" constraint="consistent"/>
-```
+* `rbf-pum-direct`, which breaks down the mapping problem in smaller clusters, solves these clusters locally and blends them afterwards together to recover a global solution. This mapping version only needs the linear-algebra library Eigen and the used linear solver is a dense solver in each cluster (actually the same as for `rbf-global-direct`, i.e., it is beneficial to configure strictily positive definite basis-functions). The mapping is specificially designed for large mapping problems and runs fully mpi-parallel. To configure the accuracy of the mapping, the number of `vertices-per-cluster` can be increased. The method can only handle sufficiently matching geometries and problems might occur if larger gaps exist between the coupling meshes. Further information, including some performance comparisons, can be found in the [talk of the preCICE workshop 2023](https://youtu.be/df-JMl7UxRg?si=18X3LFTIepmrtAMc).
```
uekerman: Should we mention that PUM is much faster than global RBF or similar?

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
 
-For a complete overview of all basis function, refer to [this paper](https://www.sciencedirect.com/science/article/pii/S0045793016300974), page 5.
+![RBF alias options](images/docs/configuration/doc-mapping-rbf-alias.svg)
 
-The interpolation problem might not be well-defined if you map along an axis-symmetric surface. This means, preCICE tries to compute, for example, a 3D interpolant out of 2D information. If so, preCICE throws an error `RBF linear system has not converged` or `Interpolation matrix C is not invertible`. In this case, you can restrict the interpolation problem by ignoring certain coordinates, e.g. `x-dead="true"` to ignore the x coordinate.
+An exemplary rbf mapping configuration would look as follows
```
uekerman:
```suggestion
An RBF mapping configuration could look as follows
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+<mapping:rbf direction="read" from="MyMesh2" to="MyMesh1" constraint="consistent">
+ <basis-function:compact-polynomial-c6 support-radius="1.8">
+</mapping:rbf>
+```
+
+The basis-function has to be defined as a subtag in all kernel methods. In this example the basis-function `compact-polynomial-c6` is used with a support radius of `r=1.8`.
 
 {% note %}
-All data mappings are executed during `advance` and not in `readData`, cf. the section on  [how to couple your own code](couple-your-code-overview.html).
+We recommend to use the alias tag, as long as there are no further requirements regarding the desired mapping method. preCICE reports initially, which mapping method was selected for the alias tag. However, the decision might vary between different versions. If you want to ensure that a specific method is used, please use the corresponding mapping tag.
```
uekerman:
```suggestion
We recommend to use the alias tag, as long as there are no further requirements regarding the desired mapping method. preCICE reports initially, which mapping method was selected for the alias tag. However, the decision might vary between different versions. If you want to ensure that a specific method is used, use the corresponding mapping tag.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
 {% endnote %}
 
+The configuration of the basis-function is problem dependent. In general, preCICE offers basis function with global and local support:
```
uekerman:
```suggestion
The configuration of the basis-function is problem-dependent. In general, preCICE offers basis function with global and local support:
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+* Basis-functions with global support (such as `thin-plate-splines`) are easier to configure as no further parameter needs to be set. However, typically it pays off in terms of accuracy to configure a basis-function according to your problem dimensions and shape.
+* Basis functions with local support need either the definition of a `support-radius` (such as for `compact-polynomial-c2`) or a `shape-parameter` (such as for `gaussian`). Selecting a larger and larger support radius leads to a more and more flat basis-function and -in theory- to more and more accurate results. However, the resulting linear system becomes more and more ill-conditioned such that the linear-solver might fail, if the basis-function is configured too flat for the given vertex distribution.
```
uekerman:
```suggestion
* Basis functions with local support need either the definition of a `support-radius` (such as for `compact-polynomial-c2`) or a `shape-parameter` (such as for `gaussian`). Selecting a larger and larger support radius leads to a more and more flat basis-function and -in theory- to more and more accurate results. However, the resulting linear system becomes more and more ill-conditioned such that the linear solver might fail eventually.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+ASTE and our [ASTE tutorial](tutorials-aste-turbine.html) enable full insight into the accuracy of the configured mapping method.
+
+#### Execution backends
+
+{% experimental %}
+Although the feature is well tested in preCICE, it relies on an unreleased version of Ginkgo. Hence, this feature might change in the future.
+{% endexperimental %}
+
+Starting from version 3, preCICE offers to execute `mapping:rbf-global...` on a different executor backends using the linear-operator library Ginkgo
```
uekerman:
```suggestion
Starting from version 3, preCICE offers to execute `mapping:rbf-global...` on different executor backends using the linear-operator library Ginkgo
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+![RBF executors](images/docs/configuration/doc-mapping-rbf-executors.svg)
+
+To configure the executor, an additional subtag can be used in the mapping configuration
```
uekerman:
```suggestion
To configure the executor, an additional subtag can be used in the mapping configuration:
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+```xml
+<mapping:rbf direction="read" from="MyMesh2" to="MyMesh1" constraint="consistent">
+ <basis-function:compact-polynomial-c6 support-radius="1.8">
+ <executor:cuda gpu-device-id="0"/>
+</mapping:rbf>
+```
+
+More details on the feature can be found in [the ECCOMAS coupled proceeding](https://doi.org/10.23967/c.coupled.2023.016).
```
uekerman:
```suggestion
More details on the feature can be found in [Schneider et al. 2023](https://doi.org/10.23967/c.coupled.2023.016).
```

uekerman (approved):
Nice job. Minor tweaks below.

@MakisH For a feature that has undergone such dramatic changes in configuration, should we add a version box here mentioning some of the old configurations and pointing to the v2 pdf?

MakisH:
> @MakisH For a feature that has undergone such dramatic changes in configuration, should we add a version box here mentioning some of the old configurations and pointing to the v2 pdf?

I think that, when we release v3, we should point to the v2 PDF in the banner for a while, to make it visible. We could then also add a "Tip" alert box in the mapping config to point to the v2 PDF: https://precice.org/docs-meta-cheatsheet.html

images/docs/configuration/doc-mapping-options.svg
MakisH: Without having looked at the text yet, what do the versions mean? When the feature was intoduced? But then, the `(v3.0.0)` is not clear to me in the iterative and direct. What is the meaning of the diamond?

Looking at the more detailed figure, I understand you want to say that some features are supported since v3. Why not have a complete figure, instead? This would give a good overview and be clearer.

The arrows probably don't mean anything here. I would rather use simple lines and start from "Mapping methods".

If you want to include GPUs in the picture, you can do that with additional visual elements (e.g., a colored dot next to each mapping).

Why are some circles filled and some not?

images/docs/configuration/doc-mapping-rbf-executors.svg
MakisH, line 1: Why are both "iterative solving" and "direct solving" point to the same GPU bubble?
Shouldn't it be two bubbles with duplicate content? Aren't these two different mapping implementations?

pages/docs/configuration/configuration-mapping.md
MakisH: I think a configuration visualization would help: the user needs to understand that the arrows need to follow the same direction. This is also a nice way to remind people that this exists.

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
             T=1           T=1
 ```
 
-For a sequential participant, any combination of `read`/`write`-`consistent/conservative` is valid. For a parallel participant (i.e. a `master` tag is defined), only `read`-`consistent` and `write`-`conservative` is possible. More details are given [further below](configuration-mapping.html#restrictions-for-parallel-participants).
+* `scaled-consistent-surface` and `scaled-consistent-volume` constraint: `scaled-consistent` constraints are used for intensive quantities (just as the `consistent` constraint) where conservation of integral values (surface or volume) is necessary (e.g. velocities when the mass flow rate needs to be conserved). The mapping executes a `consistent` mapping in a first step, and corrects the result by a subsequent scaling step using the integral data sum to ensure the conservation of the integral sum on the input mesh and the output mesh. To use the `scaled-consistent-surface` constraint, surface connectivity on input and output meshes are required. To use the `scaled-consistent-volume` constraint, volumetric connectivity on input and output meshes are required.
```
MakisH:
```suggestion
* `scaled-consistent-surface` and `scaled-consistent-volume` constraint: `scaled-consistent` constraints are used for intensive quantities (just as the `consistent` constraint) where conservation of integral values (surface or volume) is necessary (e.g. velocities when the mass flow rate needs to be conserved). The mapping executes a `consistent` mapping in a first step, and corrects the result by a subsequent scaling step using the integral data sum to ensure the conservation of the integral sum on the input mesh and the output mesh. To use the `scaled-consistent-surface` constraint, surface connectivity on input and output meshes is required. To use the `scaled-consistent-volume` constraint, volumetric connectivity on input and output meshes is required.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
-* `linear-cell-interpolation`: A second-order method similar to `nearest-projection`, but designed for volumetric coupling, where the coupling mesh is a region of space and not a domain boundary. It interpolates on triangles in 2D and on tetrahedra in 3D. If none are found, it falls back on `nearest-projection`.
-* `nearest-neighbor-gradient`: A second-order method, which uses the same algorithm as nearest-neighbor with an additional linear approximation using gradient data. This method requires additional gradient data information. On the [gradient data page](couple-your-code-gradient-data.html), we explain how to add gradient data to the mesh. This method is only applicable with the `consistent` constraint.
-* Radial-basis function mapping. Here, the configuration is more involved, so keep reading.
+![Mapping options](images/docs/configuration/doc-mapping-options.svg)
+
+Methods marked in orange have been added more recently in preCICE and are available as annotated by the corresponding preCICE version.
```
MakisH: Since you are giving the versions, I don't think that the colors are important. What is "recent"? When should we un-color these bubbles?

pages/docs/configuration/configuration-mapping.md
MakisH:
```suggestion
* `rbf-pum-direct`, which breaks down the mapping problem in smaller clusters, solves these clusters locally and blends them afterwards together to recover a global solution. This mapping version only needs the linear-algebra library Eigen and the used linear solver is a dense solver in each cluster (actually the same as for `rbf-global-direct`, i.e., it is beneficial to configure strictily positive definite basis-functions). The mapping is specificially designed for large mapping problems and runs fully mpi-parallel. To configure the accuracy of the mapping, the number of `vertices-per-cluster` can be increased. The method can only handle sufficiently matching geometries and problems might occur if larger gaps exist between the coupling meshes. Further information, including some performance comparisons, can be found in [this talk of the preCICE workshop 2023](https://youtu.be/df-JMl7UxRg?si=18X3LFTIepmrtAMc).
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,61 +56,104 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+
+* Basis-functions with global support (such as `thin-plate-splines`) are easier to configure as no further parameter needs to be set. However, typically it pays off in terms of accuracy to configure a basis-function according to your problem dimensions and shape.
+* Basis functions with local support need either the definition of a `support-radius` (such as for `compact-polynomial-c2`) or a `shape-parameter` (such as for `gaussian`). Selecting a larger and larger support radius leads to a more and more flat basis-function and -in theory- to more and more accurate results. However, the resulting linear system becomes more and more ill-conditioned such that the linear-solver might fail, if the basis-function is configured too flat for the given vertex distribution.
+
+ASTE and our [ASTE tutorial](tutorials-aste-turbine.html) enable full insight into the accuracy of the configured mapping method.
+
+#### Execution backends
+
+{% experimental %}
+Although the feature is well tested in preCICE, it relies on an unreleased version of Ginkgo. Hence, this feature might change in the future.
```
MakisH: Clarify "the feature":

```suggestion
Although the executors feature is well-tested in preCICE, it relies on an unreleased version of Ginkgo. Hence, this feature might change in the future.
```

or just write:

```suggestion
Although this feature is well-tested in preCICE, it relies on an unreleased version of Ginkgo. Hence, this feature might change in the future.
```

MakisH (commented):
Good overall, it definitely helps getting a better overview of the zoo of features.

I would like to read it again once all the current suggestions have been considered.

pages/docs/configuration/configuration-mapping.md
davidscn: I removed source and target now

pages/docs/configuration/configuration-mapping.md
davidscn: I removed the coupling scheme hint. The config visualizer is already mentioned below (the information here was a bit duplicated), I highlighted the config visualizer with a `tip` box

davidscn:
> I think that, when we release v3, we should point to the v2 PDF in the banner for a while, to make it visible. We could then also add a "Tip" alert box in the mapping config to point to the v2 PDF:

I added now an info box

images/docs/configuration/doc-mapping-options.svg
davidscn: > Why is "iterative solving" and "direct solving" in v3? Since there were GPU updates? Is this clear?

I'm not sure I understand what you mean. Why do I have to give a reason for how we configure? There is a lengthy discussion about the rationale in a precice issue  and it is independent of the GPU. The answer would probably be: because it resolves the configuration complexity of configuring all available options in the best way we could think of.

> Why not have a complete figure, instead?

Putting the GPU executors here is not a good idea. It's more of an advanced topic and I would keep it as such. Considering that the whole docs is only valid for v3, I will remove the version hints and coloring

images/docs/configuration/doc-mapping-rbf-executors.svg
davidscn: The figure shows the executor, which is the same, and not the implementation

pages/docs/configuration/configuration-mapping.md
davidscn: Let's make it more timeless

pages/docs/configuration/configuration-mapping.md
davidscn: I added a hint at the end. People should anyway use the alias. Feel free to edit though.

images/docs/configuration/doc-mapping-options.svg
uekerman: I also meant:

> But then, the (v3.0.0) is not clear to me in the iterative and direct. What is the meaning of the diamond?

davidscn: > What is the meaning of the diamond?

methods are indicated by a circle. The 'how we solve' are not methods on it's own, so I tried to differentiate.

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,63 +60,106 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+  * `rbf-global-direct`, which computes initially a dense matrix decompositions of the linear system and applies the matrix decomposition afterwards in every coupling iteration to solve the mapping problem. All involved linear-algebra data structures are dense, regardless of the basis-function configuration. This method can use differenten [execution backends](configuration-mapping.html#execution-backends). By default, the linear-algebra library Eigen is used. Eigen uses a Cholesky decomposition, if the configured basis-function is strictly positive definite, and a QR decomposition, if that's not the case. This means that using the mapping with one of the basis-functions `volume-splines`, `thin-plate-splines` or `multiquadrics` takes considerable more time to compute, as these basis-functions are not positive definite. While this mapping method can be used while running a participant in parallel via mpi, a gather-scatter approach is used in preCICE, such that the actual system is solved in serial.
```
MakisH: (maybe passing the whole file via [LanguageTool](https://languagetool.org/) would be good)

```suggestion
  * `rbf-global-direct`, which computes initially a dense matrix decompositions of the linear system and applies the matrix decomposition afterwards in every coupling iteration to solve the mapping problem. All involved linear-algebra data structures are dense, regardless of the basis-function configuration. This method can use different [execution backends](configuration-mapping.html#execution-backends). By default, the linear-algebra library Eigen is used. Eigen uses a Cholesky decomposition, if the configured basis-function is strictly positive definite, and a QR decomposition, if that's not the case. This means that using the mapping with one of the basis-functions `volume-splines`, `thin-plate-splines` or `multiquadrics` takes considerable more time to compute, as these basis-functions are not positive definite. While this mapping method can be used while running a participant in parallel via mpi, a gather-scatter approach is used in preCICE, such that the actual system is solved in serial.
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,63 +60,106 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+  * `rbf-global-iterative`, which assembles the system matrix initially and solves the resulting system iteratively in each coupling iteration. This method can use differenten [execution backends](configuration-mapping.html#execution-backends). However, the CPU version relies on the linear-algebra library PETSc, which is an optional dependency of preCICE. The PETSc implementation is fully mpi-parallel and uses sparse linear algebra to solve the resulting system. As a consequence, care has to be taken when configuring the basis-function for the mapping: it should only be used with compactly supported basis-functions and the configured support-radius or shape-parameter should only cover a limited number of vertices in radial direction (e.g. 10 vertices) such that the resulting linear system is sparse. Details about this implementation can be found in [Florian's thesis (pages 37 ff.)](https://elib.uni-stuttgart.de/bitstream/11682/10598/3/Lindner%20-%20Data%20Transfer%20in%20Partitioned%20Multi-Physics%20Simulations.pdf).
```
MakisH: Leonard would not use passive voice when explaining the preCICE mapping to Howard over coffee (see [tone of voice](https://precice.org/docs-meta-overview.html)).

```suggestion
  * `rbf-global-iterative`, which assembles the system matrix initially and solves the resulting system iteratively in each coupling iteration. This method can use differenten [execution backends](configuration-mapping.html#execution-backends). However, the CPU version relies on the linear-algebra library PETSc, which is an optional dependency of preCICE. The PETSc implementation is fully mpi-parallel and uses sparse linear algebra to solve the resulting system. As a consequence, we need to take care when configuring the basis-function for the mapping: it should only be used with compactly supported basis-functions and the configured support-radius or shape-parameter should only cover a limited number of vertices in radial direction (e.g. 10 vertices) such that the resulting linear system is sparse. Details about this implementation can be found in [Florian's thesis (pages 37 ff.)](https://elib.uni-stuttgart.de/bitstream/11682/10598/3/Lindner%20-%20Data%20Transfer%20in%20Partitioned%20Multi-Physics%20Simulations.pdf).
```

pages/docs/configuration/configuration-mapping.md
```diff
@@ -50,63 +60,106 @@ Furthermore, mappings come int two types of `constraint`: `consistent` and `cons
...
+{% endnote %}
 
-```xml
-<mapping:rbf-thin-plate-splines direction="read" from="MyMesh2" to="MyMesh1" constraint="consistent"/>
-```
+* `rbf-pum-direct`, which breaks down the mapping problem in smaller clusters, solves these clusters locally and blends them afterwards together to recover a global solution. This mapping version only needs the linear-algebra library Eigen and the used linear solver is a dense solver in each cluster (actually the same as for `rbf-global-direct`, i.e., it is beneficial to configure strictily positive definite basis-functions). The mapping is specificially designed for large mapping problems and runs fully mpi-parallel. To configure the accuracy of the mapping, the number of `vertices-per-cluster` can be increased. The method can only handle sufficiently matching geometries and problems might occur if large gaps exist between the coupling meshes. Further information, including some performance comparisons, can be found in [this talk of the preCICE workshop 2023](https://youtu.be/df-JMl7UxRg?si=18X3LFTIepmrtAMc). In practical applications, the partition of unity method typically outperforms any of the global rbf variants.
```
MakisH: Since we are already referring to dissertations etc with first names, this makes sense:

```suggestion
* `rbf-pum-direct`, which breaks down the mapping problem in smaller clusters, solves these clusters locally and blends them afterwards together to recover a global solution. This mapping version only needs the linear-algebra library Eigen and the used linear solver is a dense solver in each cluster (actually the same as for `rbf-global-direct`, i.e., it is beneficial to configure strictily positive definite basis-functions). The mapping is specificially designed for large mapping problems and runs fully mpi-parallel. To configure the accuracy of the mapping, the number of `vertices-per-cluster` can be increased. The method can only handle sufficiently matching geometries and problems might occur if large gaps exist between the coupling meshes. Further information, including some performance comparisons, can be found in [David's talk at the preCICE workshop 2023](https://youtu.be/df-JMl7UxRg?si=18X3LFTIepmrtAMc). In practical applications, the partition of unity method typically outperforms any of the global rbf variants.
```

MakisH (approved):
Looks good overall, but I think it is a good time to process the whole page with LanguageTool. No further review iterations are needed from my side, looks clear. :+1:

davidscn:
Thanks for the useful review comments! I will merge here for now, although the image has still an open discussion. We can still improve later on, I think.

PR #310, opened 2023-11-22
Refer to `precice.hpp` instead of `Participant.hpp`

davidscn:
do we want to make the `Participant.hpp` accessible at all (@fsimonis )? Users should use `precice.hpp`


fsimonis:
Accessible yes, but unnecessary. The documentation should only use `precice/precice.hpp`

PR #311, opened 2023-11-23
Explain how to use precice-config-visualizer-gui

carme-hp:
Explain how to use precice-config-visualizer-gui (so far there is no mention of the gui option)


fsimonis (approved):
Looks good!

Can you "squash and merge" this PR?

PR #312, opened 2023-11-27
Fix logger levels

### LINKED ISSUE #297: Wrong documentation of logging config
The documentation on the website https://precice.org/configuration-logging.html says: "Severity, can be trace, debug , info, warn, error". However "warn" is not accepted (leads to a rather cryptic error message), "warning" works.

MakisH:
Closes https://github.com/precice/precice.github.io/issues/297, reported by @dabele.

(tested, `warn` leads to a strange crash)


fsimonis (approved):
True! Thanks for the fix

PR #313, opened 2023-11-29
Include instructions to remove .data() in readData function

carme-hp:
Calling `.data()` is not needed anymore because the data objects are now vectors. The porting guide did not have any instructions on this


pages/docs/couple-your-code/couple-your-code-porting-v2-3.md
```diff
@@ -118,6 +118,7 @@ Please add breaking changes here when merged to the `develop` branch.
   - Directly define face elements or cells of your coupling mesh available in your solver by passing their vectices to preCICE, which automatically handles edges of triangles etc. See [Mesh Connectivity](couple-your-code-defining-mesh-connectivity) for more information.
   - Rename `setMeshTriangleWithEdges` to `setMeshTriangle` and `setMeshQuadWithEdges` to `setMeshQuad`. The edge-based implementation was removed.
   - Use the new bulk functions to reduce sanitization overhead: `setMeshEdges`, `setMeshTriangles`, `setMeshQuads`, `setMeshTetrahedra`.
+- Replace pointer data structures by vectors, e.g., `double* forces = new double[vertexSize*dim]` to `std::vector<double> forces(vertexSize*dim)`.
```
davidscn: In this example, `forces` is a data structure create and handled by the user. The way we used it in our example API was just one version of how such a data structure could look like and how to use it with preCICE. However, both versions (i.e. what you call pointer data structure and vectors) work with the v2 API as well as the v3 API. The C-style array we use in our v2 example is not really a C++ way of doing things and I would expect C++ codes to use vectors anyway. Having this in mind, I am not sure about this hint in the porting guide. We should mention somewhere that the signature (of e.g. readData) changed from a raw pointer to a `span`.

carme-hp: Okay, I agree that using raw pointers is just one specific case.
> We should mention somewhere that the signature (of e.g. readData) changed from a raw pointer to a span.

Stating this should be sufficient, since this already clarifies that raw pointers can no longer be an argument to `readData`.


In any case my motivation for this was a case in v2 of a `std::vector<double> forces` that was given to `readData` as `forces.data()`. So when I ported to v3 to me it was not obvious that I should drop the `data.()`.

uekerman: Maybe we could explain the change to `span` and then give @carme-hp 's situation as an example? Could be relevant for more users.

carme-hp:
I tried to do what @uekerman suggested. Let me know if this is better.

pages/docs/couple-your-code/couple-your-code-porting-v2-3.md
```diff
@@ -60,7 +60,7 @@ Please add breaking changes here when merged to the `develop` branch.
 - preciceDt = interface.initialize();
   
 - if(interface.isActionRequired(cowid)){
--   interface.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
+-   interface.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces.data());
```
fsimonis, line 62: This should be `forces` too.

In the example, we change from `double *forces` to `std::vector<double> forces`. The argument used to be a pointer and has changed to a `span<const double>`, so both are correct.

Yes, it is confusing, but it is correct.

```suggestion
-   interface.writeBlockVectorData(forceID, vertexSize, vertexIDs, forces);
```

--- 

Maybe a more helpful example would be to use `std::vector` in both versions.
Then the old interface `forces.data()` and the new one directly passes `forces`.

pages/docs/couple-your-code/couple-your-code-porting-v2-3.md
```diff
@@ -132,6 +132,7 @@ Please add breaking changes here when merged to the `develop` branch.
 - Replace the commands to read data: `readBlockVectorData`, `readVectorData`, `readBlockScalarData`, `readScalarData` with a single command `readData`.
 - Replace the commands to write data: `writeBlockVectorData`, `writeVectorData`, `writeBlockScalarData`, `writeScalarData` with a single command `writeData`.
 - Replace the commands to write gradient data: `writeBlockVectorGradientData`, `writeVectorGradientData`, `writeBlockScalarGradientData`, `writeScalarGradientData` with a single command `writeGradientData`.
+- The signature of `readData`, `writeData` and `writeGradientData` has changed from `const int*` and `const double*` to `span<const VertexID>` and `span<const double>`. If necessary change the data object, e.g., from `double* forces = new double[vertexSize*dim]` to `std::vector<double> forces(vertexSize*dim)` and remove `.data()` in the function arguments.
```
fsimonis:
```suggestion
- The signature of `readData`, `writeData` and `writeGradientData` has changed from `const int*`, `const double*`, and `double*` to `span<const VertexID>`, `span<const double>`, and `span<double>`. If necessary change the data object, e.g., from `double* forces = new double[vertexSize*dim]` to `std::vector<double> forces(vertexSize*dim)` and remove `.data()` in the function arguments.
```

pages/docs/couple-your-code/couple-your-code-porting-v2-3.md
carme-hp: Thanks, now it actually make sense to me. I do think that using `std::vector` in both versions is a more useful example. That we should not use raw pointers is already stated anyways in one of the bullet points
> The signature of `readData`, `writeData` and `writeGradientData` has changed from `const int*`, `const double*`, and `double*` to `span<const VertexID>`, `span<const double>`, and `span<double>`. If necessary change the data object, e.g., from `double* forces = new double[vertexSize*dim]` to `std::vector<double> forces(vertexSize*dim)` and remove `.data()` in the function arguments.

fsimonis:
@carme-hp I think a major point of confusion is that we don't only change the interfaces from pointers to spans in the examples, but we simultaneously change the types that hold the data from C array to vector.

I attempted to make this a bit clearer in https://github.com/precice/precice.github.io/pull/316. Do you think this is easier to follow?

fsimonis:
I missed the last commit. This makes things way cleaner. I'll merge this and then rebase further overhauls on it.

PR #314, opened 2023-12-01
Fix some outdated solver-interface and dimensions

MakisH:
I hope I have covered everything in the `pages/` directory in terms of:

- Removing `<solver-interface>`
- Moving the `dimensions` from the `solver-interface` to each `mesh`
- Checking that each `mesh` has a `dimensions`

### DIFF
--- a/pages/docs/configuration/configuration-basics.md
+++ b/pages/docs/configuration/configuration-basics.md
@@ -11,13 +11,11 @@ The configuration consists, in general, of the following five parts:
 
 ```xml
 <precice-configuration>
-  <solver-interface dimensions="3">
-   <data .../>
-   <mesh .../>
-   <participant .../>
-   <m2n .../>
-   <coupling-scheme .../>
- </solver-interface>
+  <data .../>
+  <mesh .../>
+  <participant .../>
+  <m2n .../>
+  <coupling-scheme .../>
 </precice-configuration>
 ```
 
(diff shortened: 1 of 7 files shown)


pages/docs/couple-your-code/couple-your-code-gradient-data.md
```diff
@@ -81,20 +81,17 @@ For the example, you can use the following `precice-config.xml` (note the versio
 ```xml
 <?xml version="1.0"?>
 
-<precice-configuration>
-
-  <solver-interface dimensions="3" experimental="on">
-
+<precice-configuration experimental="on">
```
IshaanDesai: Further below in the waveform section, it is `experimental="true"`. Lets be consistent with the value of the tag. Also, is coupling gradient data still experimental?

pages/docs/configuration/configuration-basics.md
```diff
@@ -56,12 +50,14 @@ int temperatureID = precice.getDataID("Temperature", meshID);
...
 
 ```xml
-<mesh name="MyMesh1"> 
+<mesh name="MyMesh1" dimensions="3"> 
   <use-data name="Temperature"/> 
   <use-data name="Forces"/> 
 </mesh> 
 ```
 
+The value `dimensions` needs to match the physical dimension of your simulation, i.e. the number of coordinates a vertex has in `setMeshVertex`, etc. Some solvers only support 3D simulation, such as OpenFOAM or CalculiX. In this case the adapter maps from 3D to 2D if the preCICE dimension is 2D. This, of course, only works if you simulate a quasi-2D scenario with one layer of cells in z direction.
```
IshaanDesai, line 59: Is the Z direction hard-coded as the dead direction?

IshaanDesai (approved):
Looks good, just two minor comments.

pages/docs/configuration/configuration-basics.md
MakisH: This is an example for the OpenFOAM adapter, in which yes, the Z (third index) is hard-coded to be the dead direction.

pages/docs/couple-your-code/couple-your-code-gradient-data.md
MakisH: This PR is only moving things around, but good catch anyway! :D Fixed.

PR #316, opened 2023-12-11
Cleanup porting guide example

fsimonis:
This PR attempts to cleanup the example in the porting guide.

I renamed the old `interface` to `participant`, which reduces some noise.
I also replaced the old pointer stuff by vectors, so the old and new version are now calling the functions using a consistent type.

image


carme-hp (commented):
My understanding is this changes are already merged as part  of https://github.com/precice/precice.github.io/pull/313, so I guess we can close this.

carme-hp:
Looks very good!

PR #323, opened 2024-01-10
Add PKG_CONFIG_PATH setting to libxml2 paths

IshaanDesai:
I encountered this when I had to install libxml2 from source on a cluster. I needed to set the `PKG_CONFIG_PATH` for the CMake recipe to find libxml2.


fsimonis:
Absolutely makes sense. Thanks!

PR #326, opened 2024-01-23
Add comments on pitfalls with very small time step sizes

BenjaminRodenberg:


pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,45 @@ This procedure is independent of whether a serial or a parallel coupling scheme
 For parallel coupling, both solvers run together and everything happens simultaneously in both participants, while for serial coupling, the first participant needs reach the end of the window before the second one can start.
 {% endnote %}
 
+### Possible subcycling pitfall
+
+If you are using very small many time steps in one window, you might see the following error message:
+```
+ERROR: preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows.
+```
```
BenjaminRodenberg: @fsimonis are we actually still raising this error? I lost a bit track of all the back and forth here. Couldn't find the error message anymore when looking for it in the code.

fsimonis: After a long discussion, we didn't merge this error.

https://github.com/precice/precice/pull/1867

fsimonis: But we got a warning:

```
WARN: preCICE just returned a maximum time step size of {}. Such a small value can happen if you use many substeps per time window over multiple time windows due to added-up differences of machine precision
```

fsimonis:
This entire page is still relevant, but needs some overhauling.

MakisH:
@BenjaminRodenberg @fsimonis @uekerman Is this a docs update we still want to merge already?

uekerman:
Without looking into details, I think that we fixed this in preCICE now and don't need this additional documentation anymore.
@BenjaminRodenberg @fsimonis correct?

fsimonis:
@uekerman small timestep sizes are not a problem any more.

What's still a problem is precomputing `dt = part.getMaxTimeStepSize() / nSubsteps` followed by nSubsteps calls of `part.advance(dt)`.

The last step may not necessarily end up at the end of the time window.

MakisH:
And what does this mean regarding this PR?

fsimonis:
I'll do a pass on the content now.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,50 @@ This procedure is independent of whether a serial or a parallel coupling scheme
...
+
+```txt
+preCICE just returned a maximum time step size of <SOME DT>. Such a small value can happen if you use many sub steps per time window over multiple time windows due to added-up differences of machine precision.
+```
+
+The last time step ended up close to the end of the time-window without reaching it, leading to a very small final time step.
+Such small time steps can lead to problems in the solver.
+
+One strategy to avoid this situation is to extend the last time step of a time window preventing problematic time step sizes.
+The follow example extends the time step negotiation between the solver and preCICE to ensure the next time step size `preciceDt - solverDt` stays over some threshold `tol`.
```
fsimonis: Shouldn't we call the `tol` `minDt` to make this easier to understand?

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
```diff
@@ -75,6 +75,50 @@ This procedure is independent of whether a serial or a parallel coupling scheme
...
+// dt = min(preciceDt, solverDt);
+
+// Allow some tolerance
+double tol = 10e-14;
+if (preciceDt - solverDt < tol) {
+  // Use preciceDt, if difference between preciceDt and solverDt is small enough
+  dt = preciceDt;
+} else {
+  dt = min(preciceDt, solverDt);
+}
```
fsimonis: The condition already covers the case, where `preciceDt < solverDt`, so this could be simplified to the following:

```suggestion
if (preciceDt - solverDt < tol) {
  dt = preciceDt;
} else {
  dt = solverDt;
}
```

fsimonis (commented):
@uekerman @BenjaminRodenberg I reworked the section to explain the remaining issue.

We could improve readability in some sections though.

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
uekerman: Both work for me

uekerman (approved):

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
BenjaminRodenberg: `minDt` sounds good :+1:

pages/docs/couple-your-code/couple-your-code-time-step-sizes.md
BenjaminRodenberg: Maybe put a comment above that this block replaced the `dt = min(preciceDt, solverDt);` known from other sections? Otherwise it's maybe a bit hard to directly see where the `dt = min(preciceDt, solverDt);` ended up. But I'm also fine with this simplification.

BenjaminRodenberg (commented):
Thanks for taking care of the final refinement loop! Looks good to me.

fsimonis:
I'll merge this PR

PR #337, opened 2024-02-19
Add pre-commit workflow

fsimonis:
This PR replaced the check markdown workflow with a workflow which runs pre-commit.

This ensures that the linting here is consistent with what the user should run into when committing.


MakisH:
We should also mention in the `README.md` how to get and use the pre-commit hook. Linking to the related note for preCICE is fine.

PR #338, opened 2024-02-29
Fix incorrect path

BenjaminRodenberg:


davidscn (approved):
The command is not supposed to be general, more an example, but `..` is probably more common.

PR #347, opened 2024-03-11
Change runner branch to main

uekerman:
Do I also need to update the actual submodules manually or is this automatic?


uekerman:
see also https://github.com/precice/fmi-runner/commit/b8b6e66cdca78c2336136ae7c3bc816e46f99602

MakisH (approved):
I am not sure if anything else is needed. Let's just try this, and the website build will complain if that's not enough.

PR #350, opened 2024-03-12
Add dedicated page on course

uekerman:
Main motivation is to improve the visibility of the course. We could link to this page from several places on the website (still todo), including the support program.

@MakisH What do you think?


MakisH:
It would help a lot if we also had the requirements there (always the latest), as we could then always simply link to this page for the description/preparation instructions of every workshop/conference/event we offer this course to.

uekerman:
Requirements you simply mean the latest vm? Or do want to make all components explicit?

MakisH:
I just want to list here everything that I would normally need to copy-paste and adapt for each event:

1. Get the VM from that page
2. Or install these dependencies on your laptop
3. Or we might have some Live USB sticks

I can also directly edit.

pages/community/community-training.md
```diff
@@ -0,0 +1,35 @@
+---
+title: preCICE training
```
MakisH, line 2: I renamed the page to `preCICE training`, as I think this is easier to recognize.

The URL now reads `community-training.html`, which is also fine, since we train the community.

The sidebar URL was broken before.

MakisH (commented):
Added a "How to prepare?" section.
I think we could also add a picture for each module, even though we cannot really directly reuse what we have.

MakisH (approved):
Added also some images. Looks good to me. :+1:

uekerman:
:+1:
I still want to add more links to this page. Will do the next days and merge then.

PR #355, opened 2024-03-22
Multi coupling doc

carme-hp:
Add more detailed documentation about control participant.


uekerman (changes requested):
That's not 100% correct, I'll have a look next week. Nothing we need for the distribution.

uekerman:
The restriction that we need to document is actually already covered in the versions box on the same page. This addition is redundant.

PR #361, opened 2024-03-26
Add docs on DaRUS automation

uekerman:
As we have no good place yet where to upload the scripts and the scripts need manual editing anyway, I thought adding them in plain text here could be a good compromise.

The actual component versions still need final updates.

@MakisH No need to test anything. A quick check should be enough. I will anyway "test this" with the real data.

Thanks @JR-1991 for the help setting this up :pray: Should be added as a co-author here when merging.

### DIFF
--- a/pages/docs/dev-docs/dev-publications.md
+++ b/pages/docs/dev-docs/dev-publications.md
@@ -170,4 +170,4 @@ Order of authors: maintainers are listed first followed by everybody else, both 
 
 6. Review and publish
 
-    The upload returns a url, which gives you access to the dataset. Review carefully. Currently, for example, licenses still need manual editing. One ready, publish (i.e. send to DaRUS team for review).
+    The upload returns a url, which gives you access to the dataset. Review carefully. Currently, for example, licenses still need manual editing. Once ready, publish (i.e. send to DaRUS team for review).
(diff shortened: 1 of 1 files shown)


MakisH (commented):
Basic sanity check makes sense. I will approve when the final versions are there.

MakisH (approved):
Actually, this is ready, as we don't need to update the versions here. I created a separate PR for that: https://github.com/precice/precice.github.io/pull/380

uekerman:
@MakisH In #380 will you then also update the versions on this page? (currently you don't)

MakisH:
No: the versions will then always need to be updated, and they will be duplicated. I find TODOs better for this template.

uekerman:
we could also leave the versions of the current distribution and document that this is an example and needs to be updated before using.
if not, then we should probably switch to TODOs here everywhere

uekerman:
I now added the versions and a bit of documentation. Simply saves me one TODO when I need to actually apply the script.

pages/docs/dev-docs/dev-publications.md
```diff
@@ -17,3 +17,157 @@ Authors of the data set are:
...
+    )
+
+    # delete pid to upload to new dataset, otherwise existing one overwritten
+    dataset.p_id = None
+    # upload to dataverse of US3
+    dataset.upload("ipvs_us3")
+
+6. Review and publish
+
+    The upload returns a url, which gives you access to the dataset. Review carefully. Currently, for example, licenses still need manual editing. One ready, publish (i.e. send to DaRUS team for review).
```
uekerman:
```suggestion
    The upload returns a url, which gives you access to the dataset. Review carefully. Currently, for example, licenses still need manual editing. Once ready, publish (i.e. send to DaRUS team for review).
```

PR #362, opened 2024-03-26
Remove unsupported CLI usage from tooling-config-visualization.md

MakisH:
Related to https://github.com/precice/config-visualizer/issues/15.

@fsimonis I am fine if you then consider that issue as closed as well.


fsimonis (approved):
Ok. Let's remove this use-case.

PR #364, opened 2024-03-27
Add pages for adapters of DuMuX and DUNE

IshaanDesai:
The content for the DuMuX and DUNE adapters is directly copied from their respective repository READMEs. Further cleaning and editing can be done in the future.


_data/sidebars/docs_sidebar.yml
MakisH, line 1: I wanted DUNE to appear before DuMuX, but DuMuX has multiple pages (it is a folder), while DUNE does not (it is not a folder). To be able to rearrange the sidebar, I made everything folders, as more entries would likely come. In a follow-up, I would probably sort these alphabetically already.

MakisH (approved):
Thanks for contributing!

I ended up almost rewriting this PR, by moving the new pages directly to their repositories. I should have probably just closed this and opened a new one, but why not... :sweat_smile: 

Related PRs in the adapters:

- DUNE: https://github.com/precice/dune-adapter/pull/9
- DuMuX: https://github.com/precice/dumux-adapter

PR #366, opened 2024-04-02
Rewrite coupling-data section

fsimonis:
This PR attempts to clarify the coupling-data section of the docs

### DIFF
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -35,7 +35,7 @@ The acceleration modifies coupling data in `advance()`, meaning, written values 
 
 Different acceleration schemes compute these coefficients in different ways.
 They are generally based on the values of previous iterations.
-A notable exception is the constant under-relaxation, which used fixed coefficients.
+A notable exception is the constant under-relaxation, which uses fixed coefficients.
 
 Such value-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
 We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -27,7 +27,7 @@ All data communicated within a coupling scheme needs to be configured through `e
 </coupling-scheme:serial-implicit>
 ```
 
-The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by the another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
+The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
 
 * For a **parallel coupling**, all coupling data is accelerated the same way. This means all coupling data use the same coefficients for the linear combination.
 * For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -30,7 +30,7 @@ All data communicated within a coupling scheme needs to be configured through `e
 The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
 
 * For a **parallel coupling**, all coupling data is accelerated the same way. This means all coupling data use the same coefficients for the linear combination.
-* For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
+* For a **serial coupling**, only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
 
 **Example:** For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are accelerated in case of serial coupling. For parallel coupling, both displacements and forces are accelerated.
 
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -33,12 +33,11 @@ The acceleration modifies coupling data in `advance()`, meaning, written values 
 
 **Example:** For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are accelerated in case of serial coupling. For parallel coupling, both displacements and forces are accelerated.
 
-Different acceleration schemes compute these coefficients in different ways, which can be grouped into two categories:
+Different acceleration schemes compute these coefficients in different ways.
+They are generally based on the values of previous iterations.
+A notable exception is the constant under-relaxation, which used fixed coefficients.
 
-* Coefficients are fixed for each iteration, thus independent of the accelerated values. Example being constant under-relaxation.
-* Coefficients depend on the values of previous iterations. Example being Interface Quasi-Newton methods.
-
-Data-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
+Such value-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
 We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
 In the code example above, `Displacements` is primary data and `Forces` is secondary data.
 Which data may be configured depends on the coupling scheme:
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -26,17 +26,27 @@ All data communicated within a coupling scheme needs to be configured through `e
 </coupling-scheme:serial-implicit>
 ```
 
-The acceleration modifies coupling data in `advance()`. This means, what you write to preCICE on the one participant is not the same with what you read on the other participant. The data values are stabilized (or "accelerated") instead. This happens also by using values from previous iterations. Simply think of a linear combination of previous iterations.
+The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by the another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
 
-* For a **parallel coupling**, all coupling data is post-processed the same way. This means all coupling data use the same coefficients for the linear combination.
-* For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is post-processed. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
+* For a **parallel coupling**, all coupling data is accelerated the same way. This means all coupling data use the same coefficients for the linear combination.
+* For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
 
-Let's look at an example: For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are post-processed in case of serial coupling. For parallel coupling, both displacements and forces are post-processed.
+{% example %}
+For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are accelerated in case of serial coupling. For parallel coupling, both displacements and forces are accelerated.
+{% endexample %}
 
-Next, we have to configure based on which data the acceleration computes, i.e. how the coefficients in the linear combinations get computed. These data fields are defined within the acceleration as `data` tags (such as `Displacements` in the code example above). Let's call these data fields **primary data**. (Just for completeness: All coupling data that gets post-processed and that is not primary data, is called "secondary data".)
+Different acceleration schemes compute these coefficients in different ways, which can be grouped into two categories:
 
-* For **serial coupling**, you can only configure one primary data field, which should correspond to a coupling data field that is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
-* For **parallel coupling**, an arbitrary number of primary data can be configured. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
+* Coefficients are fixed for each iteration, thus independent of the accelerated values. Example being constant under-relaxation.
+* Coefficients depend on the values of previous iterations. Example being Interface Quasi-Newton methods.
+
+Data-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
+We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
+In the code example above, `Displacements` is primary data and `Forces` is secondary data.
+Which data may be configured depends on the coupling scheme:
+
+* For **serial coupling**, you can only configure primary data from coupling data which is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
+* For **parallel coupling**, all coupling data is available as primary data. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
 
 Now, we know the difference between coupling data and primary data. Next, we have a look on how we actually configure the type of acceleration.
 
--- a/pages/docs/configuration/configuration-acceleration.md
+++ b/pages/docs/configuration/configuration-acceleration.md
@@ -48,6 +48,8 @@ Which data may be configured depends on the coupling scheme:
 
 In the code example above, `Displacements` is primary data and `Velocities` is secondary data. `Forces` is not accelerated as the case uses a serial coupling scheme.
 Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
+As mentioned above, this is a theoretical setup as one primary data per direction is highly encouraged.
+Hence, `Forces` should be marked primary as well.
 
 Now, we know the difference between coupling data and primary data. Next, we have a look on how we actually configure the type of acceleration.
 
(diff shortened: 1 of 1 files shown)


pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +26,25 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+Different acceleration schemes compute these coefficients in different ways, which can be grouped into two categories:
 
-* For **serial coupling**, you can only configure one primary data field, which should correspond to a coupling data field that is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
-* For **parallel coupling**, an arbitrary number of primary data can be configured. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
+* Coefficients are fixed for each iteration, thus independent of the accelerated values. Example being constant under-relaxation.
+* Coefficients depend on the values of previous iterations. Example being Interface Quasi-Newton methods.
+
+Data-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
```
MakisH: I assume that by "data-dependent" you mean the second category. Making the connection explicit would help.

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +26,25 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
```
MakisH, line 42: This is now way clearer than the previous version, which I was reading for a while to understand what it means. But still: what does mean "accelerated without influencing the coefficients" from the configuration perspective?

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +26,25 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+In the code example above, `Displacements` is primary data and `Forces` is secondary data.
```
MakisH: The "why" would be important here.
How can the user affect what will be primary and which secondary?

MakisH (commented):
This is now very clearer! A few suggestions/questions from the user perspective.

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +26,25 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+Different acceleration schemes compute these coefficients in different ways, which can be grouped into two categories:
 
-* For **serial coupling**, you can only configure one primary data field, which should correspond to a coupling data field that is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
-* For **parallel coupling**, an arbitrary number of primary data can be configured. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
+* Coefficients are fixed for each iteration, thus independent of the accelerated values. Example being constant under-relaxation.
+* Coefficients depend on the values of previous iterations. Example being Interface Quasi-Newton methods.
```
uekerman: Two categories sounds a bit strong here IMO. I could imagine that this confuses here more than it helps. Maybe, it is easier to consider constant under-relaxation as an exception and explain first.

pages/docs/configuration/configuration-acceleration.md
uekerman: This is only true for serial coupling. The example above considers both serial and parallel coupling.

uekerman (changes requested):
Things are getting clearer already, but some details need tweaking.

pages/docs/configuration/configuration-acceleration.md
fsimonis: The code example is in fact `serial-implicit`.

```suggestion
In the serial code example above, `Displacements` is primary data and `Forces` is secondary data.
```

This still confuses me though. The second participant receives Forces, does its step and then accelerates the data. Why would it accelerate `Forces`?

pages/docs/configuration/configuration-acceleration.md
fsimonis: > Such value-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
The former is explicitly listed in the `acceleration` using `data` tags, while the latter is implied by the primary data and the configured coupling scheme.

Is that clearer?

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -33,13 +33,13 @@ The acceleration modifies coupling data in `advance()`, meaning, written values
 
 **Example:** For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are accelerated in case of serial coupling. For parallel coupling, both displacements and forces are accelerated.
 
-Different acceleration schemes compute these coefficients in different ways, which can be grouped into two categories:
+Different acceleration schemes compute these coefficients in different ways.
+They are generally based on the values of previous iterations.
+A notable exception is the constant under-relaxation, which used fixed coefficients.
```
MakisH:
```suggestion
A notable exception is the constant under-relaxation, which uses fixed coefficients.
```

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -33,13 +33,13 @@ The acceleration modifies coupling data in `advance()`, meaning, written values
...
 
-* Coefficients are fixed for each iteration, thus independent of the accelerated values. Example being constant under-relaxation.
-* Coefficients depend on the values of previous iterations. Example being Interface Quasi-Newton methods.
-
-Data-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
+Such value-dependent acceleration schemes need to select which data to compute these coefficients from by listing them as `data` tags inside the `acceleration` tag.
 We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
+The former is explicitly listed in the `acceleration` using `data` tags, while the latter is implied by the primary data and the configured coupling scheme.
```
MakisH, line 43: What does _"the latter is implied by the primary data and the configured coupling scheme."_ mean?

pages/docs/configuration/configuration-acceleration.md
uekerman: Data sent from first to second in a serial scheme is never accelerated. Thus, it is neither primary nor secondary, but not related to the acceleration at all. That's still the confusing part, right?
(so the current statement is actually still wrong)

fsimonis: Understood :+1: 

```suggestion
In the code example above, `Displacements` is primary data and there is no secondary data.
```

pages/docs/configuration/configuration-acceleration.md
fsimonis: This should be pretty clear.

The secondary data is implicit. It contains all accelerated data according to the coupling scheme (see the list above the example) minus the data configured as primary data.

pages/docs/configuration/configuration-acceleration.md
uekerman:
```suggestion
In the code example above, for serial coupling, `Displacements` is primary data there is no secondary data. For parallel coupling, on the other hand, both data fields are primary data.
```

Maybe it makes more sense to use a new example here (simply give configuration of acceleration) with primary and secondary data?

pages/docs/configuration/configuration-acceleration.md
fsimonis: Obsolete ass the categories are no more

pages/docs/configuration/configuration-acceleration.md
fsimonis: I had a look at our tutorials, and we don't seem to have one with a serial coupling scheme and secondary data.

I would propose to add another data exchanged from second to first in our example above. This way, we keep one example for the whole section.

fsimonis: I added `Density` as some material property to the code example and moved the explanation to the bottom. IMHO the result flows better.

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -17,7 +17,8 @@ All data communicated within a coupling scheme needs to be configured through `e
 <coupling-scheme:serial-implicit>
   <participants first="FluidSolver" second="StuctureSolver"/>
   <exchange data="Displacements" mesh="StructureMesh" from="StuctureSolver" to="FluidSolver"/>  
-  <exchange data="Forces" mesh="StructureMesh" from="FluidSolver" to="StuctureSolver"/>        
+  <exchange data="Density"       mesh="StructureMesh" from="StuctureSolver" to="FluidSolver"/>
```
uekerman:
```suggestion
  <exchange data="Velocities" mesh="StructureMesh" from="StuctureSolver" to="FluidSolver"/>  
```
is a more realistic case IMO

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +27,27 @@ All data communicated within a coupling scheme needs to be configured through `e
 </coupling-scheme:serial-implicit>
 ```
 
-The acceleration modifies coupling data in `advance()`. This means, what you write to preCICE on the one participant is not the same with what you read on the other participant. The data values are stabilized (or "accelerated") instead. This happens also by using values from previous iterations. Simply think of a linear combination of previous iterations.
+The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by the another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
```
uekerman:
```suggestion
The acceleration modifies coupling data in `advance()`, meaning, written values by the one participant are not the same when read by another participant. The values are stabilized (or "accelerated") based on a linear combination of previous iterations.
```

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +27,27 @@ All data communicated within a coupling scheme needs to be configured through `e
...
 
-* For a **parallel coupling**, all coupling data is post-processed the same way. This means all coupling data use the same coefficients for the linear combination.
-* For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is post-processed. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
+* For a **parallel coupling**, all coupling data is accelerated the same way. This means all coupling data use the same coefficients for the linear combination.
+* For a **serial coupling** only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
```
uekerman:
```suggestion
* For a **serial coupling**, only coupling that is exchanged from the `second` to the `first` participant is accelerated. Coupling data exchanged in the other direction (from `first` to `second`) is not modified.
```

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +27,27 @@ All data communicated within a coupling scheme needs to be configured through `e
...
 
-Let's look at an example: For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are post-processed in case of serial coupling. For parallel coupling, both displacements and forces are post-processed.
+**Example:** For fluid-structure interaction, if we first execute the fluid solver with given interface displacements followed by the structure solver taking forces from the fluid solver and computing new interface displacements, (only) the displacements are accelerated in case of serial coupling. For parallel coupling, both displacements and forces are accelerated.
```
uekerman: Is this the example from above? Then, we should also mention the velocities values or say that we ignore them for the moment.

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +27,27 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
+The former is explicitly listed in the `acceleration` using `data` tags, while the latter is implied by the primary data and the configured coupling scheme.
+Which data may be configured depends on the coupling scheme:
+
+* For **serial coupling**, you can only configure primary data from coupling data which is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
+* For **parallel coupling**, all coupling data is available as primary data. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
+
+In the code example above, `Displacements` is primary data and `Density` is secondary data. `Forces` is not accelerated as the case uses a serial coupling scheme.
+Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
```
uekerman, line 50: But this last case is discouraged.

uekerman (changes requested):
We still have some issues here with the examples. When do we use which example? It might be easier to use multiple ones and make things explicit.

pages/docs/configuration/configuration-acceleration.md
fsimonis: This example was already there. It fits the narrative, but there was no direct link to the code.

pages/docs/configuration/configuration-acceleration.md
fsimonis: Do we need to specify that though?

pages/docs/configuration/configuration-acceleration.md
fsimonis: Moving to other threads

pages/docs/configuration/configuration-acceleration.md
uekerman: I would say yes. Further above we say "you should not do this" and then we do it ourselves?
Maybe a "theoretically, ..." + "discouraged" could be a fix here.

fsimonis:
```suggestion
Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
As mentioned above, this is a theoretical setup as one primary data per direction is highly encouraged.
Hence, `Forces` should be marked primary as well.
```

uekerman (approved):

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -17,7 +17,8 @@ All data communicated within a coupling scheme needs to be configured through `e
 <coupling-scheme:serial-implicit>
```
MakisH, line 17: From this example, I understand that the `<data>` inside the `<acceleration>` does not mean "which data is accelerated", but "which data contributes to the coefficients".

This is not obvious, and we should clarify it somewhere.

(indeed, this is what our config reference also [says](https://precice.org/configuration-xml-reference.html#data-1))

pages/docs/configuration/configuration-acceleration.md
```diff
@@ -26,17 +27,29 @@ All data communicated within a coupling scheme needs to be configured through `e
...
+We call data which influences the coefficients **primary data** and data which is accelerated without influencing the coefficients **secondary data**.
+The former is explicitly listed in the `acceleration` using `data` tags, while the latter is implied by the primary data and the configured coupling scheme.
+Which data may be configured depends on the coupling scheme:
+
+* For **serial coupling**, you can only configure primary data from coupling data which is exchanged from the `second` to the `first` participant. In the FSI example, the `Displacements`.
+* For **parallel coupling**, all coupling data is available as primary data. For numerical performance reasons, you should define at least one coupling data field of each direction (one from `second` to `first`, one from `first` to `second`). In the FSI example, configure `Displacements` and `Forces`.
+
+In the code example above, `Displacements` is primary data and `Velocities` is secondary data. `Forces` is not accelerated as the case uses a serial coupling scheme.
+Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
```
MakisH, line 50: I feel that such an overview would help get a quick idea:

```suggestion
In the code example above, `Displacements` is primary data and `Velocities` is secondary data. `Forces` is not accelerated as the case uses a serial coupling scheme:

| Category            | Data fields |
| ---                 | ---         |
| Exchanged           | `Displacements`, `Velocities`, `Forces` |
| Accelerated         | `Displacements`, `Velocities` |
| Define coefficients | `Displacements` |
| Not accelerated     | `Forces` |

Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
```

I think that the terminology "primary/secondary" is not so important / does not help here, but it could also go into a third, "comments" column.

pages/docs/configuration/configuration-acceleration.md
fsimonis: The config reference could be a bit more verbose though.

"compute the acceleration" can mean many things.

pages/docs/configuration/configuration-acceleration.md
fsimonis: Feels like overkill.

If we add such a table, then I would suggest doing it as a summary adding columns for serial and parallel cases.


| Category            | Serial Coupling | Parallel Coupling |
| ---                 | ---         | --- |
| Exchanged           | `Displacements`, `Velocities`, `Forces` | `Displacements`, `Velocities`, `Forces` |
| Accelerated         | `Displacements`, `Velocities` | `Displacements`, `Velocities`, `Forces` |
| Define coefficients | `Displacements` | `Displacements`, `Forces` |
| Not accelerated     | `Forces` | - |

MakisH: Such a table would help me, but if you think this is an overkill, I guess we are fine if people don't understand as easily as we could enable them to. :upside_down_face: 

I like the distinction with the serial and parallel coupling.

fsimonis: If it helps you, then its reason enough to add it. @uekerman any thoughts / improvements?

The categories are inconsistent though. We talk about primary and secondary data, and then use "define coefficients" and "accelerated".

MakisH: As I wrote:

> I think that the terminology "primary/secondary" is not so important / does not help here, but it could also go into a third, "comments" column.

MakisH:
```suggestion
In the code example above, `Displacements` is primary data and `Velocities` is secondary data. `Forces` is not accelerated as the case uses a serial coupling scheme:

| Category | Serial Coupling | Parallel Coupling |
| ---  | ---  | --- |
| Exchanged | `Displacements`, `Velocities`, `Forces` | `Displacements`, `Velocities`, `Forces` |
| Accelerated | `Displacements`, `Velocities` | `Displacements`, `Velocities`, `Forces` |
| Define coefficients ("primary data") | `Displacements` | `Displacements`, `Forces` |
| Exchanged but not accelerated ("secondary data") | `Forces` | - |

Changing the `serial-implicit` to a `parallel-implicit` scheme would turn `Forces` into secondary data.
```

fsimonis: But we introduce these terms for the sole reason to explain what is happening in preCICE. Using the suggested categories does question the value of the entire section.

That said, this may as well be a symptom, that "primary/secondary" is a useless classification. Maybe explaining what is accelerated and what is used to compute the coefficients is way easier to follow without attempting to tie them to "primary/secondary".

MakisH: I don't challenge the terms: I understand there are good (literature) reasons that they are named like this.

I am just trying to help new readers get quicker to the point and understand what is going on and why. In my latest suggestion, I have added both the description and the names in the first column.

uekerman: I also think that such a table is overkill. I see all this information here merely as docs for devs, especially primary vs. secondary. For users not interested in the math, we could also just write "trust that preCICE does the right thing".

MakisH: ok, go ahead.

MakisH (approved):

PR #368, opened 2024-04-03
Introduce folders in the tutorials sidebar and sort tutorials

MakisH:
This is motivated by our sidebar not being scrollable, neither in Firefox, nor in Chromium.
Since we now have way too many tutorials, this is a bit chaotic.

This PR:

- Introduces two folders: `Basic cases` and `Further cases` in the sidebar, similarly to our current categorization in the overview page.
- Makes the "Introduction" a folder as well, which now includes two pages: "Overview" and "Visualization", which moves some useful information from the bottom of the previous "Introduction" page. The URL remains the same for the Introduction page (not for the visualization subsection, which should not have too many references yet). This now also looks better and clearer with two pages instead of one.
- Adds a title above the sidebar.
- The title includes a version: This is the version of the preCICE distribution. @uekerman: Should we change this to anything else?

Partially related to #345.

Screenshot from 2024-04-03 19-58-58

Screenshot from 2024-04-03 19-59-03

Screenshot from 2024-04-03 19-59-12

PR #370, opened 2024-04-04
Add flow around controlled moving cyl to sidebar etc

uekerman:
We forgot to do this.

https://github.com/precice/tutorials/pull/474


MakisH (approved):
Looks good. Did not check if this renders properly, but I will merge into #368 and check there.

Issue #345, opened 2024-03-07
Categorize the tutorials


fsimonis:
Currently, [the tutorials](https://precice.org/tutorials.html) are unstructured on the website.

It's a topnav item that leads to a sidebar containing all tutorials. This looks a bit odd and is also inconsistent with `Docs` and `Community`:

![image](https://github.com/precice/precice.github.io/assets/13552216/8733f92e-68a2-4bda-89f5-2e7d3650415e) ![image](https://github.com/precice/precice.github.io/assets/13552216/3434e212-14ee-489b-b98e-30c20a92cf9b) ![image](https://github.com/precice/precice.github.io/assets/13552216/c2a62efb-a0b3-4e00-8bfd-a6fb557c2fe4)

Docs and community have titles above the sidebar, while Tutorials doesn't.
Also, there is only one pretty long folder containing all tutorials.

It may be an idea to categorize the tutorials into subfolders to split this list into logical parts and make the growing list of tutorials easier to digest.
Example:

* Introduction
* FSI
* CHT
* Fluid-Fluid
* ...

MakisH:
We definitely need to structure the rapidly growing collection, in some way. Deal.II has a very nice overview: https://dealii.org/current/doxygen/deal.II/Tutorial.html

I think categorization (putting things in folders) is the wrong way, as several of our tutorials belong to multiple categories.

What we can easily do, instead, is tagging each tutorial with different tags, and render them in different categories: "these tutorials use OpenFOAM", "these tutorials do an FSI application", etc.

For this, we just need a jekyll collection with a field for tags. I already did something similar in the [RSE learning and teaching resources](https://de-rse.org/learn-and-teach/competencies/) and we have something similar for our publications here. We could have a more dynamic approach and nicer visualization, though.

This is definitely something for PreECO.

> Docs and community have titles above the sidebar, while Tutorials doesn't.

I guess this is easy to add, no?

MakisH:
As @uekerman pointed out in #368, we still have the same issue with the sidebar in the "further cases". We should either fix that (I cannot find any quick fix), or at least add categories in the sidebar. Contradicting what I wrote above, maybe introducing some rough categories now would be 90% of a solution, and we could duplicate if we think that something should belong to multiple categories.

I would go with:

- Basic cases
- Conjugate heat transfer
- Fluid-structure interaction
- Partitioned heat conduction
- Partitioned flows
- Volume coupling
- Further cases
   - ASTE turbine
   - Oscillator
[...]

PR #372, opened 2024-04-09
Arrange tutorials sidebar and overview in collections

MakisH:
Second/follow-up attempt to #345. @uekerman how does this feel?

Main changes:

- Grouped together similar/derived tutorials into the "Channel transport", "Flow over heated plate", "Partitioned flow", and "Partitioned heat conduction" collections in the sidebar (folders) and in the README (second-level items).
- Renamed ~~"Basic cases" -> "Featured" and~~ "Further cases" -> "All tutorials" (and the list now contains all, which I think is easier to parse)

Note that now clicking on, e.g., "Partitioned heat conduction" will highlight it in both the "Featured" and the "All tutorials" parts of the sidebar. I guess this is something we can live with.

tutorials-sidebar

Screenshot 2024-04-09 at 15-22-37 A handful of exciting tutorials preCICE - The Coupling Library

### DIFF
--- a/_data/sidebars/tutorial_sidebar.yml
+++ b/_data/sidebars/tutorial_sidebar.yml
@@ -34,7 +34,7 @@ entries:
       url: /tutorials-perpendicular-flap.html
       output: web
 
-  - title: Further cases
+  - title: All tutorials
     output: web
     folderitems:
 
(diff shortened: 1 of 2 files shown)


uekerman:
The duplication feels a bit odd, but not a big issue. I am not sure about the wording "featured". I still think that sth like "basic" describes better what these cases are.

MakisH:
> The duplication feels a bit odd, but not a big issue. I am not sure about the wording "featured". I still think that sth like "basic" describes better what these cases are.

Reverted

uekerman (approved):
I only looked at the screenshots, not at the diff.

PR #382, opened 2024-04-18
Replace global data documentation with workaround

fsimonis:
This PR replaces the documentation of the global data API with a workaround until the feature is actually in preCICE.


uekerman:
I see this documentation mainly for us. Couldn't we, instead of deleting, point to the branch?

fsimonis:
This is bound to confuse people though.
That's why I opened https://github.com/precice/precice.github.io/pull/383 to revert once global data is actually merged.

PR #387, opened 2024-04-24
Update Roadmap

MakisH:
At the moment, the Roadmap looks a bit outdated, since many features shipped in v3.

As a first step, I have removed some of these features, keeping those that I understand we are still actively working on. @uekerman anything else to remove or now already add?

Does not need to be thorough, but let's not have something that looks immediately outdated.

- [x] Remove completed entries from Roadmap
- [ ] Add new points


MakisH:
Something for the next coding days.

PR #401, opened 2024-05-03
Document running participants in different systems

MakisH:
@LouieVoit published a [very interesting paper](https://doi.org/10.1007/978-3-031-40843-4_42), using preCICE to couple a CFD solver running on a laptop (_"while waiting at the airport and connected to a poor wi-fi"_) to a massively-parallel MD simulation running on a cluster.

The paper is helpful to demonstrate the concept, but lacks some technical details. I got these from Louis via e-mail and I am summarizing them in this documentation addition at the bottom of the [communication configuration](https://precice.org/configuration-communication.html) page.

Note that this is not the first user that has asked for something like this. In the context of the helicopter simulations Master's thesis I supervised, we had a very similar issue, where each code was licensed for a different system.

- @LouieVoit Have I forgotten any details?
- @fsimonis Does this make sense, especially regarding `gather-scatter`?
- @davidscn Do you maybe want to test this? Could be very easy for you, since you daily work on a cluster.

### DIFF
--- a/pages/docs/running/running-distributed.md
+++ b/pages/docs/running/running-distributed.md
@@ -51,6 +51,7 @@ One advantage of communicating via TCP/IP sockets is that participants can be ex
 
 - Coupling participants with licenses restricting them to a specific system
 - Coupling participants running on a cluster with GUI-based participants running locally, while exchanging only interface data
+- Coupling participants running on containers or virtual machines (see [related discussion](https://precice.discourse.group/t/how-to-configure-precice-communication-on-kubernetes-with-tcp-port-access-control/1451))
 
 You can achieve this by configuring an `m2n:sockets` communication interface with an explicitly-defined port, which you can then forward via your SSH connection. Since you only open one port, you need to select `enforce-gather-scatter`, so that only one rank of each participant communicates. For example:
 
--- a/pages/docs/running/running-distributed.md
+++ b/pages/docs/running/running-distributed.md
@@ -44,3 +44,30 @@ If you have logged into remote machine via SSH, then be careful to use a tool su
 This allows to safely detach from a session without the running process getting killed.
 
 If your cluster is managed using SLURM, then [further steps may be necessary](running-slurm).
+
+## Running participants in different systems
+
+One advantage of communicating via TCP/IP sockets is that participants can be executed in different systems, even connected via the internet. This feature [has been demonstrated in the literature](https://doi.org/10.1007/978-3-031-40843-4_42) and has a few use cases, such as:
+
+- Coupling participants with licenses restricting them to a specific system
+- Coupling participants running on a cluster with GUI-based participants running locally, while exchanging only interface data
+
+You can achieve this by configuring an `m2n:sockets` communication interface with an explicitly-defined port, which you can then forward via your SSH connection. Since you only open one port, you need to select `enforce-gather-scatter`, so that only one rank of each participant communicates. For example:
+
+```xml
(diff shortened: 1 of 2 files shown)


fsimonis:
I think most of this information fits better in the [Distributed Systems page](https://precice.org/running-distributed.html).

Not so sure about where the licensing would fit though. Maybe this is a use-case that deserves far more attention and could even be a major selling point of preCICE.

A hint on using `gather-scatter` if you need to deal with a single port between solvers would be useful on the configuration page though.

MakisH:
> I think most of this information fits better in the [Distributed Systems page](https://precice.org/running-distributed.html).

> A hint on using `gather-scatter` if you need to deal with a single port between solvers would be useful on the configuration page though.

Good point, I forgot we now have this. I can move it there, and duplicate the ports thingy here.

> Not so sure about where the licensing would fit though. Maybe this is a use-case that deserves far more attention and could even be a major selling point of preCICE.

I deserves some attention, but the use-case is rather niche. I don't see any obvious solution, as we don't seem to mention such advantages anywhere.

What I can do, however, is create a page `License`, explaining what LGPL means, what licenses we follow for adapters, how everything else (e.g., the website) is licensed, and what other advantages our approach brings (e.g., this).

@uekerman what do you think?

fsimonis:
> I deserves some attention, but the use-case is rather niche. I don't see any obvious solution, as we don't seem to mention such advantages anywhere.
>
> What I can do, however, is create a page License, explaining what LGPL means, what licenses we follow for adapters, how everything else (e.g., the website) is licensed, and what other advantages our approach brings (e.g., this).

To clarify: the feature I am talking about is the ability to couple commercial code whose licences are bound to machines or cannot be used/fetched outside the company network.

MakisH:
I moved the section to the distributed systems page and left a note for the specific port in the communication. I will work on the license-related page in a separate PR.

> To clarify: the feature I am talking about is the ability to couple commercial code whose licences are bound to machines or cannot be used/fetched outside the company network.

This is clear enough and also what I already wrote in the documentation. I will elaborate on these in the license page PR.

LouieVoit:
@MakisH Looks good to me !

MakisH:
> I will work on the license-related page in a separate PR.

Done: #407.

MakisH:
Rendering, to help with reviewing (@fsimonis @davidscn):

Screenshot from 2024-05-13 14-40-48

Screenshot 2024-05-13 at 14-43-38 Distributed systems preCICE - The Coupling Library

pages/docs/running/running-distributed.md
```diff
@@ -44,3 +44,30 @@ If you have logged into remote machine via SSH, then be careful to use a tool su
 This allows to safely detach from a session without the running process getting killed.
 
 If your cluster is managed using SLURM, then [further steps may be necessary](running-slurm).
+
+## Running participants in different systems
```
fsimonis, line 48: They don't need to be different systems though.  docker containers or VMs are also possible.

https://precice.discourse.group/t/how-to-configure-precice-communication-on-kubernetes-with-tcp-port-access-control/1451

MakisH: In my view, a Docker container is a different system than its host.
Any other name I can think of is unnecessarily technical. Any suggestions?

MakisH: Instead of changing the title, I added coupling to containers and VMs explicitly as a use case, linking to Olivia's post.

Issue #237, opened 2023-01-17
Update search engine Algolia in cron job


uekerman:
Otherwise, we forget to update. Cron job could run every week (once per month might be too rarely, users are also searching for new things).

How to update: https://precice.org/docs-meta-search.html#updating-the-index

MakisH:
We could automate this and provide the API key as a GitHub secret.

chlorenz:
👍 

Currently we are on the free plan which includes 10000 'units' per month, of which we consume approximately 70%, at the moment. Which means we can't update the index more than once per month.
Algolia does have an [open source plan](https://www.algolia.com/for-open-source/) which would be perfect for us. Last time I contacted them, they weren't very responsive, we should definitely try again.

MakisH:
I will open a new Algolia account on the name of preCICE and request the FOSS plan, in case the external contact was the reason for ghosting.

I will then create a workflow executed manually and on schedule (e.g., once per month).

It needs an API key, which I will share via the GitHub secrets.

fsimonis:
We now have an open-source plan.

It may be possible to integrate the updates into the deployment pipeline of the website.
Alternative would be a cron job

uekerman:
> It may be possible to integrate the updates into the deployment pipeline of the website.
Alternative would be a cron job

Could you please investigate? (not urgent)

PR #408, opened 2024-05-13
Migrate to preCICE Algolia account, update index in a workflow

MakisH:
@chlorenz I updated the `_config.yml`, is anything else needed to use the new index?

@fsimonis I also added a workflow to update the search index every Sunday at 16:00 (or manually), due to which, this solves #237.

I have not tested the Action.

### DIFF
--- a/.github/workflows/update-algolia.yml
+++ b/.github/workflows/update-algolia.yml
@@ -1,6 +1,8 @@
 name: Update the Algolia search index
 on:
   workflow_dispatch:
+  schedule:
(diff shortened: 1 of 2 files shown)


MakisH:
@chlorenz can I merge this already, or did you want to have a look first on whether I forgot anything?

chlorenz:
Seems all good. If I remember correctly when you run the algolia gem the first time it should build the index from scratch and everything should be as before the migration. Only downside I can see is that we lose the continuity in analytics, but that's definitely worth it.

PR #414, opened 2024-06-19
Bugfix release of distribution v2211.1

BenjaminRodenberg:
For reproducibility reasons it would be good to have a bugfix release of the last version 2 distribution that I could use as a baseline in my thesis.

I added changes that I would like to add to this release in this PR. I think we should definitely let this sit for a while since I'm still doing experiments that might require extending this list.


MakisH:
Depending on https://github.com/precice/tutorials/pull/558, which we decided to close as not planned.

PR #428, opened 2024-07-18
Literature guide: Replace FTP URLs

MakisH:
Our literature guide currently includes a couple of links to FTP resources. FTP is not supported by web browsers (at least Firefox and Chrome) anymore, so the links practically don't work.

I tried to find the equivalent HTTP resources. @uekerman are these the same?


uekerman (approved):
Looks good

PR #429, opened 2024-07-18
rbfShape: Document related paper

MakisH:
@uekerman I guess this is the original paper. I link to the literature guide for more. Anything else to mention/update here?


davidscn:
For PUM, the supposed tuning does not apply anymore. I would still be in favor of removing this script.

uekerman:
Yes, this is the paper, but yes, we could also remove the script. Or maybe write that is no longer needed for v3 and PUM.

MakisH:
I added a note related to preCICE v3 and PUM.

The script itself is in the preCICE repository, so we should first remove it there (if we want that), and remove the documentation page with the next or another future release.

uekerman (approved):

PR #437, opened 2024-07-29
Add single-node MPI instructions

fsimonis:
This PR adds instructions on how to run MPI programs on a single node

### DIFF
--- a/pages/docs/running/running-simple.md
+++ b/pages/docs/running/running-simple.md
@@ -74,7 +74,7 @@ set -m
 ## Run locally in parallel using MPI
 
 If your solvers support it, your can run simulations in parallel using MPI.
-Whilst relatively simple to do on a single machine, this gets more complicated when dealing with distributed systems and job managers like SLURM, which are handled in following steps.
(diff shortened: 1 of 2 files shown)


uekerman (approved):
Good to have such things documented :+1:
I am still a bit puzzled on how this pinning is different from the pinning explained on the `running-distributed` page. Why could the pinning there not be used on a single node?

uekerman:
A link back from `running-distributed` to here could also help.

fsimonis:
@uekerman 

"Running locally" is about running the simulation locally aka on one node. We need to pin CPUs here.

"Running distributed" is about running across machines. This doesn't necessarily mean MPI. It could mean one participant runs locally on my laptop and one runs on my cluster. Or both run in separate containers.

"Running SLURM" is about running on a cluster. Here we need to distribute whole nodes allocated to the job across multiple participants.

> A link back from running-distributed to here could also help.

The entire section is kind of meant to be read linearly, right?
Where would you mention it?

uekerman:
> Where would you mention it?

Maybe a hint below https://precice.org/running-slurm.html#running-partitioned-simulations that one you could also use the "running local" pinning mechanism? Or briefly explain there how both differ.

> The entire section is kind of meant to be read linearly, right?

I don't think that we can expect this anywhere in our documentation. Users follow links or search and then end up on any page.

PR #440, opened 2024-08-08
Add imported to .gitignore

carme-hp:
Hey, is there a reason why imported is not in .gitignore? It keeps messing with my PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ _pdf
 .idea
 vendor/
 .bundle/
+imported/


fsimonis:
We solved the underlying problem differently.

MakisH:
@fsimonis Can you please document the outcomes of the discussion here, and why you don't think it's a good idea to add the directory to the `.gitignore`?

fsimonis:
The problem was that the submodules in imported don't update when changing branches. So there can be changes displayed even though the submodules haven't been touched.

The solution is to ignore the submodules and only git add the relevant changes.

If the submodule content is important for the changes, the use `git pull --recurse-submodules` to bring everything up to date.

MakisH:
> The solution is to ignore the submodules

But the robust solution would be to add the complete `imported/` directory in `.gitignore`, so that people don't accidentally commit such changes, no?

fsimonis:
`.gitignore` ignores untracked files, not changes.

MakisH:
Ah, right, good point...
This would have been enough of a review.

PR #442, opened 2024-08-09
Simplify the SLURM hostfile setup

fsimonis:
This PR simplifies the SLURM explanation.
Hostfiles are now generated using slot notations.
This allows to run mpi without `-n` as it scales to all available slots.

The hostfiles now don't contain repeated hosts, as repeating hosts could still lead to double allocation of CPUs on nodes that used by multiple hostfiles.

@davidscn could you test these formats on the AMD cluster?

I also created a python version of the script to generate hostfiles for an active SLURM session. Is this actually useful, or is it easier to code the generation into the bash of the SLURM jobs? If it is useful, where should it go? In a gist, added to the website as a file, added as text, or added to a separate repo in the org?
https://gist.github.com/fsimonis/4e312c3875c276d96f358bb0ff8ce7a2

### DIFF
--- a/pages/docs/running/running-slurm.md
+++ b/pages/docs/running/running-slurm.md
@@ -53,12 +53,12 @@ To generate a file containing a list of all hosts use:
 #!/bin/bash
 rm -f hosts.intel host.ompi
 for host in `scontrol show hostname $SLURM_JOB_NODELIST`; do
-  # IntelMPI requires one entry per node
-  echo $host >> hosts.intel;
-  # OpenMPI requires one entry per slot
-  for j in $(seq 1 ${SLURM_TASKS_PER_NODE%%(*}); do
-    echo $host >> hosts.ompi;
-  done
+  # IntelMPI, MPICH, and MVAPICH2 use the column notation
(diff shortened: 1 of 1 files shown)


fsimonis:
Merging now as it's definitely an improvement.

uekerman:
>  In a gist, added to the website as a file, added as text, or added to a separate repo in the org?

Do you link to the gist from the website? That would be good already.

fsimonis:
No, the gist is still completely separate until we know where to put it.

uekerman:
I used to code the generation in the SLURM bash files, but I still think that this could be useful. Simply linking to the gist from the website is enough in my opinion. If we don't do it now, it will get lost :grin:

fsimonis:
> If we don't do it now, it will get lost 😁

Good point

A valuable addition for the script could be the automatic detection of the MPI implementation.

That would be tedious in bash.

fsimonis:
@uekerman Done https://github.com/precice/precice.github.io/commit/0f39d2a88d779f8e28de1db9c37816593507bc1f

PR #453, opened 2024-09-16
Restore comments in news_banner.html (and disable banner)

MakisH:
These comments were moved to the [README.md](https://github.com/precice/precice.github.io/blob/master/README.md) in https://github.com/precice/precice.github.io/commit/d68cf8afc84bbb71f050a564d73a076fe056f395, but I think they are useful in the file as well, for quick access.

Even if we end up disagreeing about this, we need at least a link to the README from the file itself (and potentially move the whole thing into https://precice.org/docs-meta-overview.html or a subpage).

Also disables the banner of the workshop registration, per @IshaanDesai's request.

### DIFF
--- a/_includes/news_banner.html
+++ b/_includes/news_banner.html
@@ -1,7 +1,14 @@
-{% if include.onlanding %}
-{% else %}
-{% endif %}
+{% comment %}{% raw %}
 
+For landing-only news use:
+{% if include.onlanding %}
+
+For news visible on all pages use:
+{% if true %}.
+
+{% endraw %}{% endcomment %}
+
+{% if false %}
 <div class="background-light banner-container">
   <div class="container">
     <div class="row no-margin">
@@ -13,3 +20,5 @@
     </div>
   </div>
 </div>
+{% else %}
+{% endif %}


fsimonis (commented):
Firstly, can we reword the workshop banner instead of disabling it?
I find the banner super useful to quickly get to the workshop page.
Example: `Attending the preCICE Workshop 2024? Click to see the program.`

Secondly, I personally think there is 0 reason to duplicate the comments in the include file, to keep the code around, and to link back to the Readme. If you know this file exists, then you don't need the comments. If you don't, then you need to read the Readme.

I moved the code example and documentation to the Readme because it is easier to find, allows for more verbose documentation, and doesn't mess with formatting of the rest of the file due to nested tags.

This is not important, you don't need my review, and won't get my OK.

IshaanDesai:
Having a workshop banner still is a good idea indeed :+1: Let us indeed have `Attending the preCICE Workshop 2024? Click to see the program.`

I stay clear of the other crossfire.

MakisH:
> Firstly, can we reword the workshop banner instead of disabling it? I find the banner super useful to quickly get to the workshop page. Example: `Attending the preCICE Workshop 2024? Click to see the program.`

Sure, this was just implementing a request.

> Secondly, I personally think there is 0 reason to duplicate the comments in the include file, to keep the code around, and to link back to the Readme. If you know this file exists, then you don't need the comments. If you don't, then you need to read the Readme.

My journey leading to this:

- I had previously searched for the content of the banner to find the right file (so, my entry point it different)
- I saw the content outside an empty `if` structure
- I looked into the git blame to find again what options were there, saw the bit about documentation
- Went to the documentation page on the website, did not see it there
- Went back to the previous PR, so it was actually in the README

Next time I will probably go again to the documentation, not find anything there, and then end up by chance finding something in the README.

Sure, let's not duplicate, but do we agree on the need for a back-reference, due to the different entry point?

> I moved the code example and documentation to the Readme because it is easier to find, allows for more verbose documentation, and doesn't mess with formatting of the rest of the file due to nested tags.
> 
> This is not important, you don't need my review, and won't get my OK.

I understand that the file looks messier like this.

I will open another PR suggesting to move the documentation to a different place and add a link to that. I am just looking for a better solution.

PR #454, opened 2024-09-16
Update Python installation instructions in community-training.md

MakisH:
There are several more places on the website and our documentation in general that need updating (including the Python bindings), this is only focusing on the training page.

Note: I am wondering if the suggestion to create the venv as `.venv` conflicts with the run scripts that also call `python -m venv .venv` (and later try to re-source the venv). I can eventually try this.

### DIFF
--- a/pages/community/community-training.md
+++ b/pages/community/community-training.md
@@ -57,8 +57,10 @@ In case you would like something more permanent, but still with low setup effort
 Finally, in case you prefer to install everything in your laptop, you will need the following:
 
 - [preCICE](installation-overview.html)
-- [preCICE Python bindings](installation-bindings-python.html): `pip3 install --user pyprecice` (check with running `import precice` in a Python interpreter)
-- matplotlib: `pip3 install --user matplotlib`
+- [preCICE Python bindings](installation-bindings-python.html):
+  - Create a virtual environment: `python3 -m venv .venv && source .venv/bin/activate`
+  - Install the bindings: `pip3 install pyprecice` (check with running `import precice` in a Python interpreter)
+- matplotlib: In the same virtual environment, run `pip3 install matplotlib`
 - [ParaView](https://www.paraview.org/) (visualization, used in most modules apart from the basics)
 
 The tools module also needs (all optional):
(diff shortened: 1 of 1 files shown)


pages/community/community-training.md
```diff
@@ -83,5 +85,5 @@ The FSI workflow module also needs:
 - [OpenFOAM-preCICE](adapter-openfoam-get.html) (check with running the Quickstart tutorial)
 - CalculiX and the [CalculiX adapter](https://precice.org/adapter-calculix-overview.html) (check by running `ccx_preCICE`)
 - [FreeCAD](https://www.freecad.org/) 0.21 or later (check by starting the GUI)
-- [ccx2paraview](https://github.com/calculix/ccx2paraview) (check by starting a Python terminal and executing `import ccx2paraview`)
-- [PyFoam](https://pypi.org/project/PyFoam/) (check by running `pyFoamPlotWatcher.py --help`)
+- [ccx2paraview](https://github.com/calculix/ccx2paraview) (`pip3 install ccx2paraview`, check by starting a Python terminal and executing `import ccx2paraview`)
```
fsimonis, line 88: Do you need to use this as a library in the course?
If no, then we should use `pipx install ccx2paraview` and run `ccx2paraview --help` instead

pages/community/community-training.md
```diff
@@ -57,8 +57,10 @@ In case you would like something more permanent, but still with low setup effort
 Finally, in case you prefer to install everything in your laptop, you will need the following:
 
 - [preCICE](installation-overview.html)
-- [preCICE Python bindings](installation-bindings-python.html): `pip3 install --user pyprecice` (check with running `import precice` in a Python interpreter)
-- matplotlib: `pip3 install --user matplotlib`
+- [preCICE Python bindings](installation-bindings-python.html):
+  - Create a virtual environment: `python3 -m venv .venv && source .venv/bin/activate`
```
fsimonis: Maybe a hint like this could be useful:

> As long as the environment is active, you will see `(venv)` before your command prompt. You need to activate the venv in new terminal windows.

An alternative to activating the prompt could be:

> Install the bindings in the virtual environment:
> .venv/bin/pip install pyprecice

pages/community/community-training.md
MakisH: Does this work for you, however? For me, I get:

```
pipx install ccx2paraview

No apps associated with package ccx2paraview or its dependencies. If you are attempting to install a library, pipx should not be used.
Consider using pip or a similar tool instead.
```

Same if I do `pipx install 'ccx2paraview[VTK]'`.

fsimonis: Ah right. The release of cxx2paraview which allows this is still pending

MakisH: Then, I would say let's proceed as-is for now?
It's not that we don't have other instances in our website where pipx would have been better. We will anyway need to update various parts to address the venv requirement.

fsimonis (approved):

MakisH:
I tried following the instructions for the CHT part of the course (Nutils), and having a venv already activated does not conflict: the `run.sh` script of `solid-nutils` creates and activates a new venv in a local directory `.venv` and re-installs the requirements, using cached packages.

Creating the venv directly in the place where the run script automatically creates a new venv does not matter as well (even if this is not a realistic use case), as the directory is not overwritten, and no error is thrown.

If anything, we might need to improve the tutorials assumptions.

PR #460, opened 2024-10-02
Update algolia daily on changes

### LINKED ISSUE #237: Update search engine Algolia in cron job

fsimonis:
This should be a good compromise between scheduled updates and on-time updates.

It runs daily if the latest git commit is less than a day old.

Closes #237

### DIFF
--- a/.github/workflows/update-algolia.yml
+++ b/.github/workflows/update-algolia.yml
@@ -2,7 +2,7 @@ name: Update the Algolia search index
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


MakisH (approved):
Thanks! The strategy makes sense, the code makes sense, I have not tested anything.

fsimonis:
Let's test in production as usual with actions :wink:

PR #462, opened 2024-10-17
Update installation overview figure

MakisH:
I noticed that one arrow in the overview figure did not make much sense: the question after "I use Ubuntu / Arch" is still "I have all the dependencies", because Spack is still an option.


uekerman (approved):
Yes

PR #464, opened 2024-10-23
Add image hook for lazily loading images

fsimonis:
This PR adds a hook that automatically adds css aspect ratio information of existing images and sets them to be lazily loaded.

This speeds up page loading and prevents layout shift


fsimonis:
Closing as we are not planning to invest further into the jekyll backend.

PR #466, opened 2024-10-31
Add posix-spawn to gemfile.lock

MakisH:
As suggested by the build failure message in https://github.com/precice/precice.github.io/actions/runs/11479909314/job/31947178415 (commit https://github.com/precice/precice.github.io/commit/d6433acc2fbb005c03d9f119315eac63c4f57d44):

```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.
```


fsimonis (approved):
Thanks! Forgot to push that part.

PR #474, opened 2024-11-29
Fix testprecice labels argument

uekerman:
@fsimonis correct?

I only get:

```
./testprecice --list_labels
This test suite runs on rank 0 of 1
Available labels:
  MPI_Ports
```

### DIFF
--- a/pages/docs/dev-docs/dev-testing.md
+++ b/pages/docs/dev-docs/dev-testing.md
@@ -55,7 +55,7 @@ The latter is only useful for running individual tests in a debugger.
 
 Tests in CTest are named `precice.` followed by the full test name of boost test.
 To get a list of all tests for scripting purposes, use `./testprecice --list_units`.
-CTest additionally groups tests in labels. Use `./testprecice --print-labels` to display existing labels.
+CTest additionally groups tests in labels. Use `./testprecice --list_labels` to display existing labels.
 
 Some important options for `ctest` are:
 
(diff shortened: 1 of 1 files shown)


pages/docs/dev-docs/dev-testing.md
```diff
@@ -55,7 +55,7 @@ The latter is only useful for running individual tests in a debugger.
 
 Tests in CTest are named `precice.` followed by the full test name of boost test.
 To get a list of all tests for scripting purposes, use `./testprecice --list_units`.
-CTest additionally groups tests in labels. Use `./testprecice --print-labels` to display existing labels.
+CTest additionally groups tests in labels. Use `./testprecice --list_labels` to display existing labels.
```
fsimonis: CTest knows the labels, not testprecice. So the mistake is:

```suggestion
CTest additionally groups tests in labels. Use `ctest --print-labels` to display existing labels.
```

uekerman: OK, this works, but now also executes the tests? I expected that it only prints the labels without running the tests. I must admit I don't understand yet what these labels are.

fsimonis: The argument is `--print-labels` with a dash. Fixed it above.
CTest and CMake have very strange ways of handling incorrect input.

Tests can have multiple labels. You can use these labels to include/exclude entire subsets of tests.

uekerman: Looks good now :+1:

fsimonis (approved):

PR #475, opened 2024-12-06
Document release schedule

MakisH:
As decided in the coding days, Stuttgart, November 2024

Will appear in https://precice.org/dev-docs-release-strategy.html


pages/docs/dev-docs/release-strategy.md
```diff
@@ -38,6 +38,19 @@ permalink: dev-docs-release-strategy.html
 5. Make sure everything is working
 6. Release R
 
+## Release schedule
+
+The core library repository follows semantic versioning.
+
+We group together breaking changes and release a new breaking version as needed, but not more frequently than once every two years (aiming for longer). This aims to give time and motivation to users to adopt new versions, and spreads the effort of preparing a release over a longer period of time. There is no fixed schedule for major releases, and we take time to ensure that no further breaking changes are needed soon after.
```
BenjaminRodenberg: Do we already have a section where we explain the general workflow for a breaking release? This is not directly related to the schedule but I think it would be helpful to at least point to this workflow from here. 

Some points that are important in my opinion:

* What we did last time: At some point there was a `v3_develop` branch in parallel to the `develop` branch. At some later point we decided to merge `v3_develop` into `develop`. This closed the door for any v2 releases and forced everybody to integrate new features directly with v3.
* I think it would also be helpful to point to the milestone that we usually create in order to cluster breaking releases way before the breaking release is scheduled (see https://github.com/precice/precice/milestone/20).

pages/docs/dev-docs/release-strategy.md
```diff
@@ -38,6 +38,19 @@ permalink: dev-docs-release-strategy.html
...
+
+We aim for feature releases twice per year, so that new features can be continuously shipped to new users, but we again don't get overly burdened by the release process, and feature releases maintain their significance. We define a feature freeze date (date after which no new feature contributions can be merged) and a release date (date of merging to the main branch and tagging the release). Since most of the development happens in universities, we align these dates with the lecture-free periods between semesters in these universities:
```
BenjaminRodenberg: I think the feature freeze is still unclear. What does this mean? We freeze `develop` between the feature freeze and the release? According to the plan below this means that up to 40 days per release and 80 days per year `develop` is blocked and no integration of new features for the complete team is possible. Do we want this?

I would prefer to create a release branch from `develop` on the date of the feature freeze since this would allow to continue working on `develop`. But we probably discussed this and there was some argument for not having a release branch.

pages/docs/dev-docs/release-strategy.md
```diff
@@ -38,6 +38,19 @@ permalink: dev-docs-release-strategy.html
...
+
+- spring release: feature freeze on March 20 at 23:59 CET, release as soon as possible in April
+- fall release: feature freeze on September 20 at 23:59 CET, release as soon as possible in October
```
BenjaminRodenberg, line 50: Ok :+1:

pages/docs/dev-docs/release-strategy.md
MakisH: As far as I know, we don't yet document anything special regarding breaking releases.
I would prefer to keep this point completely separate from this PR, as it goes in a different direction, with a lot more content and required iterations.

pages/docs/dev-docs/release-strategy.md
MakisH: The "Release procedure" section in this page ([above](https://precice.org/dev-docs-release-strategy.html#release-procedure)), as well as the PR template currently specifie a release branch: https://github.com/precice/precice/blob/develop/tools/releasing/PULL_REQUEST_TEMPLATE/release_pull_request_template.md

I now modified the definition to explicitly mention the release branch.

pages/docs/dev-docs/release-strategy.md
BenjaminRodenberg: I agree. How about moving these points into an issue or other PR? (we could also point there from here)

fsimonis (approved):

uekerman (approved):

BenjaminRodenberg (approved):

pages/docs/dev-docs/release-strategy.md
MakisH: https://github.com/precice/precice.github.io/issues/477

MakisH: After discussing with @davidscn, I am removing the explicit number of years, as the exact frequency is not what is rigidly decided upon:

```suggestion
We group together breaking changes and release a new breaking (major) version as needed. There is no fixed schedule for major releases, and we take time to ensure that no further breaking changes are needed soon after. This aims to give time and motivation to users to adopt new versions, and spreads the effort of preparing a release over a longer period of time.
```

PR #478, opened 2024-12-12
Fix layout semantics

fsimonis:
This PR fixes the landmark semantics of the HTML layout using the [Accessibility insights for web](https://accessibilityinsights.io/docs/web/overview/) inspectors.

- sidebar and table of content navigations are now correctly wrapped in `nav` tags and use `aria-label` to describe which label they are
- the `main` content does not contain elements which are repeated over the website
- the website footer is outside of the `main` tag
- the sidebar is outside of the `main` tag
- the news banner is a `<aside role="banner">` and outside the `main` tag
- the edit section is wrapped in an `aside`

| | Doc | Home |
--- | --- | ---
Old|old doc | old home
New|new doc | new home

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,5 +1,6 @@
-<footer>
-	<div class="row text-center">
+<div class="container">
+  <div class="row text-center">
+    <hr class="faded">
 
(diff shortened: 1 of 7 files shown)


MakisH (approved):
Sounds like a good change that should not harm anything.
I only had a quick look, and I do not have enough expertise in web development to judge the technicalities.

PR #481, opened 2024-12-18
Remove version from docs sidebar

MakisH:
Do we really need the version number in the sidebar?

Screenshot from 2024-12-18 13-28-52

- It is not very accurate: We sometimes include forward-referencing content, and it is not fixed at the time of the respective release.
- It is misleading: users think that they can click somewhere to change that (user observation at the SSE).
- It is yet another hard-coded place to update.

This PR removes the version and expands "Docs" to "Documentation".

I have not tested how this change renders.

### DIFF
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -2,8 +2,8 @@
 
 entries:
 - title: sidebar
-  product: Docs
-  version: v3.1.2
+  product: Documentation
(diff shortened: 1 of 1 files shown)


uekerman (approved):
I am OK with the change, but neither have tested the rendering.
Further problem: The docs is not only preCICE, but many components with different versions. The "last edited on" feature is much more transparent than this.

PR #483, opened 2025-01-13
Update news banner: workshop instead of MS CfC

uekerman:
@MakisH A bit unrelated, but we still need to update the workshop logo to include Sep 8.


MakisH (approved):
News banner looks good.
I updated the banner in this PR.

PR #490, opened 2025-01-29
Make macOS instructions more easily discoverable

MakisH:
@davidscn is this enough, or should we link from further places? Which path did the user take?

### DIFF
--- a/pages/docs/installation/installation-overview.md
+++ b/pages/docs/installation/installation-overview.md
@@ -34,7 +34,7 @@ Linux is the easiest option, see figure above. We provide [binary packages](inst
 
 **macOS**
 
-The recommended way is to use [Homebrew](https://brew.sh/) to install the preCICE dependencies and then [compile preCICE from source](installation-source-preparation.html). You can alternatively [build preCICE using Spack](https://precice.org/installation-spack.html).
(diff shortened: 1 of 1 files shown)


pages/docs/installation/installation-overview.md
```diff
@@ -34,7 +34,7 @@ Linux is the easiest option, see figure above. We provide [binary packages](inst
 
 **macOS**
 
-The recommended way is to use [Homebrew](https://brew.sh/) to install the preCICE dependencies and then [compile preCICE from source](installation-source-preparation.html). You can alternatively [build preCICE using Spack](https://precice.org/installation-spack.html).
+The recommended way is to use [Homebrew](https://brew.sh/) to install the preCICE dependencies and then [compile preCICE from source](installation-source-preparation.html) (see [how to install the dependencies on macOS](installation-source-dependencies.html#macos)). You can alternatively [build preCICE using Spack](https://precice.org/installation-spack.html).
```
davidscn: Maybe we can even put the link into the homebrew. Do we really need to link to homebrew itself? 
The user asked me how I found the page and I used the searchbar. We have many more systems with packages explicitly mentioned. What do we do about these? I would make the whole https://precice.org/installation-source-dependencies.html more visible, as it covers use cases for quite a range of users.

PR #492, opened 2025-02-04
Documentation for Just in time mapping

davidscn:
Documentation for precice/precice#2099

### DIFF
--- a/pages/docs/configuration/configuration-mapping.md
+++ b/pages/docs/configuration/configuration-mapping.md
@@ -12,10 +12,6 @@ Therefore, preCICE provides data mapping methods to map coupling data from one m
 The mapping configuration has undergone a major revision between preCICE version 2 and preCICE version 3. For the documentation of version 2, checkout the documentation of [our previous versions](fundamentals-previous-versions.html).
 {% endnote %}
 
-{% tip %}
-Instead of using static meshes as described here, we are currently working on a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
-{% endtip  %}
-
 Each data mapping definition refers to two meshes in the participant configuration: a `provide` mesh defined by the participant and a `receive` mesh defined by another participant (e.g. `MySolver2`):
 
 ```xml
@@ -67,6 +63,10 @@ Each mapping defines a `constraint`, which defines how the data is mapped betwee
 * `scaled-consistent-surface` and `scaled-consistent-volume` constraint: `scaled-consistent` constraints are used for intensive quantities (just as the `consistent` constraint) where conservation of integral values (surface or volume) is necessary (e.g. velocities when the mass flow rate needs to be conserved). The mapping executes a `consistent` mapping in a first step, and corrects the result by a subsequent scaling step using the integral data sum to ensure the conservation of the integral sum on the input mesh and the output mesh. To use the `scaled-consistent-surface` constraint, surface connectivity on input and output meshes is required. To use the `scaled-consistent-volume` constraint, volumetric connectivity on input and output meshes is required.
 
 For a sequential participant, any combination of `read`/`write`-`consistent/conservative` is valid. For a participant running in parallel, only `read`-`consistent` and `write`-`conservative` is possible. More details are given [further below](configuration-mapping.html#restrictions-for-parallel-participants).
+
+{% tip %}
+Instead of using static meshes as described here, you may consider using a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
+{% endtip  %}
 
 ## Available mapping methods in preCICE
 
--- a/pages/docs/configuration/configuration-mapping.md
+++ b/pages/docs/configuration/configuration-mapping.md
@@ -65,7 +65,7 @@ Each mapping defines a `constraint`, which defines how the data is mapped betwee
 For a sequential participant, any combination of `read`/`write`-`consistent/conservative` is valid. For a participant running in parallel, only `read`-`consistent` and `write`-`conservative` is possible. More details are given [further below](configuration-mapping.html#restrictions-for-parallel-participants).
 
 {% tip %}
-Instead of using static meshes as described here, you may consider using a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
+Instead of using static meshes as described here, you may consider using a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where one of both meshes can change at runtime.
 {% endtip  %}
 
 ## Available mapping methods in preCICE
--- a/pages/docs/configuration/configuration-mapping.md
+++ b/pages/docs/configuration/configuration-mapping.md
@@ -12,6 +12,10 @@ Therefore, preCICE provides data mapping methods to map coupling data from one m
 The mapping configuration has undergone a major revision between preCICE version 2 and preCICE version 3. For the documentation of version 2, checkout the documentation of [our previous versions](fundamentals-previous-versions.html).
 {% endnote %}
 
+{% tip %}
+Instead of using static meshes as described here, we are currently working on a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
+{% endtip  %}
+
 Each data mapping definition refers to two meshes in the participant configuration: a `provide` mesh defined by the participant and a `receive` mesh defined by another participant (e.g. `MySolver2`):
 
 ```xml
--- a/pages/docs/couple-your-code/couple-your-code-direct-access.md
+++ b/pages/docs/couple-your-code/couple-your-code-direct-access.md
@@ -5,9 +5,7 @@ keywords: api, adapter, mapping, meshes
 summary: "You can access received meshes and their data directly by using specific optional API functions."
 ---
 
-<!-- TODO: needs update -->
-
-This concept is required if you want to access received meshes directly. It might be relevant in case you don't want to use the mapping schemes in preCICE, but rather want to use your own solver for data mapping. As opposed to the usual preCICE mapping, only a single mesh (from the other participant) is now involved in this situation since an 'own' mesh defined by the participant itself is not required any more. In order to re-partition the received mesh, the participant needs to define the mesh region it wants read data from and write data to. The complete concept on the receiving participant looks as follows:
+This concept is required if you want to access received meshes directly. It might be relevant in case you don't want to use the mapping schemes in preCICE, but rather want to use your own solver for data mapping. As opposed to the usual preCICE mapping, only a single mesh (from the other participant) is now involved in this situation since an 'own' mesh defined by the participant itself is not required any more. In order to re-partition the received mesh, the participant needs to define the mesh region it wants to read data from and write data to. The complete concept on the receiving participant looks as follows:
 
 ```cpp
     // Note that "ReceivedMeshname" is not a name of a
@@ -55,7 +53,7 @@ This concept is required if you want to access received meshes directly. It migh
 ```
 
 {% tip %}
-If you want to have the flexibility of direct mesh access, but still want to use data mappings provided by preCICE, you may want to use [just-in-time data mappings](couple-your-code-just-in-time-mapping.html) in preCICE.
+If you want to have the flexibility of direct mesh access, but still want to use data mappings provided by preCICE, you may want to use [just-in-time data mappings](couple-your-code-just-in-time-mapping.html).
 {% endtip  %}
 
 ## Concept and API
--- a/pages/docs/couple-your-code/couple-your-code-direct-access.md
+++ b/pages/docs/couple-your-code/couple-your-code-direct-access.md
@@ -54,6 +54,10 @@ This concept is required if you want to access received meshes directly. It migh
 
 ```
 
+{% tip %}
+If you want to have the flexibility of direct mesh access, but still want to use data mappings provided by preCICE, you may want to use [just-in-time data mappings](couple-your-code-just-in-time-mapping.html) in preCICE.
+{% endtip  %}
+
 ## Concept and API
 
 Defining a bounding box for serial runs of the solver (not to be confused with serial coupling mode) is valid. However, a warning is raised in case vertices are filtered out completely on the receiving side, since the associated data values of the filtered vertices are filled with zero data values in order to make the 'read' operation of the other participant valid.
@@ -63,11 +67,15 @@ In order to use the feature, it needs to be enabled explicitly in the configurat
 ```xml
 ...
 <participant name="MyParticipant">
-  <receive-mesh name="ReceivedMeshName" from="OtherParticipant" direct-access="true" />
+  <receive-mesh name="ReceivedMeshName" from="OtherParticipant" api-access="true" />
   <write-data name="WriteDataName" mesh="ReceivedMeshName" />
 </participant>
 ...
 ```
+
+{% version 3.2.0 %}
+The flag on the received mesh was called `<receive-mesh ... direct-access="true" />` instead of `<receive-mesh ... api-access="true" />` before preCICE version 3.2.0.
+{% endversion %}
 
 Note that we write the data on a mesh we received and no mapping and no mesh need to be defined as opposed to the usual case. If you want to read data on a provided mesh additionally, a mesh can (and must) be provided, as usual. Note also that you probably need to reconfigure the mesh, which is used for the data exchange (`<exchange data=..`), the data acceleration and convergence measure within the coupling scheme. Minimal configuration examples can also be found in the integration tests located in the preCICE repository `precice/src/precice/tests`. All relevant test files have '`direct-access`' in the file name, e.g. `explicit-direct-access.xml`.
 
--- a/pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
+++ b/pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
@@ -38,7 +38,7 @@ Using just-in-time mapping requires small configuration changes in the preCICE c
 2. The `read-data` and `write-data` tags need to use the received mesh with api-access enabled.
 3. The mapping only defines (depending on the direction) a "from" or a "to" mesh. For read-consistent mappings, only the "from" mesh must be defined. For write-conservative mappings, only the "to" mesh must be defined. In both cases, the defined mesh needs to match the received mesh with api-access enabled (which is also used in the `read-data` or `write-data` tag).
 
-At its core, the API now makes use of two new API functions, called `mapAndReadData` and `writeAndMapData`. Both functions are very similar to the analogous functions `readData` and `writeData`: the only difference is that the functions take spatial coordinates as function arguments instead of static vertexIDs stemming from the preCICE initialization. In addition to `mapAndReadData` and `writeAndMapData`, the access region needs to be defined by "SolverOne" using `setMeshAccessRegion`. The full code example reads the following:
+The API makes use of two new API functions, called `mapAndReadData` and `writeAndMapData`. Both functions are very similar to the analogous functions `readData` and `writeData`: The only difference is that the functions take spatial coordinates as function arguments instead of static vertexIDs. In addition to `mapAndReadData` and `writeAndMapData`, the access region needs to be defined by "SolverOne" using `setMeshAccessRegion`. A full code example reads as follows:
 
 ```cpp
     // Note that "ReceivedMeshName" is defined and received
--- a/pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
+++ b/pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
@@ -13,7 +13,32 @@ This feature is new in preCICE version 3.2.0.
 This is an experimental feature.
 {% endexperimental %}
 
-Just-in time data mapping is related to the [Solver-based data mapping](couple-your-code-direct-access.html) in the sense that a participant doesn't need to provide a mesh.
+Just-in time data mapping combines the idea of a [direct-mesh access](couple-your-code-direct-access.html) with the [conventional mapping](configuration-mapping.html) methods in preCICE: a solver does not provide a static mesh during the initialization phase, but instead defines an access region, in which vertices and data can be read and written to and from preCICE just-in-time. Using the just-in-time API (see below) of preCICE comes with performance implications: reading and writing becomes more expensive, but it gives the user more flexibility for moving meshes. The feature was originally designed with mesh-particle coupling in mind: one participant (the mesh-based participant) defines a static mesh during initialization, and the other participant (particle participant) defines a spatial access region, where data can be provided just-in-time along with the current particle position. The feature might be useful for other scenarios as well. The following sections describe the API the configuration and limitations of the current implementation.
+
+## Concept and API
+
+Using Just-in-time mapping requires smaller configuration changes in the precice configuration file and the source code of the adapter. The following configuration is an example for a just-in-time mapping preCICE configuration of our example participant called "SolverOne":
+
+```xml
+  <participant name="SolverOne">
+    <receive-mesh name="ReceivedMeshName" from="SolverTwo" api-access="true" />
+    <!-- data reading and writing is performed on MeshTwo, which is a received mesh with api-access enabled -->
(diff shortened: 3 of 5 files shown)


pages/docs/configuration/configuration-mapping.md
```diff
@@ -12,6 +12,10 @@ Therefore, preCICE provides data mapping methods to map coupling data from one m
 The mapping configuration has undergone a major revision between preCICE version 2 and preCICE version 3. For the documentation of version 2, checkout the documentation of [our previous versions](fundamentals-previous-versions.html).
 {% endnote %}
 
+{% tip %}
+Instead of using static meshes as described here, we are currently working on a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
+{% endtip  %}
```
uekerman: Could we move this tip further down? Maybe even the note above as well. Let's start with the main content first and explain alternatives later.
And please make the docs timeless. "we are currently working on" will be outdated eventually and we will forget to update it.

pages/docs/couple-your-code/couple-your-code-direct-access.md
uekerman, line 1: Could we touch the

`<!-- TODO: needs update -->` 

in line 8 in this PR as well?

pages/docs/couple-your-code/couple-your-code-direct-access.md
```diff
@@ -54,6 +54,10 @@ This concept is required if you want to access received meshes directly. It migh
 
 ```
 
+{% tip %}
+If you want to have the flexibility of direct mesh access, but still want to use data mappings provided by preCICE, you may want to use [just-in-time data mappings](couple-your-code-just-in-time-mapping.html) in preCICE.
```
uekerman:
```suggestion
If you want to have the flexibility of direct mesh access, but still want to use data mappings provided by preCICE, you may want to use [just-in-time data mappings](couple-your-code-just-in-time-mapping.html).
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
+---
+title: Just-in-time data mapping
+permalink: couple-your-code-just-in-time-mapping.html
+keywords: api, adapter, mapping, meshes, just-in-time
```
uekerman:
```suggestion
keywords: api, adapter, mapping, meshes, just-in-time, mesh-particle coupling, DEM
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+summary: "You can read and write data by providing dynamic coordinates instead of static vertex IDs using specific optional API functions."
```
uekerman:
```suggestion
summary: "You can read and write data at dynamic coordinates instead of static vertex IDs using specific optional API functions."
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+{% endversion %}
+
+{% experimental %}
+This is an experimental feature.
+{% endexperimental %}
+
+Just-in time data mapping combines the idea of a [direct-mesh access](couple-your-code-direct-access.html) with the [conventional mapping](configuration-mapping.html) methods in preCICE: a solver does not provide a static mesh during the initialization phase, but instead defines an access region, in which vertices and data can be read and written to and from preCICE just-in-time. Using the just-in-time API (see below) of preCICE comes with performance implications: reading and writing becomes more expensive, but it gives the user more flexibility for moving meshes. The feature was originally designed with mesh-particle coupling in mind: one participant (the mesh-based participant) defines a static mesh during initialization, and the other participant (particle participant) defines a spatial access region, where data can be provided just-in-time along with the current particle position. The feature might be useful for other scenarios as well. The following sections describe the API the configuration and limitations of the current implementation.
```
uekerman:
```suggestion
Just-in time data mapping combines the idea of [direct-mesh access](couple-your-code-direct-access.html) with the [conventional mapping](configuration-mapping.html) methods in preCICE: a solver does not provide a static mesh during the initialization phase, but instead defines an access region, in which vertices and data can be read and written to and from preCICE just-in-time. Using the just-in-time API (see below) of preCICE comes with performance implications: reading and writing becomes more expensive, but it gives the user more flexibility for moving meshes. The feature was originally designed with mesh-particle coupling in mind: one participant (the mesh-based participant) defines a static mesh during initialization, and the other participant (particle participant) defines a spatial access region, where data can be provided just-in-time along with the current particle position. The feature might be useful for other scenarios as well. The following sections describe the API, the configuration, and the limitations of the current implementation.
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Concept and API
+
+Using Just-in-time mapping requires smaller configuration changes in the precice configuration file and the source code of the adapter. The following configuration is an example for a just-in-time mapping preCICE configuration of our example participant called "SolverOne":
```
uekerman:
```suggestion
Using just-in-time mapping requires small configuration changes in the preCICE configuration file and the source code of the adapter. The following configuration is an example for a just-in-time mapping configuration of an example participant called "SolverOne":
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+```xml
+  <participant name="SolverOne">
+    <receive-mesh name="ReceivedMeshName" from="SolverTwo" api-access="true" />
+    <!-- data reading and writing is performed on MeshTwo, which is a received mesh with api-access enabled -->
```
uekerman:
```suggestion
    <!-- data reading and writing is performed on ReceivedMeshName, which is a received mesh with api-access enabled -->
```
Maybe "ReceivedMesh" is better than "ReceivedMeshName"? "SolverTwo-Mesh" would still be good IMO.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    <read-data name="Velocities" mesh="ReceivedMeshName" />
+    <write-data name="Forces" mesh="ReceivedMeshName" />
+    <!-- define the just-in-time mapping (read-consistent) to read the velocities, note the empty "to" mesh for the read direction-->
+    <mapping:nearest-neighbor direction="read" from="ReceivedMeshName" constraint="consistent" />
+    <!-- define the just-in-time mapping (write conservative) to write forces just-in-time, note the empty "from" mesh for the write direction-->
```
uekerman:
```suggestion
    <!-- define the just-in-time mapping (write-conservative) to write forces, note the empty "from" mesh for the write direction-->
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+## Concept and API
+
+Using Just-in-time mapping requires smaller configuration changes in the precice configuration file and the source code of the adapter. The following configuration is an example for a just-in-time mapping preCICE configuration of our example participant called "SolverOne":
+
+```xml
+  <participant name="SolverOne">
+    <receive-mesh name="ReceivedMeshName" from="SolverTwo" api-access="true" />
+    <!-- data reading and writing is performed on MeshTwo, which is a received mesh with api-access enabled -->
+    <read-data name="Velocities" mesh="ReceivedMeshName" />
+    <write-data name="Forces" mesh="ReceivedMeshName" />
```
uekerman:
```suggestion
    <read-data name="Velocity" mesh="ReceivedMeshName" />
    <write-data name="Force" mesh="ReceivedMeshName" />
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    <!-- define the just-in-time mapping (read-consistent) to read the velocities, note the empty "to" mesh for the read direction-->
+    <mapping:nearest-neighbor direction="read" from="ReceivedMeshName" constraint="consistent" />
+    <!-- define the just-in-time mapping (write conservative) to write forces just-in-time, note the empty "from" mesh for the write direction-->
+    <mapping:nearest-neighbor direction="write" to="ReceivedMeshName" constraint="conservative" />
+  </participant>
+```
+
+Here, "SolverOne" now defines a just-in-time mapping access in read-direction for data "Velocities" and in write direction for data "Forces". In its core, using just-in-time mapping requires three changes in the configuration file:
```
uekerman:
```suggestion
"SolverOne" defines access through just-in-time mappings in read direction for velocities and in write direction for forces. In its core, using just-in-time mappings requires three changes in the configuration file:
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+summary: "You can read and write data by providing dynamic coordinates instead of static vertex IDs using specific optional API functions."
+---
+
+{% version 3.2.0 %}
+This feature is new in preCICE version 3.2.0.
+{% endversion %}
+
+{% experimental %}
+This is an experimental feature.
+{% endexperimental %}
```
uekerman: We should add somewhere (here?) that the experimental switch in the config needs to be enabled, right?

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+1. `api-access` needs to be enabled on the `receive-mesh` (similar to the direct access), "ReceivedMeshName" in our example above.
+2. The `read-data` and `write-data` tags need to use the received mesh with api-access enabled.
+3. The mapping only defines (depending on the direction) a "from" or a "to" mesh. For read-consistent mappings, only the "from" mesh must be defined. For write-conservative mappings, only the "to" mesh must be defined. In both cases, the defined mesh needs to match the received mesh with api-access enabled (which is also used in the `read-data` or `write-data` tag).
+
+At its core, the API now makes use of two new API functions, called `mapAndReadData` and `mapAndWriteData`. Both functions are very similar to the analogous functions `readData` and `writeData`: the only difference is that the functions take spatial coordinates as function arguments instead of static vertexIDs stemming from the preCICE initialization. In addition to `mapAndReadData` and `mapAndWriteData`, the access region needs to be defined by "SolverOne" using `setMeshAccessRegion`. The full code example reads the following:
```
uekerman:
```suggestion
The API makes use of two new API functions, called `mapAndReadData` and `mapAndWriteData`. Both functions are very similar to the analogous functions `readData` and `writeData`: The only difference is that the functions take spatial coordinates as function arguments instead of static vertexIDs. In addition to `mapAndReadData` and `mapAndWriteData`, the access region needs to be defined by "SolverOne" using `setMeshAccessRegion`. A full code example reads as follows:
```

uekerman: There is still the ongoing discussion in https://github.com/precice/precice/issues/1843: `readDataAt` and `writeDataAt` as alternative names. Once concluded there, we need to update here.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    // Allocate and fill a 'boundingBox' to define an access region
+    // in our example, we use the unit cube.
+    // Assuming dim == 3, means that the bounding box has
+    // dim * 2 == 6 elements.
+    std::vector<double> boundingBox {
+        0, 0, 0, // minimum corner
+        1, 1, 1 // maximum corner
+    };
+
+    // Define region of interest, where we want to obtain the direct access.
```
uekerman:
```suggestion
    // Define region of interest, where we want to obtain API access.
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    // See also the API documentation of this function for further notes.
```
uekerman:
```suggestion
```
This should be clear.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    precice.setMeshAccessRegion(otherMesh, boundingBox);
+
+    // initialize preCICE as usual
+    // initializing data (requiresInitialData) before calling initialize is not possible, as we first need to exchange the
+    // meshes before being able to access the mesh
```
uekerman: What does "not possible" exactly mean? Return false? Throws an error?
What if other data (e.g. in a different coupling scheme) requires initialization?

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+    precice.initialize();
+
+    // until this point, everything was just the same as for
+    // the direct mesh access, now we enter the main time loop
+    // and use dedicated read and write functions
```
uekerman: I think we could use wider comments here. Do you break at 80 chars?

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+        // we simply select here the end of the time window
+        double dt = couplingInterface.getMaxTimeStepSize();
+
+        // The just-in-time reading function: its interface
+        // is very similar to the conventional readData
+        // function, only the vertexIDs are now dynamic
+        // coordinates:
+
+        // we need a vector, where we store the data we read
+        std::vector<double> velocityValue(1);
```
uekerman: If we use velocities and forces, we should go to vectorial data. Will otherwise be confusing.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+        // however,
+        std::vector<double> writeCoordinates({0.75, 0.2, 0.3});
+        precice.mapAndWriteData(otherMesh, "Forces", writeCoordinates, forceValue);
+
+        // finally advance to the next timestep, as usual
+        precice.advance(dt);
+    }
+```
+
+A more comprehensive description of all involved API function and their arguments is given in our [doxygen API documentation](/doxygen/main/classprecice_1_1Participant.html) (see the section on Just-in-time mapping). Many other configuration and code examples can be found in related [integration tests](https://github.com/precice/precice/tree/develop/tests/serial/just-in-time-mapping). Just-in-time mapping includes full support for [time interpolation](couple-your-code-waveform.html) and subcycling.
```
uekerman:
```suggestion
A more comprehensive description of all involved API function and their arguments is given in the [API documentation](/doxygen/main/classprecice_1_1Participant.html) (see the section on just-in-time mapping). Many other configuration and code examples can be found in related [integration tests](https://github.com/precice/precice/tree/develop/tests/serial/just-in-time-mapping). Just-in-time mapping includes full support for [time interpolation](couple-your-code-waveform.html) and subcycling.
```

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Limitations
+
+Just-in-time data mapping is currently only implemented for the mapping combinations read-consistent (`<mapping:... direction="read" constraint="consistent"/>`) and write-conservative (`<mapping:... direction="write" constraint="conservative"/>`). Furthermore, only the mapping types `<mapping:nearest-neighbor .../>`, `<mapping:rbf .../>` and `<mapping:rbf-pum-direct .../>` are implemented. The general configuration of other mapping attributes follows the usual [mapping convention](configuration-mapping.html), i.e., configuring a `rbf-pum-direct` mapping for a just-in-time mapping would look as follows
```
uekerman:
```suggestion
Just-in-time data mapping is currently only implemented for the mapping combinations read-consistent (`<mapping:... direction="read" constraint="consistent"/>`) and write-conservative (`<mapping:... direction="write" constraint="conservative"/>`). Furthermore, only the mapping types `<mapping:nearest-neighbor .../>` and `<mapping:rbf-pum-direct .../>`, and the alias `<mapping:rbf .../>` are implemented. The general configuration of other mapping attributes follows the usual [mapping convention](configuration-mapping.html), i.e., configuring a `rbf-pum-direct` mapping for a just-in-time mapping would look as follows:
```

pages/docs/couple-your-code/couple-your-code-moving-or-changing-meshes.md
```diff
@@ -13,6 +13,10 @@ Note that spatially aware mapping-schemes in preCICE work in the reference domai
 
 Examples of this method are the [perpendicular flap tutorials](tutorials-perpendicular-flap) and the [Turek-Hron FSI3 benchmark](tutorials-turek-hron-fsi3).
 
+## Just-in-time data mapping
+
+Instead of defining a static mesh during initialization, coordinates can be provided at runtime to read and write data at the provided coordinates using the so-called [just-in-time data mapping](couple-your-code-just-in-time-mapping.md). This feature was particularly developed for mesh-particle coupling, where one participant has a static mesh whereas the other participant has time-dependent coupling locations. However, it might be useful in other scenarios as well.
```
uekerman:
```suggestion
Instead of defining a static mesh during initialization, coordinates can be provided at runtime to read and write data at the provided coordinates using so-called [just-in-time data mappings](couple-your-code-just-in-time-mapping.md). This feature has particularly been developed for mesh-particle coupling, where one participant has a static mesh and the other participant has time-dependent coupling locations. However, it might be useful in other scenarios as well.
```

uekerman (changes requested):

pages/docs/couple-your-code/couple-your-code-direct-access.md
davidscn: I think this was already done. I can't find outdated information.

pages/docs/configuration/configuration-mapping.md
davidscn: Moved it in front of the 'Available methods' section and made it timeless.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: We converged to names. Updated them here as well.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: Removed the linebreaks

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: used vectorial data

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
uekerman: I assume you checked how this looks rendered?

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: "not possible" means that it is not possible by construction.

Usually, we have 

```
requiresInitialData() -> initialize() -> getMeshVertexIDsAndCoordinates()
```
where the last function gives us only the IDs we need for `writeData` inside the `requiresInitialData()`, i.e., the above is not possible at all, as we simply don't have the vertex IDs where we need them (inside the `requiresInitialData()`.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: In the initial version, yes. In the updated version it looks as follows

image

which exceeds the box size. The initial version was this one:

image

what exactly would you like?

uekerman: how do we do this in other code examples? automatic wrap would be nice.
or as a compromise sth between both of your versions: same length as the longest code length?

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
uekerman: Being more explicit could probably help here. It is still allowed to call `requiredInitialData()` as other meshes could need this.

pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md
davidscn: used now clang-format with `ColumnLimit:     87` (which is the longest code line)

image

BenjaminRodenberg: I can add my Python perspective here: In `autopep8` the default is also 80 characters. In the CI of preCICE-related Python projects I set the limit to 120. See [here](https://github.com/precice/python-bindings/blob/1cba0cffb9d595bda807660387ff09167e3dee6a/.github/workflows/check-pep8.yml#L18). Main motivation for this change was that I wanted to avoid linebreaks in preCICE API calls that show up regularly.[^1]

87 sounds to me pretty arbitrary. I would not use this value. Under https://stackoverflow.com/questions/88942/why-does-pep-8-specify-a-maximum-line-length-of-79-characters you can find some further discussion on the topic.

[^1]: e.g. [this line](https://github.com/precice/python-bindings/blob/1cba0cffb9d595bda807660387ff09167e3dee6a/examples/solverdummy/solverdummy.py#L44) has 98 characters.

uekerman: The "87" version looks decent to me. I am pretty to sure that we do this pretty arbitrarily across all code blocks in the documentation. 

@BenjaminRodenberg We rightfully have higher standards in codes than docs :)

If we really want a consistent solution, I would suggest to go ahead here with 87 and open an issue instead.

pages/docs/configuration/configuration-mapping.md
```diff
@@ -64,6 +64,10 @@ Each mapping defines a `constraint`, which defines how the data is mapped betwee
 
 For a sequential participant, any combination of `read`/`write`-`consistent/conservative` is valid. For a participant running in parallel, only `read`-`consistent` and `write`-`conservative` is possible. More details are given [further below](configuration-mapping.html#restrictions-for-parallel-participants).
 
+{% tip %}
+Instead of using static meshes as described here, you may consider using a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where the meshes are not statically defined during initialization, but change at runtime.
```
uekerman:
```suggestion
Instead of using static meshes as described here, you may consider using a [just-in-time data mapping](couple-your-code-just-in-time-mapping.html), where one of both meshes can change at runtime.
```

uekerman (approved):

r-walloner:
Hi, I noticed that the code example in `pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md` defines the bounding box as:
```c++
std::vector<double> boundingBox{
    0, 0, 0, // minimum corner
    1, 1, 1  // maximum corner
};
```
So in the order `x_min, y_min, z_min, x_max, y_max, z_max`.
But in the test cases (e.g., `AccessReceivedMeshAndMapping`) a different order is used: `x_min, x_max, y_min, y_max`.

In fact, when using the order form the code example, I get
```
(0) 11:45:07 [partition::ReceivedPartition]:106 in compute: ERROR: You are running this participant in serial mode and the bounding box on mesh "Fluid-Mesh", is empty. Did you call setMeshAccessRegion with valid data?
```
This is to be expected, since the bounding box defined this way has length zero in at least one dimension.

davidscn:
Ah that's a copy mistake, we have the same problem in the [direct access docs](https://precice.org/couple-your-code-direct-access.html). Thanks for pointing this out. The doxygen documentation (and of course also the tests) are correct.

@r-walloner would you like to open a PR to fix it or should I take care of this?

r-walloner:
@davidscn I just opened #501

PR #498, opened 2025-02-26
Extend docs on compositional coupling with deadlock example

uekerman:
I promised in https://github.com/precice/precice/issues/1907 to add a docs section on deadlocks in compositional coupling. I tried to add the simplest possible example.

The mermaid graph will probably not work on the website. Would this be easy to add? If not, I would add the comment in a comment and add the svg instead.

```mermaid
graph LR;
    A -- first to second --> B;
    B -- first to second --> C;
    C -- first to second --> A;
```

@fsimonis @carme-hp Enough information covered and would you prefer to see more examples and problems?

@orlandoisepic Could be a potential case for the checker a well.


pages/docs/configuration/configuration-coupling-multi.md
```diff
@@ -11,28 +11,74 @@ preCICE allows us to combine multiple coupling schemes by only specifying them o
...
 </coupling-scheme:parallel-explicit>
 <coupling-scheme:parallel-explicit>
-  <participants first="MySolver1" second="MySolver3"/>
+  <participants first="C" second="A"/>
   ...
 </coupling-scheme:parallel-explicit>
 ```
 
-For this example, all three participants are executed in parallel to one another, whereas `MySolver1` exchanges data with `MySolver2` and `MySolver3`, but not the latter two with each other. To also get an interaction between `MySolver2` and `MySolver3`, add a third coupling scheme.
+For this example, all three participants are executed in parallel to one another, whereas `A` exchanges data with `B` and `C`, but not the latter two with each other.
```
carme-hp:
```suggestion
For this example, all three participants are executed in parallel to one another, whereas `A` exchanges data with `B` and `C`, but `B` and `C` do not exchange data with each other.
```
I know it says the same, but I think it's better to make it even more clear. 

Maybe it will be nice to add a comment that "first" and "second" doesn't matter?

carme-hp (approved):
I think that the explanation for the deadlock and its solution is clear:)

fsimonis:
> The mermaid graph will probably not work on the website. Would this be easy to add? If not, I would add the comment in a comment and add the svg instead.

Not difficult. But wouldn't it be neater to see an actual config visualization? You mentioned that you have a minimal example, right?

orlandoisepic:
This error can only occur with serial coupling schemes? And is it possible to occur with two or more participants?

uekerman:
@fsimonis 

> But wouldn't it be neater to see an actual config visualization? You mentioned that you have a minimal example, right?

https://github.com/precice/precice/blob/develop/tests/serial/circular/Explicit.xml and swap participants in line 65. In the config visualization the coupling scheme is separated from the exchanges, right?

uekerman:
@orlandoisepic 

> This error can only occur with serial coupling schemes? 

yes

> And is it possible to occur with two or more participants?

You need at least three participant to get such a circular dependency. For two participants, you can only define one coupling scheme.

pages/docs/configuration/configuration-coupling-multi.md
```diff
@@ -11,28 +11,74 @@ preCICE allows us to combine multiple coupling schemes by only specifying them o
...
+```
+
+The mesh over which the data is communicated plays no role. We get a circular dependency:
+
+```mermaid
+graph LR;
+    A -- first to second --> B;
+    B -- first to second --> C;
+    C -- first to second --> A;
+```
```
MakisH: Does not work at the moment: https://github.com/precice/precice.github.io/issues/552

pages/docs/configuration/configuration-coupling-multi.md
```diff
@@ -11,28 +11,74 @@ preCICE allows us to combine multiple coupling schemes by only specifying them o
...
+  </coupling-scheme:serial-explicit>
+```
+
+The mesh over which the data is communicated plays no role. We get a circular dependency:
+
+```mermaid
+graph LR;
+    A -- first to second --> B;
+    B -- first to second --> C;
+    C -- first to second --> A;
```
uekerman:
```suggestion
```txt
A -- first to second --> B;
B -- first to second --> C;
C -- first to second --> A;
```

uekerman:
Merging here. Let's add the mermaid graph eventually, see #552

PR #499, opened 2025-02-28
Add link to Plausible analytics

MakisH:
@fsimonis mentioned a few times lately how cool our open analytics are, so I thought of adding a link to make this more discoverable.

I first tried adding a link in the footer, but that is already getting busy, and is now nicely aligned.


fsimonis (approved):
Cool!

PR #500, opened 2025-03-04
Remove option `PYTHON_EXECUTABLE`

BenjaminRodenberg:
This option does not exist anymore


fsimonis:
It refers to the PYTHON_EXECUTABLE provided by [CMake FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html).

PR #501, opened 2025-03-07
Fix bounding box definitions in code examples

r-walloner:
I noticed that the code example in `pages/docs/couple-your-code/couple-your-code-just-in-time-mapping.md` defines the bounding box as:
```c++
std::vector<double> boundingBox{
    0, 0, 0, // minimum corner
    1, 1, 1  // maximum corner
};
```
So in the order `x_min, y_min, z_min, x_max, y_max, z_max`.
But in the test cases (e.g., `AccessReceivedMeshAndMapping`) a different order is used: `x_min, x_max, y_min, y_max`.

In fact, when using the order form the code example, I get
```
(0) 11:45:07 [partition::ReceivedPartition]:106 in compute: ERROR: You are running this participant in serial mode and the bounding box on mesh "Fluid-Mesh", is empty. Did you call setMeshAccessRegion with valid data?
```
This is to be expected, since the bounding box defined this way has length zero in at least one dimension.


davidscn (approved):

davidscn:
Thanks!

PR #510, opened 2025-03-25
Update installation instructions for building boost from source

BenjaminRodenberg:
I'm currently building boost and preCICE from source and I have the impression that the instructions here are outdated. I'm configuring the build process of preCICE using `ccmake`. Here, the variable `Boost_DIR` has to be set to `<prefix>/lib/cmake/Boost-1.87.0`. I thought that only the prefix would be required here.


pages/docs/installation/installation-source-dependencies.md
```diff
@@ -151,21 +151,13 @@ The following header-only Boost libraries are also needed: 'asio', `vmd`, `geome
    The directory you chose as prefix now contains libraries in `<prefix>/lib` and the all the Boost headers in `<prefix>/include`.
    You may now safely remove the boost directory from step 1.
 
-4. If you selected `/usr/local` as prefix, update the dynamic linker's run-time bindings:
-
+4. If you selected `/usr/local` as prefix, you do not have to do anything else. If you used a custom `<prefix>` you have to add the following line to your `.bashrc`:
```
MakisH: This depends on the system. Ubuntu ignores `/usr/local` by default. Maybe you have added it to your respective paths? Or is it just that newer Boost changed something?

BenjaminRodenberg: I was working on CoolMUC and building Boost from source. Maybe the point with *"...update the dynamic linker's run-time bindings..."* is still relevant if you use `/usr/local` as a prefix (On Ubuntu I use the system Boost). We can also revert this bit and only keep the part about `CMAKE_PREFIX_PATH`.

fsimonis: > Ubuntu ignores /usr/local by default.

I just checked in a fresh Ubuntu 24.04 docker container. `/usr/local` is in the linker, dynamic lib and include search directories.

```
g++ --verbose
ld --verbose
ldconfig --verbose
```

MakisH:
```suggestion
4. If you selected `/usr/local` as prefix, you probably do not have to do anything else (depends on your system). If you used a custom `<prefix>`, you have to add the following line to your `.bashrc`:
```

MakisH (approved):

fsimonis (approved):

PR #519, opened 2025-04-17
Add travel awards info to workshop 25

uekerman:


pages/community/precice-workshop-2025.md
```diff
@@ -108,6 +108,10 @@ Preliminary schedule:
 
 A few weeks after the workshop, we will also have an online follow-up user support session (TBA).
 
+## Travel awards
+
+We have up to 5 travel awards available for those, who cannot join otherwise. Each award includes co-funding of 500€ for travel and accomodation and a waiver for the registration fee. If you are interested in this support, please *briefly* explain your situation by mailing to `precice` at `hsu-hh.de`.
```
MakisH: Why the emphasis on "briefly"? Sounds strange to me. Also, a typo fix.
```suggestion
We have up to 5 travel awards available for those who cannot join otherwise. Each award includes co-funding of 500€ for travel and accomodation and a waiver for the registration fee. If you are interested in this support, please briefly explain your situation by mailing to `precice` at `hsu-hh.de`.
```

MakisH (approved):
I would have put it next to Registration and Accommodation, since it is cost- and pre-conference related. The "help us advertise" is at the bottom mainly because it is a "by the way" that might not be interesting for many people.

But good anyway!

uekerman:
> I would have put it next to Registration and Accommodation, since it is cost- and pre-conference related.

Let's move when we add more program details.

PR #520, opened 2025-04-18
Add section on parallelism to terminology

uekerman:
Feedback from @orlandoisepic


MakisH (approved):
Very useful point, I think always needed it but never really had a clear mention in an easy-to-find place.

Issue #522, opened 2025-04-23
Landing page: Citation details outdated/broken


MakisH:
In the landing page:

- We still have the v1 paper in the counter
- The URL does not work
- The fallback citation numbers is used (222 instead of currently 397)

Maybe it's time to replace the counter with the v2 paper, or add both somehow. We should also fix the link issue.

kss682:
Continuing the discussion from #524, 

It seems like google scholar technically do not allow scraping and since the github action servers might be using IP ranges which could be blacklisted by google. 

One option would be use to add `User-Agent` header or we could use public APIs like [semantic-api](https://api.semanticscholar.org/api-docs/) ( they have a limit of 1000 requests per second among all unauthenticated users.)

Eg: 
```
https://api.semanticscholar.org/graph/v1/paper/10.12688/openreseurope.14445.1?fields=referenceCount
```

![Image](https://github.com/user-attachments/assets/9bdd1abc-cd36-4458-bfed-b0208431199e)

MakisH:
I am not sure which solution would be the easiest to implement, but both sound sustainable to me if they do the job. Does the Semantic Scholar API report the same numbers as Google Scholar, or will we confuse people if we link to the one and get numbers from the other?

Alternative: Do you think we could (mis)use this GitHub Action to get what we need?

https://github.com/marketplace/actions/google-scholar-fetcher

kss682:
I tried with the Semantic API and the citation count is different from what we have from google scholar. 

![Image](https://github.com/user-attachments/assets/887e7d08-a70f-4b7f-8c4e-0081271975ba)
Also is this what we want for the citation layout ?

MakisH:
I would expect all buttons to be next to each other, potentially with shorter text or some grouping.

kss682:
![Image](https://github.com/user-attachments/assets/d45cf7af-1827-4455-98c8-cdb835cace79)

Reverted back to google scholar due to citation mismatch with semantic API, fetching these results using `user-agent`. 
I have reduced the spacing but not sure of how we can shorten the text.

uekerman:
"preCICE" is clear from context. Maybe simply "v2 citations (134)"?

Or:
- Cited by 134
- v1 cited by 399

kss682:
![Image](https://github.com/user-attachments/assets/69f2edb6-886c-48ae-86d7-bccb74d8c809)

Do you feel the whitespace below needs to be reduced ?

MakisH:
[...]

PR #524, opened 2025-04-30
Fix citation details

kss682:
Fix citation details (#522): Update citation ID of google scholar to preCICE v2 paper.

[google scholar](https://scholar.google.com/scholar?hl=en&cites=17974677460269868025)

Screenshot from 2025-04-30 22-55-31

Screenshot from 2025-04-30 23-00-40

### DIFF
--- a/_plugins/googlescholar.rb
+++ b/_plugins/googlescholar.rb
@@ -6,8 +6,8 @@ require 'nokogiri'
 
 module Jekyll
   class ScholarStats < Generator
-    # ID for Bungartz et al. paper
-    CITATION_ID = '5053469347483527186'.freeze
+    # ID for PreCICE v2 paper
+    CITATION_ID = '17974677460269868025'.freeze
     SCHOLAR_URL = 'http://scholar.google.com/scholar?hl=en&cites='.freeze
     def generate(site)
       begin
--- a/pages/index.html
+++ b/pages/index.html
@@ -311,7 +311,7 @@ layout: landing_page
             <a href="assets/precice-v2.bib">Download BibTeX &nbsp;<i class="fas fa-download"></i></a>
           </div>
         </div>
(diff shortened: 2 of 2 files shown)


uekerman (approved):

uekerman:
I merged too quickly :) Seems that the link is not yet working.

kss682:
For some reason, the `CITATION_ID` is not getting set . 
Local (Docker setup) 
image
Prod
image

kss682:
@uekerman 
The build logs in github actions shows that the API call is blocked.  
image
Maybe we can set user-agent in the header when making the call ?

MakisH:
@kss682 let's continue the discussion in https://github.com/precice/precice.github.io/issues/522 The issue might be that Scholar blocks GitHub due to too many requests. But this is just a wild guess.

In the next PR, let's fix the fallback number as well: https://github.com/precice/precice.github.io/blob/0a28be152281bbd095dc5ded0e992d5dc7dc5139/_config.yml#L322

_plugins/googlescholar.rb
```diff
@@ -6,8 +6,8 @@
 
 module Jekyll
   class ScholarStats < Generator
-    # ID for Bungartz et al. paper
-    CITATION_ID = '5053469347483527186'.freeze
+    # ID for PreCICE v2 paper
```
MakisH, line 9: Also important for consistency: `preCICE` instead of `PreCICE`. We always use `preCICE` where capitalization is expected (even in the beginning of a sentence), `precice` where not (repositories, code, etc).

Read more: https://precice.org/dev-docs-naming.html#the-precice-name

MakisH:
I took too long to review, but something also for the second PR: I think that we should maintain a link/mention to the v1 paper in this box, for being clearer with the history, and since the v1 paper already has quite some citations that are also useful to look at as a list of titles.

The easiest way to implement this would be as an additional button:

- Citations of preCICE v2 paper (XYZ)
- Citations of preCICE v1 paper (XYZ)
- Literature guide

edit: And the most important thing I forgot: thank you! :hugs:

PR #527, opened 2025-05-16
Fix citation details

### LINKED ISSUE #522: Landing page: Citation details outdated/broken

kss682:
PR fixes #522 
- Add user-agent to google scholar request
- Align citation buttons
- Update default values in config
- Add citation id in config

image

### DIFF
--- a/.github/workflows/update-citation.yml
+++ b/.github/workflows/update-citation.yml
@@ -1,0 +1,48 @@
+name: Update Record
+
+on:
+  # Create a scheduled task, in this example we run it at the first day of every month.
+  schedule:
+    - cron: "0 0 */14 * *"
+  # Enable manually executing.
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    # Fetch record with Google Scholar ID
+    - name: Get record with token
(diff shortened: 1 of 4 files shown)


kss682:
Unfortunately, the citation request fails with user-agent too. 
image

MakisH:
Are there any usage/request limits?

kss682:
It does have a rate limit which is what happens when we run it locally multiple times, that usually gives a 429. 
But the request from actions are forbidden.

MakisH:
In #651, I updated the hard-coded data, and added a URL to the v1 paper. I created the following conflicts that should be easy to fix, depending on the future direction of this PR:

- In the `_config.yml`, I renamed the variables pointing to the two papers. I did not use IDs, since the v2 paper already appears with two IDs on Scholar. I used URLs, instead.
- In the `pages/index.html`, I removed the if/else, since the automatic solution was already broken. Should be trivial to add again, once there is a variable for v1.

MakisH:
Closing as automatically accessing Google Scholar seems to be a dead end. See the original issue: https://github.com/precice/precice.github.io/issues/522#issuecomment-4485879061

Issue #531, opened 2025-06-20
Mention student tools on website


Logende:
* generate
* checker

PR #532, opened 2025-06-20
Mention precice/config-check on website

Logende:
Not fully convinced whether this is the best place to put the reference to the student project, but also could not think of a better place. 
In the near to mid future, we anyways want to re-structure the tooling some more and combine both precice checkers to be executed with the same command (If both are installed, otherwise only the installed one is executed and a message printed that another check exists too).

### DIFF
--- a/pages/docs/tooling/tooling-builtin.md
+++ b/pages/docs/tooling/tooling-builtin.md
@@ -73,7 +73,7 @@ The `check` runs the preCICE configuration parsing and checking logic on the giv
 This will find the majority of the configuration mistakes without having to start a simulation.
 These checks include wrong tags and attribute values, and more elaborate naming checks.
 More advanced logic, such as checks if all necessary data are exchanged in a coupling scheme, are not covered.
-For more logical checks, see [here](https://github.com/precice/config-check).
+For more logical checks, see the [preCICE config checker](https://github.com/precice/config-check).
 
 The basic usage is to check a configuration file:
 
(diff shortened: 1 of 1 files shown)


Logende:
This is connected to https://github.com/precice/precice.github.io/issues/531#issue-3162620550

pages/docs/tooling/tooling-builtin.md
```diff
@@ -73,6 +73,7 @@ The `check` runs the preCICE configuration parsing and checking logic on the giv
 This will find the majority of the configuration mistakes without having to start a simulation.
 These checks include wrong tags and attribute values, and more elaborate naming checks.
 More advanced logic, such as checks if all necessary data are exchanged in a coupling scheme, are not covered.
+For more logical checks, see [here](https://github.com/precice/config-check).
```
MakisH: Links should be descriptive:

```suggestion
For more logical checks, see the [preCICE config checker](https://github.com/precice/config-check).
```

(see the page for [content guidelines](https://precice.org/docs-meta-content-guidelines.html))

MakisH (commented):
The place is right, but you could also add it on https://precice.org/tooling-overview.html

This page will anyway need an extensive rewriting at some point, following https://github.com/precice/precice/issues/2313.

Logende:
> The place is right, but you could also add it on https://precice.org/tooling-overview.html
> 
> This page will anyway need an extensive rewriting at some point, following [precice/precice#2313](https://github.com/precice/precice/issues/2313).

I believe putting it on https://precice.org/tooling-overview.html could be confusing for users, as they might be familiar to precice check, but the other precice check one. In the near to mid future, we want to put both checks behind the same command, then it will also make sense to explain both within one documentation page. And until then, it should be sufficient to just mention the existence of the other checker, but in a way that does not create any confusion to the user. 

@MakisH do you agree? Then I would merge this PR.

MakisH (approved):
Ah, indeed, this page is already linked to from the tooling-overview, so this is fine as-is now. Good to go!

Issue #540, opened 2025-06-27
Docs may be outdated


prathamgit:
I was reading the precice documentation and came across this
https://precice.org/couple-your-code-distributed-meshes.html#:~:text=In%20this%20approach%20we,and%20precice%3A%3AwriteBlockScalarData(...))

The documentation needs to be updated because, readBlockScalarData() is no longer valid since precice 3.x versions.

MakisH:
Related: #542 should catch many similar issues in the future.

fsimonis:
Closed via #542 

@prathamgit thanks for pointing it out!

PR #542, opened 2025-07-03
Add hook to check for old preCICE API and config

fsimonis:
This PR adds a database of old API, a script that checks files against matches, and a pre-commit hook that uses the script.


tools/oldAPI.txt
```diff
@@ -0,0 +1,62 @@
+# port 1-2
+readBlockScalarData
+isTimestepComplete
+fulfilledAction
+
+# port 2-3
+precice::SolverInterface
```
MakisH: What is flagged if we remove the `precice::`? `SolverInterface` should catch more false negatives.

```suggestion
SolverInterface
```

tools/check-old-api.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def oldAPI():
+    return [entry
+            for entry in (pathlib.Path(__file__).parent / "oldAPI.txt").read_text().splitlines()
+            if entry and not entry.startswith("#")]
+
+
+def fileFilter(file: pathlib.Path):
+    for s in ["porting", "CHANGELOG", "README"]:
```
MakisH: Why the `README`?

```suggestion
    for s in ["porting", "CHANGELOG"]:
```

tools/oldAPI.txt
```diff
@@ -0,0 +1,62 @@
+# port 1-2
+readBlockScalarData
```
MakisH: Duplicate with v2 (and missing the `writeBlockScalarData` anyway):

```suggestion
```

tools/check-old-api.py
```diff
@@ -0,0 +1,43 @@
...
+from sys import exit, stderr
+
+
+def oldAPI():
+    return [entry
+            for entry in (pathlib.Path(__file__).parent / "oldAPI.txt").read_text().splitlines()
+            if entry and not entry.startswith("#")]
+
+
+def fileFilter(file: pathlib.Path):
```
MakisH: In some pages, we sometimes write "before vx.y, this used to be `xyz`".
This hook does not seem to flag any such instances right now, but this might be something to consider in the filtering.

We could have a `compatibility` alert, and then this filter could ignore content in such alerts.

tools/oldAPI.txt
```diff
@@ -0,0 +1,62 @@
...
+isTimestepComplete
+fulfilledAction
```
MakisH: Do we also want to add header files? `constants.hpp` could be an option (but too generic).

tools/oldAPI.txt
```diff
@@ -0,0 +1,62 @@
...
+preallocation="tree"
+preallocation="compute"
+preallocation="estimate"
+preallocation="save"
+preallocation="off"
+use-qr-decomposition="true"
+extraplation-order
+min-iteration-convergence-measure
+acceleration:broyden
+sync-mode="on"
```
MakisH: Are the config options from v1 considered?
https://precice.org/couple-your-code-porting-v1-2.html#precice-configuration-file

I see the `max-timesteps` and the `post-processing` missing, for example.

.pre-commit-config.yaml
```diff
@@ -16,3 +16,11 @@ repos:
     rev: v0.10.0
     hooks:
       - id: shellcheck
+  - repo: local
+    hooks:
+      - id: check-api
+        name: check old API use
+        entry: tools/check-old-api.py
```
MakisH: Maybe rename all these to include `precice`? It took me a bit to understand the context when I first opened this PR.

```suggestion
      - id: check-precice-api
        name: check old preCICE API use
        entry: tools/check-old-precice-api.py
```

MakisH (commented):
Excellent idea!

tools/oldAPI.txt
fsimonis: Ah no. Haven't done those

tools/oldAPI.txt
fsimonis: We migrated everything to `precice/precice` I don't think we need those.

tools/check-old-api.py
fsimonis: This is what the porting guides are for. There are some version mentions, but nothing triggers this check.

tools/oldAPI.txt
fsimonis: Added

tools/check-old-api.py
MakisH: I see the porting guides and the compatibility notes as serving different use cases. But since the current notes don't trigger this check, that's all fine at the moment.

MakisH:
The check looks good. Would you like me to fix the content in this PR?

tools/check-old-api.py
fsimonis: > nothing triggers this check.

There are no such compatibility notes.

MakisH: You mean there are no such compatibility notes anymore. This PR removes one:

https://github.com/precice/precice.github.io/pull/542/files#diff-6e985446a5a752c484e9a2111ba9dc0f351b276a3af61ef4b07e237f85480ffa

pages/docs/installation/installation-linking.md
```diff
@@ -99,11 +99,11 @@ Now you can use `pkg-config` to extract the necessary flags.
 There are two reasons you may be getting this error:
 
 * `pkg-config` could not find a `libprecice.pc` file (keep reading)
-* you are including the file as `SolverInterface.hpp` and not as `precice/SolverInterface.hpp`
+* you are including the file as `Participant.hpp` and not as `precice/Participant.hpp`
```
MakisH: The user-facing header is `precice.hpp`:

```suggestion
* you are including the file as `precice.hpp` and not as `precice/precice.hpp`
```

pages/docs/installation/installation-linking.md
```diff
@@ -99,11 +99,11 @@ Now you can use `pkg-config` to extract the necessary flags.
...
 
 ### `preciceC.h` cannot be found
 
-If you are using the C bindings, please note that they are now installed in `[prefix]/include/precice/`, alongside the C++ headers. If your code includes e.g. `precice/bindings/c/SolverInterfaceC.h`, please update this to `precice/SolverInterfaceC.h`.
+If you are using the C bindings, please note that they are now installed in `[prefix]/include/precice/`, alongside the C++ headers. If your code includes e.g. `precice/bindings/c/ParticipantC.h`, please update this to `precice/preciceC.h`.
```
MakisH: If we are touching this line, then we should write it with today's perspective:

```suggestion
Similarly to `precice.hpp` not being found, make sure you include `precice/preciceC.h`, not just `preciceC.h`.
```

MakisH (approved):
Looks good, and also closes #540 (the rest of the page seems to be up-to-date).

Just another small change.

pages/docs/installation/installation-linking.md
fsimonis: Done

PR #546, opened 2025-07-08
Add note on spack installations

davidscn:
... which was direct feedback from Thomas on the spack installation.

### DIFF
--- a/pages/docs/installation/installation-spack.md
+++ b/pages/docs/installation/installation-spack.md
@@ -84,6 +84,15 @@ If you want to uninstall preCICE, `spack uninstall precice` or delete the comple
 
 🎉
 
+Note that loading preCICE using `spack` only affects your current shell. If you start a new shell (e.g. open a new terminal) you have to load preCICE again to be able to use it. To automatically load preCICE in every new shell environment, you may want to add
+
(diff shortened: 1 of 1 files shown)


pages/docs/installation/installation-spack.md
```diff
@@ -84,6 +84,15 @@ If you want to uninstall preCICE, `spack uninstall precice` or delete the comple
 
 🎉
 
+Note that loading preCICE using `spack` only affects your current shell. If you start a new shell (e.g. open a new terminal) you have to load preCICE again to be able to use it. To automatically load preCICE in every new shell environment, you may want to add
+
+```bash
+source ~/spack/share/spack/setup-env.sh
+spack load precice@
```
MakisH: Why the `@`? It should work without

```suggestion
spack load precice
```

MakisH (approved):
Good point.

We have the `setup-env.sh` as a comment earlier https://precice.org/installation-spack.html#setting-up-spack

Does not harm to repeat.

pages/docs/installation/installation-spack.md
davidscn: Copy-paste error issue

PR #547, opened 2025-07-08
Remove yaml-cpp as a dependency of the CalculiX adapter

IshaanDesai:
In https://github.com/precice/calculix-adapter/pull/143 the CalculiX adapter was changed to use the header-only YAML. Hence the information about yaml-cpp as a dependency needs to be removed.


MakisH (approved):
Looks good, provided that the update is released.
I will review the pending PRs, and we can then release everything together.

fsimonis:
General question: how do we handle documentation of older versions?

The YAML-cpp documentation is still relevant for everything prior to the latest/upcoming version.

MakisH:
> General question: how do we handle documentation of older versions?
> 
> The YAML-cpp documentation is still relevant for everything prior to the latest/upcoming version.

That's why I suggested a `{% compatibility %}` alert in https://github.com/precice/precice.github.io/pull/542#discussion_r2182570860

Assuming we don't want that for some reason, the current strategy is only relying on the exported PDF: https://precice.org/fundamentals-previous-versions.html

fsimonis:
> That's why I suggested a {% compatibility %} alert in https://github.com/precice/precice.github.io/pull/542#discussion_r2182570860

Explaining how to install yaml cpp and how to set everything up is enough content for a subsection. It's too much for an alert/note.

> Assuming we don't want that for some reason, the current strategy is only relying on the exported PDF: https://precice.org/fundamentals-previous-versions.html

So far, this versioning strategy has been regarding preCICE configuration etc. We never nailed down a strategy on how to handle adapter/binding/tooling versions. I also don't have a good idea on how to handle that.

MakisH:
Let's continue in a separate issue for the general strategy: https://github.com/precice/precice.github.io/issues/548

MakisH:
I will implement the compatibility label for #548 and add one here.

fsimonis (changes requested):
This information needs to be kept around as long as it is relevant.

pages/docs/adapters/calculix/adapter-calculix-supermuc.md
MakisH, line 1: This system is anyway decommissioned, and the page is extremely outdated, so I would rather remove it.

MakisH: @fsimonis the bottom of that page states that one can use CMake to build the adapter. Is this a relic of your past endeavors? This section should probably be removed.

Not sure how much of the rest of the content is relevant to anyone.

MakisH (approved):
I updated all pages to keep the information but add version context around it. Looks ready to merge for me.

We could additionally discuss on whether the SuperMUC page still makes sense to keep around, but I have added the context there as well.

fsimonis (approved):

PR #553, opened 2025-07-18
Literature guide: Add links to external lists

### LINKED ISSUE #12: Add a page for student projects
We could create a page similar to our [Publications](https://www.precice.org/publications.html), move there the student projects that have been completed, and a call for applications to the different groups.

MakisH:
As discussed in the coding days.

Closes https://github.com/precice/precice.github.io/issues/12 for now, though I still think that an automatic solution to have (again) such content listed in a central place would be nice.


uekerman (approved):
Content looks good, thanks!

Location: No strong preference, but I would have probably put it to the bottom of the page as a separate subsection. I am fine with the current location as well.

> I still think that an automatic solution to have (again) such content listed in a central place would be nice.

Did we really ever have this? Can't remember, but maybe that's me :see_no_evil:

MakisH:
My motivation for having it at the top is that it is a short, "by the way" thing that could be more easily discoverable. But looking at the whole thing again, it does make sense to have it at the bottom, like in the Quickstart. I will move.

We did use to have a list of publications, which also included student projects:

https://web.archive.org/web/20190904131130/https://www.precice.org/publications/

PR #556, opened 2025-07-18
Add a compatibility notes alert

MakisH:
We already have the `version` alert: https://precice.org/docs-meta-cheatsheet.html#version-information, which I only re-discovered after starting to work on this feature.

This is connected to the preCICE version (and checks whether this is a new or old release), but it can also be used without a version argument.

I am not sure if this new compatibility note is needed on top, but we could use it for more than the version-specific incompatibilities, and it would let the `version` alert evolve independently.

I have not yet tested if this works as expected.


MakisH:
Not needed, using the `{% version %}` without arguments, instead.

PR #567, opened 2025-08-05
Rewrite performance analysis

fsimonis:
This PR updates the performance analysis section to the new tools.

Open question is if we should recommend using the preCICE CLI directly `precice-cli profiling X` or using the profiling package `precice-profiling-X`

### DIFF
--- a/pages/docs/tooling/tooling-performance-analysis.md
+++ b/pages/docs/tooling/tooling-performance-analysis.md
@@ -76,6 +76,7 @@ This feature is new in preCICE version 3.3.0.
 To profile the full API including mesh definition and data access functions, you can use the `api` mode.
 
 This can be especially useful to understand the cost of just-in-time mappings or when dealing with large meshes in general.
+Be aware that this may lead to large profiling files especially when using the API in an inner loop.
 
 ## User-defined events
 
--- a/pages/docs/tooling/tooling-performance-analysis.md
+++ b/pages/docs/tooling/tooling-performance-analysis.md
@@ -175,9 +175,10 @@ The rest of the section will go through the process of analyzing participants st
 
 ### Merging Event Files
 
-{% note %}
-The file formats changed in preCICE version 3.2 from JSON to a text-based format `.txt` for rank files and a sqlite3 database `profiling.db` for profiling data.
-{% endnote %}
+{% version 3.3.0 %}
+The file formats changed from JSON to a text-based format `.txt` for rank files and a sqlite3 database `profiling.db` for profiling data.
+The `merge` tool supports both formats.
+{% endversion %}
 
 After the simulation completes, you can find `precice-profiling` folders in the configured location, defaulting to the working directory of each participant.
 An example could look like this:
--- a/pages/docs/tooling/tooling-performance-analysis.md
+++ b/pages/docs/tooling/tooling-performance-analysis.md
@@ -19,23 +19,27 @@ You can then use various tools to analyze, visualize, and export the profiling d
 
 ## Tools
 
-The tools for processing profiling data can be found in various locations:
+The tools for processing profiling data can be found in two main locations:
 
 1. Up to version 3.2, the preCICE library installation provides a `precice-profiling` script. For full functionality, some additional dependencies must be installed separately. This script has been deprecated after version 3.2, reduced to merge support and will be removed in version 4.0.
 2. The PyPi package `precice-profiling` installs the profiling tools as separate executables. It has the advantage of installing all necessary dependencies along it. The names of installed tools start with `precice-profiling-`.
-3. The PyPi package `precice-cli` installs the full suite of preCICE tools including profiling tools use `precice-cli profiling` to call them.
 
 {% note %}
 We recommend using the `precice-profiling` PyPi package for scripting and CI.
-Use the `precice-cli` PyPi package for ease of installation and normal usage.
+Use the profiling tools via the `precice-cli` PyPi package for ease of installation and normal usage.
 {% endnote %}
 
 Install the tools using pipx:
 
-```bash
-pipx install precice-cli 
-# or
-pipx install precice-profiling
+```console
+$ pipx install precice-profiling 
+  installed package precice-profiling 2.0.1, installed using Python 3.13.3
+  These apps are now globally available
+    - precice-profiling-analyze
+    - precice-profiling-export
+    - precice-profiling-histogram
+    - precice-profiling-merge
+    - precice-profiling-trace
 ```
 
 The most important tools include:
--- a/pages/docs/tooling/tooling-performance-analysis.md
+++ b/pages/docs/tooling/tooling-performance-analysis.md
@@ -1,7 +1,7 @@
 ---
 title: Performance analysis
 permalink: tooling-performance-analysis.html
-keywords: tooling, json, performance, events, analysis, cpu time, hpc, profiling
+keywords: tooling, json, db, sqlite, txt, performance, events, analysis, cpu time, hpc, profiling
 summary: "preCICE comes with an internal performance analysis framework"
 ---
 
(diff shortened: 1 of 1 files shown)


pages/docs/tooling/tooling-performance-analysis.md
```diff
@@ -12,10 +12,38 @@ Existing external frameworks cannot cope with the complexity of the multi-execut
...
+The tools for processing profiling data can be found in various locations:
+
+1. Up to version 3.2, the preCICE library installation provides a `precice-profiling` script. For full functionality, some additional dependencies must be installed separately. This script has been deprecated after version 3.2, reduced to merge support and will be removed in version 4.0.
+2. The PyPi package `precice-profiling` installs the profiling tools as separate executables. It has the advantage of installing all necessary dependencies along it. The names of installed tools start with `precice-profiling-`.
+3. The PyPi package `precice-cli` installs the full suite of preCICE tools including profiling tools use `precice-cli profiling` to call them.
+
+{% note %}
+We recommend using the `precice-profiling` PyPi package for scripting and CI.
+Use the `precice-cli` PyPi package for ease of installation and normal usage.
+{% endnote %}
```
davidscn: Having the two installation options is somewhat confusing and doesnt seem to deliver any benefits. I guess the cli anyway installs the profiling tools as subpackage. So, maybe just drop the cli tool, this section is really about the profiling tool, not the cli. If the connection between both is relevant, you could add a note saying that "profiling might already be installed as part of the cli" or similar.

fsimonis: The problem is that pipx only exposes the defined scripts of the installed package, meaning `pipx install precice-cli` installs `precice-profiling` as a dependency, but doesn't create links for its scripts (such as `precice-profiling-merge`).

> So, maybe just drop the cli tool, this section is really about the profiling tool, not the cli.

Given that we wanted another page for the cli, this looks like the best option.

fsimonis: Done

davidscn: Ok but then using the CLI for the installation anyway leads to a different command line interface?! Maybe we could add a script for creating the links?!

davidscn:
> Open question is if we should recommend using the preCICE CLI directly precice-cli profiling X or using the profiling package precice-profiling-X

Ideally, both tools would exhibit the same CLI. Why not keeping the current interface? I like it. Having precice-cli as prefix seems a bit annoying. precice-profiling-X could have the advantage of auto-completion

pages/docs/tooling/tooling-performance-analysis.md
fsimonis: The CLI provides `precice-cli profiling merge`, while the `precice-profiling` repo provides `precice-profiling-merge`. Both use the same argument parser and same runner. There is essentially no difference on their functionality.

The pipx output makes my point clearer:

```console
$ pipx install precice-cli  
  installed package precice-cli 0.2.0, installed using Python 3.13.3
  These apps are now globally available
    - precice-cli
done! ✨ 🌟 ✨
$ pipx install precice-profiling
  installed package precice-profiling 2.0.1, installed using Python 3.13.3
  These apps are now globally available
    - precice-profiling-analyze
    - precice-profiling-export
    - precice-profiling-histogram
    - precice-profiling-merge
    - precice-profiling-trace
done! ✨ 🌟 ✨
```

davidscn: I understand your point, but the ambiguity feels like a guarantee for confusion

pages/docs/tooling/tooling-performance-analysis.md
```diff
@@ -12,10 +12,42 @@ Existing external frameworks cannot cope with the complexity of the multi-execut
...
+## Tools
+
+The tools for processing profiling data can be found in two main locations:
+
+1. Up to version 3.2, the preCICE library installation provides a `precice-profiling` script. For full functionality, some additional dependencies must be installed separately. This script has been deprecated after version 3.2, reduced to merge support and will be removed in version 4.0.
+2. The PyPi package `precice-profiling` installs the profiling tools as separate executables. It has the advantage of installing all necessary dependencies along it. The names of installed tools start with `precice-profiling-`.
+
+{% note %}
+We recommend using the `precice-profiling` PyPi package for scripting and CI.
+Use the profiling tools via the `precice-cli` PyPi package for ease of installation and normal usage.
```
MakisH: But why? Is it somehow better, or just more stable at the moment?

Reading this without explanation, I would feel a bit confused and I would end up avoiding using precice-cli altogether (and try to use all the individual tools) or not knowing when to use what.

Reminds me of the `apt` and `apt-get`. I am not sure to what extent people are happy about that.

And if we see this as layers (precice-cli at the top, calls precice-profiling), it sounds like we are asking the user to interact with an internal layer. But both layers are above the ground anyway and accessible to users, so this is also fine.

pages/docs/tooling/tooling-performance-analysis.md
```diff
@@ -131,17 +173,21 @@ The rest of the section will go through the process of analyzing participants st
 
 ### Merging Event Files
 
+{% note %}
+The file formats changed in preCICE version 3.2 from JSON to a text-based format `.txt` for rank files and a sqlite3 database `profiling.db` for profiling data.
+{% endnote %}
```
MakisH: Would it be better to use the `{% version %}` alert without an argument here, to make it clear that this is a compatibility note? I started implemented such an alert for other components, but then realized we can already just use this.

pages/docs/tooling/tooling-performance-analysis.md
MakisH: In the long run, and once precice-cli is stable, I guess we want to use it everywhere as the default entry point. The user should mainly know and interact with this.

I would expect the interface to be rather consistent: it would be nice if going from `precice-cli profiling X` to `precice-profiling X` would mean keeping `X` the same.

> If the connection between both is relevant, you could add a note saying that "profiling might already be installed as part of the cli" or similar.

I agree that such a note would be a bit easier.

> The problem is that pipx only exposes the defined scripts of the installed package, meaning `pipx install precice-cli` installs `precice-profiling` as a dependency, but doesn't create links for its scripts (such as `precice-profiling-merge`).

Sounds like a technical problem that could be fixed, right?

> > So, maybe just drop the cli tool, this section is really about the profiling tool, not the cli.
> 
> Given that we wanted another page for the cli, this looks like the best option.

I would be fine with anything, as long as there is an intuitive and documented mapping between the interface of both.

fsimonis: > In the long run, and once precice-cli is stable, I guess we want to use it everywhere as the default entry point. The user should mainly know and interact with this.

Then let's just port everything to `precice-cli`.

> Sounds like a technical problem that could be fixed, right?

It's by design. If you install application X, you are getting X. It's a bit like installing a flatpak.

> >     If the connection between both is relevant, you could add a note saying that "profiling might already be installed as part of the cli" or similar.
>
> I agree that such a note would be a bit easier.

&

> > So, maybe just drop the cli tool, this section is really about the profiling tool, not the cli.
>
> Given that we wanted another page for the cli, this looks like the best option.

Suggestion: I change everything to `precice-cli` and add a section for using installing only these tools which could be handy in situations such as scripting/CI/pipelines/workflows whatever.
`precice-cli` installs a lot of extra stuff.

pages/docs/tooling/tooling-performance-analysis.md
fsimonis: Confusing indeed. See https://github.com/precice/precice.github.io/pull/567#discussion_r2259890099

pages/docs/tooling/tooling-performance-analysis.md
MakisH: > Suggestion: I change everything to `precice-cli` and add a section for using installing only these tools which could be handy in situations such as scripting/CI/pipelines/workflows whatever. `precice-cli` installs a lot of extra stuff.

Sounds good to me :+1:

MakisH:
> Ideally, both tools would exhibit the same CLI.

That's what I advocate for as well.

> Why not keeping the current interface? I like it. Having precice-cli as prefix seems a bit annoying.

Having `precice-cli` as prefix is, I think, the long-term better solution, as this would be something that users would get familiar with (especially with the pipx context that @fsimonis mentioned above). If the interface is the same, and you know about `precice-profiling`, then an experienced user could directly work in the lower layer.

> precice-profiling-X could have the advantage of auto-completion

This should be possible to achieve in the `precice-cli` as well. Other tools support auto-completion in deeper levels as well (e.g., `docker run` or `apt install`.

fsimonis:
> > Ideally, both tools would exhibit the same CLI.
>
> That's what I advocate for as well.

They CLI of the tools are identical.

> If the interface is the same, and you know about precice-profiling, then an experienced user could directly work in the lower layer.

It's easier to install a fixed version directly via precice-profiling.

> This should be possible to achieve in the precice-cli as well. Other tools support auto-completion in deeper levels as well (e.g., docker run or apt install.

With the `precice-profiling` package, you get executable-based autocompletion `precice-profiling-<TAB>` lists all commands. This PATH-based autocompletion is built into many shells.

Argument autocompletion is way more tricky as it is shell-specific. There exists [argcomplete](https://pypi.org/project/argcomplete/) for python, but this needs to be installed and configured by the user. pipx uses this for example.

pages/docs/tooling/tooling-performance-analysis.md
```diff
@@ -22,33 +22,34 @@ You can then use various tools to analyze, visualize, and export the profiling d
...
 The most important tools include:
 
 * `merge` merges rank files after a simulation to a single profiling file for further processing.
 * `trace` converts the profiling data into the [trace events format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview) for visualization.
 * `export` converts the profiling data into a CSV file to interface with other programs.
 * `analyze` to analyze the profiling information of a single solver
 
+{% note %}
+To install the profiling tools without the full CLI, install the `precice-profiling` PyPi package.
+Each command is available as a separate executable starting with `precice-profiling-`.
```
MakisH, line 50: An example here would be quite convenient and quicker to notice and comprehend:

"For example, `precice-profiling merge A B` becomes `precice-profiling-merge A B`."

pages/docs/tooling/tooling-performance-analysis.md
fsimonis: Done

fsimonis:
I incorporated the feedback and will merge now.

PR #570, opened 2025-08-14
Literature guide: Add more completed dissertations

MakisH:
@BenjaminRodenberg Please suggest a DOI-based URL for your thesis, if one exists

@uekerman anything else I forget?


BenjaminRodenberg:
DOI for my thesis does not exist (yet). As soon as there is a DOI we can still update it. Please don't wait for me.

PR #572, opened 2025-08-22
Add time interpolation example using different time scales

fsimonis:
This PR adds a basic explanation on how to use time interpolation to provide smooth input data for a solver operating on a smaller timescale.

### DIFF
--- a/pages/docs/couple-your-code/couple-your-code-waveform.md
+++ b/pages/docs/couple-your-code/couple-your-code-waveform.md
@@ -94,10 +94,10 @@ while (not simulationDone()){ // time loop
 ### Different time scales
 
 For solvers operating on different time-scales, the solver with the smaller time-step size may need smooth input data.
-In this case the solver with the smaller step size may call `readData()` with `dt=0` to sample data at the beginning of each of it's time steps.
+In this case the solver with the smaller step size may call `readData()` with `relativeReadTime=0` to sample data at the beginning of each of its time steps.
 By default, this results in piecewise linearly interpolated data.
 
-For parallel-implicit coupling, interpolation will be available starting from the second iteration.
+For parallel-implicit coupling, interpolation is available starting from the second iteration.
 For serial coupling schemes, the `second` solver will always have linear interpolation available.
 
 To achieve higher-order interpolation, let the solver with the larger time-steps perform multiple time-steps per time-window and increase the `waveform-degree="2"` of the data in question.
--- a/pages/docs/couple-your-code/couple-your-code-waveform.md
+++ b/pages/docs/couple-your-code/couple-your-code-waveform.md
@@ -66,11 +66,11 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
 ## Availability
 
 Depending on the used coupling scheme, time interpolation may not always be available.
-The following table shows the availability, which without further changes means linear interpolation between the time samples at the beginning and the end of the window.
-Note that due to the staggered nature of serial schemes, time-interpolation is always available in the participant that goes second.
-This is the only instance where time-interpolation is available in an explicit scheme.
+The following table shows when it is available, which without further changes means linear interpolation between the data samples at the beginning and the end of the window.
+Note that due to the staggered nature of serial schemes, time interpolation is always available in the participant that goes second.
+This is the only instance where time-interpolation is available in an explicit coupling scheme.
 
-| Scheme            | first iteration | later iterations |
+| Coupling scheme | First iteration | Later iterations |
 | ---               | ---             | ---              |
 | serial-explicit   | second only     |                  |
 | serial-implicit   | second only     | yes              |
@@ -117,7 +117,7 @@ while (not simulationDone()){ // time loop
 
 ### Different time scales
 
-For solvers operating on different time-scales, the solver with the smaller time-step size may not be able to handle constant input data over the entire time-window.
+For solvers operating on different time scales, the solver with the smaller time-step size may not work with constant input data over the entire time-window.
 In cases where [time-interpolation is available](#availability), the solver with the smaller step size may call `readData()` with `relativeReadTime=0` to sample data at the beginning of each of its time steps.
 
 By default, this results in piecewise linearly interpolated data.
--- a/pages/docs/couple-your-code/couple-your-code-waveform.md
+++ b/pages/docs/couple-your-code/couple-your-code-waveform.md
@@ -63,7 +63,9 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
 </precice-configuration>
 ```
 
-## Usage example
+## Usage examples
+
+### Implicit coupling
 
 We are now ready to extend the example from ["Step 6 - Implicit coupling"](couple-your-code-implicit-coupling.html) to use waveforms. Let us assume that our fluid solver uses a midpoint rule as time stepping method. In this case, only few changes are necessary to sample the `Displacements` at the middle of the time window:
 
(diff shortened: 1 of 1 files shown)


pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
 ...
 ```
 
+### Different time scales
+
+For solvers operating on different time-scales, the solver with the smaller time-step size may need smooth input data.
```
davidscn:
```suggestion
For solvers operating on different time-scales, the solver with the smaller time-step size may need interpolated input data.
```
?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
...
+In this case the solver with the smaller step size may call `readData()` with `dt=0` to sample data at the beginning of each of it's time steps.
```
davidscn:
```suggestion
In this case the solver with the smaller step size may call `readData()` with `relativeReadTime=0` to sample data at the beginning of each of its time steps.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
...
+By default, this results in piecewise linearly interpolated data.
+
+For parallel-implicit coupling, interpolation will be available starting from the second iteration.
+For serial coupling schemes, the `second` solver will always have linear interpolation available.
+
+To achieve higher-order interpolation, let the solver with the larger time-steps perform multiple time-steps per time-window and increase the `waveform-degree="2"` of the data in question.
```
davidscn:
```suggestion
To achieve higher-order interpolation, let the solver with the larger time-steps perform two time-steps per time-window and increase the `waveform-degree="2"` of the data in question.
```
? I dont think higher is possible?

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
 ...
 ```
 
+### Different time scales
+
+For solvers operating on different time-scales, the solver with the smaller time-step size may need smooth input data.
+In this case the solver with the smaller step size may call `readData()` with `dt=0` to sample data at the beginning of each of it's time steps.
+By default, this results in piecewise linearly interpolated data.
+
+For parallel-implicit coupling, interpolation will be available starting from the second iteration.
```
davidscn:
```suggestion
For parallel-implicit coupling, interpolation is available starting from the second iteration.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
...
+For serial coupling schemes, the `second` solver will always have linear interpolation available.
```
davidscn:
```suggestion
For serial-implicit coupling schemes, the `second` solver has linear interpolation available.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +91,18 @@ while (not simulationDone()){ // time loop
...
+
+To achieve higher-order interpolation, let the solver with the larger time-steps perform multiple time-steps per time-window and increase the `waveform-degree="2"` of the data in question.
+Serial-explicit schemes need to manually enable exchange of substeps from the `first` to the `second` participant using `<exchange substeps="true" />`.
```
davidscn: Maybe also mention that time interpolation in first participant is not possible

davidscn (changes requested):

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: Linear interpolation is also available for the second in a serial-explicit scheme as it received time-samples at the beginning and the end of the time-window.

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: I picked `smooth` because interpolation is more of a necessity, rather than a requirement. Maybe it is better to explain the actual problem here instead:

```suggestion
For solvers operating on different time-scales, the solver with the smaller time-step size may not be able to handle constant input data over the entire time-window.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: preCICE can generate interpolate of arbitrary order https://github.com/precice/precice/pull/2039.

For an order to two we need at least two time-steps per time-window, hence multiple.
`at least two time-steps` is also a possibility.

pages/docs/couple-your-code/couple-your-code-waveform.md
davidscn: Sure, but then it reads odd, since you mention serial-explicit schemes later on after talking about order. So maybe then move the configuration sentence up

pages/docs/couple-your-code/couple-your-code-waveform.md
davidscn: Ah ok I didnt know that

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: True, it reads odd. I think we need an availability section to generally explain this. I'll add one

pages/docs/couple-your-code/couple-your-code-waveform.md
davidscn: I think "interpolated" just describes it all. The motivation to use time interpolation is up to the user

fsimonis: Reworded it.

pages/docs/couple-your-code/couple-your-code-waveform.md
fsimonis: Added the availability section

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -63,7 +63,33 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
 </precice-configuration>
 ```
 
-## Usage example
+## Availability
+
+Depending on the used coupling scheme, time interpolation may not always be available.
+The following table shows the availability, which without further changes means linear interpolation between the time samples at the beginning and the end of the window.
```
IshaanDesai:
```suggestion
The following table shows when it is available, which without further changes means linear interpolation between the data samples at the beginning and the end of the window.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -63,7 +63,33 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
...
+Note that due to the staggered nature of serial schemes, time-interpolation is always available in the participant that goes second.
```
IshaanDesai:
```suggestion
Note that due to the staggered nature of serial schemes, time interpolation is always available in the participant that goes second.
```
To be consistent with what is written above.

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -63,7 +63,33 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
...
+This is the only instance where time-interpolation is available in an explicit scheme.
```
IshaanDesai:
```suggestion
This is the only instance where time-interpolation is available in an explicit coupling scheme.
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -63,7 +63,33 @@ This allows us to define the degree of the interpolant in the `read-data` tag of
...
+
+| Scheme            | first iteration | later iterations |
```
IshaanDesai:
```suggestion
| Coupling scheme | First iteration | Later iterations |
```

pages/docs/couple-your-code/couple-your-code-waveform.md
```diff
@@ -89,6 +115,14 @@ while (not simulationDone()){ // time loop
 ...
 ```
 
+### Different time scales
+
+For solvers operating on different time-scales, the solver with the smaller time-step size may not be able to handle constant input data over the entire time-window.
```
IshaanDesai:
```suggestion
For solvers operating on different time scales, the solver with the smaller time-step size may not work with constant input data over the entire time-window.
```

IshaanDesai (approved):
Looks good. Minor comments from my side.

PR #574, opened 2025-08-27
Workshop 2025: Change the number of the second room close to the main room

gdenayer:
Hi,

We book a new room (404 instead of 205) closer to the main room (405/406) for the preCICE workshop.

Regards


MakisH (approved):
Thanks!

Closes #575.

PR #576, opened 2025-08-27
Add unreleased changes to the roadmap

fsimonis:
The roadmap mentions

> If you are looking for features introduced already in the past, have a look at our Changelog.

this doesn't mention unreleased changes though.

This PR adds a dynamic section called unreleased changes to the roadmap, which is generated directly from the preCICE repo. Therefore, it is not indexed by search engines.

The code imitates the final changelog style.

image

### DIFF
--- a/pages/docs/fundamentals/fundamentals-roadmap.md
+++ b/pages/docs/fundamentals/fundamentals-roadmap.md
@@ -10,8 +10,8 @@ preCICE applies [Semantic Versioning](https://semver.org/), introducing new func
 
 In this page, you can find information about features that we plan to introduce in next releases. This is not meant to be a strict schedule, but rather a hint on the directions that preCICE is heading towards. We also have a few [issue milestones](https://github.com/precice/precice/milestones), which are updated more often. Issues and work packages of bigger features are generally grouped in [projects](https://github.com/precice/precice/projects).
 
-If you are looking for features introduced already in the past, have a look at our [Changelog](https://github.com/precice/precice/blob/develop/CHANGELOG.md).
-For most recent developments waiting for the next release, have a look at the [unreleased changes](#unreleased-changes).
+If you are looking for library features introduced already in the past, have a look at our [Changelog](https://github.com/precice/precice/blob/develop/CHANGELOG.md).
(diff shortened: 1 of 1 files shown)


MakisH:
Nice idea, nice feature!

I guess this will typically get quite long, making the rest of the page a bit daunting. Could we wrap this list in a collapsible element?

Note that this only focuses on the core library, but in this page we often include major developments in other components as well. We could depict this in the section heading.

fsimonis:
> I guess this will typically get quite long, making the rest of the page a bit daunting. Could we wrap this list in a collapsible element?

That's why I put it at the bottom of the page, so it's essentially trailing. I was thinking to add a "click to load" button.

> Note that this only focuses on the core library, but in this page we often include major developments in other components as well. We could depict this in the section heading.

Good point.

MakisH:
> > I guess this will typically get quite long, making the rest of the page a bit daunting. Could we wrap this list in a collapsible element?
> 
> That's why I put it at the bottom of the page, so it's essentially trailing. I was thinking to add a "click to load" button.

If we add multiple repositories (which are not yet ready, as they don't follow the same system), a "click to load" would be a nice solution to minimize requests and reduce load time.

For the length, any of the two (collapsible element or "click to load") would be good.

In any case, it should never get that long, I just find short pages more inviting to read.

fsimonis:
Done.

image

Once we move to a more sustainable system like hugo, we can build this from the repos whenever we update the site and wrap it in a collapsible container.

pages/docs/fundamentals/fundamentals-roadmap.md
```diff
@@ -41,12 +41,17 @@ For most recent developments waiting for the next release, have a look at the [u
 - [A general mocked interface for testing](https://github.com/precice/preeco-orga/issues/4)
 - Data compression for waveform relaxation
 
-## Unreleased changes
+## Unreleased changes in preCICE
```
MakisH, line 45: We could have here an opening sentence, also linking to https://precice.org/dev-docs-release-strategy.html#release-schedule

fsimonis: Done

MakisH (approved):

PR #577, opened 2025-08-27
Add documentation on restarting CalculiX simulations

MakisH:
This PR shapes [this Medium article](https://pawel-lojek.medium.com/resuming-fsi-simulations-with-openfoam-calculix-896088861ae) by @moon1144 (written for a much older state of preCICE) into a section of the CalculiX adapter documentation.

I did not have success restarting yet (my flow simulation crashes after a few iterations in the first time window after restarting), but I think this is a first step long overdue.

Why now:

- I recently found out about @moon1144's article (I think suggested by @IshaanDesai) and did not want to let it slip through
- this year's workshop is approaching, and last year I talked a bit about this with @MatthiasFreimuth
- @vidulejs is currently debugging some simulations using restarting, so he might have some additional insight

All kind of feedback welcome right now. This is something that many users are asking for often.


MakisH:
Additional notes on my failing case:

- I am using the perpendicular-flap tutorial
- I am trying both with and without `FSI { restartFromDeformed true; }` in the `fluid-openfoam/system/preciceDict`.
- I am running until and restarting from 0.5s
- I am not cleaning up before running again, just setting `startFrom 0.5` in the `controlDict` and using the modified input file for CalculiX.
- I am setting just `<max-time value="0.5" />` in the preCICE config. The `max-time-windows` stated in the article should not be needed.

content/docs/adapters/calculix/adapter-calculix-configure.md
```diff
@@ -280,3 +280,44 @@ Make sure to replace `[CalculiX input file]` and `[CalculiX participant name]` w
...
+  ```text
+  *RESTART,READ,STEP=1
+  ```
+
+  For every new step of restarting the simulation, increase the respective number: when you restart again to go further beyond in time, set `STEP=2`.
+4. Since all the rest of the configuration is included in the restart file, we need to remove the rest of the definitions. In the end, the input file should look like this:
+
+   ```text
+   *RESTART,READ,STEP=1
+   *STEP,NLGEOM,  INC=1000000
```
vidulejs: The perpendicular-flap case doesn't use NLGEOM I believe

MakisH: Thanks for the observation! This shouldn't matter though, right?

MakisH:
This content is more useful on the website than hidden in this branch. I added a note that the section might be incomplete, asking for help.

PR #603, opened 2025-09-15
Update Gemfile for compatibility

BytePaul:
## Description

This PR updates the Gemfile to improve compatibility with newer Ruby/Jekyll versions and to remove deprecated or unnecessary dependencies.

### Changes:

Added explicit Ruby version requirement: ruby ">= 3.1"

Pinned Jekyll version with an upper bound: gem "jekyll", ">= 4.1.1", "< 5.0"

Switched posix-spawn back to the released gem (removed GitHub PR reference)

Removed jekyll-algolia (deprecated, pins outdated Nokogiri)

Added a version constraint for Nokogiri: ~> 1.18

Removed open-uri (it’s part of Ruby stdlib, no gem needed)

Kept webrick for Ruby 3+ compatibility

### Rationale:

Avoids build errors caused by outdated jekyll-algolia dependency.

Ensures consistent builds by constraining Jekyll and Nokogiri versions.

Removes unnecessary gems (open-uri).

Clarifies Ruby version support explicitly.


MakisH:
Decision: While we could use this as a reference to update the dependencies in the future, there is no explicit need right now.

PR #606, opened 2025-09-19
Add hint for disabling unity builds in Building from Source: Troubleshooting

gdenayer:
If you encounter an error such as redefinition of ‘std::unique_ptr<precice::Participant> impl’ during compilation, it is caused by CMake’s Unity (Jumbo) build mode. To resolve this, disable Unity builds when configuring preCICE.


MakisH:
Thanks for the contribution! I polished it a bit, here is how it looks now:

Screenshot from 2025-10-07 17-38-21

@fsimonis since this is related to the (already heavily discussed) unity builds, your feedback would be helpful here.

MakisH (approved):

PR #584, opened 2025-09-03
Feature/forum topics fetch

BytePaul:
This PR adds a new "Forum Browser" page under the Community section.  
1. Introduces /community-forum-browser.html with search, filter, and sort for mirrored Discourse topics  
2. Adds link to Community sidebar  
3. Includes tools/fetch_precice_forum.py to update assets/data/forum/topics.json

PR #611, opened 2025-10-09
News cards on the landing page

### LINKED ISSUE #402: Integrate blog posts on website
Blog posts live on Discourse. Currently we link from the button "blog" in the header to the news category on Discourse.
It would be nice if we could integrate the content directly on the website somehow without duplication.
Could also be important for SEO?

Connected to #146 ?

BytePaul:
Adds a "Latest news" section on the landing page.

The data is fetched using `tools/fetch-news.py`.
The CI workflow `update-discourse-data.yml` is executing this script nightly (or on demand) and adds the respective `assets/data/news.json` file in the repository, committing as the preCICE bot.

The data is populated on the client side using `news-collect.js`.

This PR also cleans up some minor style issues in `css/customstyles-precice.css` and `css/landing-page.css`.

Closes https://github.com/precice/precice.github.io/issues/402


MakisH:
Thank you for the contribution! Notes from our discussion:

- [x] Make it work for the production site (there are a few requests to `127.0.0.1`)
- [x] Add some text from the body of each post
- [x] Always hide the pinned "Welcome to the category" item
- [x] Remove the external link marker at the bottom
- [x] Resolve the conflicts in `css/customstyles-precice.css`

Since this is now fetching data dynamically, don't forget to use the same (or similar) in #584 as well, so that we can merge it.

MakisH:
I can now build and run this locally, which adds both the news ~~and the FAQ~~* on the landing page. This is very nice, thank you for the quick updates yesterday and today!

Could you please still:

- [x] Fix the conflicts on `css/customstyles-precice.css`
- [x] Make the "more news" button the same color as the "More questions" button in the FAQ (or maybe the same color as the next background, or something else from the existing palette)
- [x] Remove the external link icon from the button
- [x] Use the same arrow as in the FAQ button

Screenshot from 2025-10-16 16-33-43

* (I got a bit confused, this is was hard-coded implementation that already existed)

BytePaul:
Hi Gerasimos, 

thank you for the feedback.

i have recently fixed the issue with `css/customstyles-precice.css`. Can you check and let me know ?

Regards,
Paul

MakisH:
Notes from our discussion:

- ~~[ ] Add a `package.json` and a `package-lock.json`~~
- Resolve the conflict in the css files (see https://github.com/precice/precice.github.io/pull/611/files and the bottom of this PR)
- [x] Link to more news points to the wrong category

BytePaul:
Hi Gerasimos,
I have made the improvements as discussed.

package.json
MakisH, line 1: This should be in the `tools/` directory, no? Jekyll itself already uses a `package.json` in this location.

tools/news-proxy.js
MakisH, line 1: How would we run this script in production? The website is hosted on GitHub Pages, which cannot run additional nodejs applications.

However, we only need to fetch data, which is publicly available without any authentication.

In the end, we can also do the data fetching in the CI (using this or a similar script) when we build the website (which is frequent enough), but we need a way to render it in the static page. For that, I would expect that the corresponding script runs once, gets the data from Discourse, and creates the respective HTML elements.

MakisH (changes requested):
While the solution now looks visually good, there is a blocking issue that I did not realize before figuring out how to try this PR out (a more detailed PR description would have been helpful here): the approach with having a separate server is not possible in production, since the website is hosted from this repository using GitHub Pages.

This does not mean that this is a dead-end, there are a few solutions. Let's discuss in the thread below.

tools/news-proxy.js
BytePaul: Hi @MakisH 

do you have admin rights for the precice.discourse.group forum ?

MakisH: yes

MakisH: More context here:

1. There are currently two PRs that rely on the same mechanism: this and #612
2. The first approach was adding the JSON file directly in the repository. As the data file does not need to be version-controlled, I suggested either fetching it dynamically, or adding it to Git LFS, as a first  (https://github.com/precice/precice.github.io/pull/584#discussion_r2318596246).
3. The CORS approach should work, as Discourse allows us to set `precice.org` as a CORS header. However, this might be an overkill in terms of features and number of requests. Enabling CORS in the Discourse settings should change the response of `https://github.com/precice/precice.github.io/pull/612#issuecomment-3421788511`.
4. In a discussion between @BytePaul and me, and with feedback from @fsimonis in https://github.com/precice/precice.github.io/pull/612#issuecomment-3421788511, we decided to indeed try a solution that fetches the data on the CI when building the website.

.github/workflows/fetch-news.yml
```diff
@@ -0,0 +1,33 @@
+name: Fetch Discourse News
+
+on:
+  workflow_dispatch:    
+  schedule:
+    - cron: "0 6 * * *"
```
MakisH: I am thinking whether it makes sense to have this:

a) as a separate workflow (could also be merged with `update-submodules.yml`, but that's actually also conveniently stand-alone)
b) running as a cron job (the updates would only be reflected if the website is rebuilt, i.e., in the `deloy.yml` and the `build.yml` workflows)
c) if as a cron job, maybe closer (but not racing with) the `update-submodules.yml`

The good thing of having it as an independent workflow is that, in case this (or the submodules workflow) fails, the website build can still go through. Same for PRs.

Let's keep it as-is.

.github/workflows/fetch-news.yml
MakisH, line 1: For later integrating with the FAQ entries as well, and for consistency with the `update-submodules.yml`, we could already name this workflow `update-discourse-data.yml`.

.github/workflows/fetch-news.yml
```diff
@@ -0,0 +1,33 @@
+name: Fetch Discourse News
```
MakisH: Similarly to the file comment:


```suggestion
name: Update Discourse data
```

.github/workflows/fetch-news.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        run: npm ci || npm install
+
+      - name: Fetch latest news from Discourse
+        run: node tools/fetch-news.js
+
+      - name: Commit and push updated news.json
```
MakisH: While this goes against what we already discussed in https://github.com/precice/precice.github.io/pull/611#discussion_r2443476398, having these json files in the repository does not seem to be a bad idea:

- it is a very simple solution
- the files are ascii, small, and can be versioned

We could alternatively add these files in Git LFS, which we already use for the submodules in the `imported/` directory. Not sure if that would work, but it should.

I would implement it with LFS, but it is also fine as-is for now.

.github/workflows/fetch-news.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          git config user.name "GitHub Action"
+          git config user.email "actions@github.com"
```
MakisH: For consistency with the `update-submodules.yml`:


```suggestion
          git config --local user.name "precice-bot"
          git config --local user.email "info@precice.org"
```

.github/workflows/fetch-news.yml
```diff
@@ -0,0 +1,33 @@
...
+          git add assets/data/news.json
+          git commit -m "Update news data [skip ci]" || echo "No changes to commit"
+          git push
```
MakisH: This should not work. Instead, for consistency:

```suggestion
      - name: Push commit
        uses: ad-m/github-push-action@master
        with:
          github_token: ${{ steps.app-token.outputs.token }}
          branch: ${{ github.ref }}
```

We might even be able to use the GitHub CLI for this (but no need to).

_layouts/landing_page.html
```diff
@@ -16,7 +16,11 @@
 
 	{% include footer.html %}
 
-	<script src="{{ "js/github-queries.js" }}"></script>
+	<!-- Existing script -->
+	<script src="{{ '/js/github-queries.js' | relative_url }}"></script>
+
+	<!-- for dynamic Latest News -->
+	<script src="{{ '/js/news-collect.js' | relative_url }}"></script>
```
MakisH: These comments are not relevant/helpful:


```suggestion
	<script src="{{ '/js/github-queries.js' | relative_url }}"></script>

	<script src="{{ '/js/news-collect.js' | relative_url }}"></script>
```

css/customstyles-precice.css
MakisH, line 1: This file includes several unrelated formatting changes and some commented-out code.

In the end, it should only include what is relevant for this feature.

css/customstyles-precice.css
```diff
@@ -378,5 +389,113 @@ details > summary::after {
 
 /* open */
 details[open] > summary::after {
+    content: "▲";
+}
+/* ========================================================================= */
+/* === START: Latest News Section Styles (Cleaned & Unified) === */
+/* ========================================================================= */
```
MakisH: How does this overlap with `landing-page.css`? Maybe some cleanup is needed here.

tools/fetch-news.js
MakisH: I understand that this script is not needed anymore. Correct?

tools/news-proxy.js
MakisH, line 1: I understand that the `news-proxy.js` is not needed anymore, since the files are always included as assets. Correct?

tools/fetch-news.js
MakisH, line 1: Now that we don't need an additional server, having such a data fetching script written in JS with node seems a bit unmotivated (and rather overkill), which might be complicated to maintain in the future.

This could easily be converted to Python, which we already use everywhere else, and should be easier to maintain.

MakisH (changes requested):
Getting closer!

The functionality is there, but a few small things need to be improved. Moreover, converting the `fetch-news.js` to Python would make the whole thing easier to maintain in the future.

Some cleanup is also still needed.

Let's also document the respective workflow for updating the data from Discourse in the `README.md`, below the seciton for updating the submodules.

Don't forget to update the PR description to reflect the latest state.

tools/news-proxy.js
BytePaul: yes

css/customstyles-precice.css
MakisH: Still open:

- [x] Why removing lines 9-11 (`main {}`)? While it seemingly does nothing right now, it does not seem to be related to the feature. If there is a reason, happy to learn about it. -> Because it is empty and VSCode was giving a warning. No real reason to keep it around.
   - [x] Same in `landing-page.css`. 
- [x] Line 233 (old): Why renaming the `h2.header` `border-top-radius` to `border-radius`? -> Not a valid CSS property.
- [x] Lines 243-253 (old): Formatting (I guess the `.post-content` also needs to be indented in). Why remove the navtabs comment? -> Fixed.

tools/news-proxy.js
MakisH: Then, I would suggest removing it, together with the respective `package.json`.

Same for the `fetch-news.js`.

If you think it would be a good idea to maintain them, then we should add comments/documentation that they are not used and in which case they might become relevant again.

.github/workflows/update-discourse-data.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
```
MakisH: In CI workflows, I would rather pin the Python version, to avoid surprises.
There is also already a newer version of this Action.
How about:


```suggestion
      - name: Set up Python
        uses: actions/setup-python@v6
        with:
          python-version: "3.14"
```

MakisH (commented):
The new script also works, nice!

Main pending question is what to do with the old ones that are not needed anymore. I suggest removing them.

See also some tiny pending cleanup tasks.

.github/workflows/fetch-news.yml
MakisH, line 1: This file also needs to be removed, since we now have `update-discourse-data.yml`.

MakisH (approved):

Issue #31, opened 2018-12-21
Add physical units to the description of adapters


MakisH:
We have a [list of all the fields an adapter supports for coupling](https://www.precice.org/codes/) (e.g. temperature, heat flux, etc), but we should also mention the expected physical units / dimensions. These sometimes differ from adapter to adapter, or may differ for in-house adapted solvers.

MakisH:
I feel that this issue is rather out of scope for us and impossible to maintain.
The units are stemming from the solvers and sometimes from the user configurations (e.g., in OpenFOAM).

PR #622, opened 2025-10-29
Update main description on landing page

uekerman:
The old description was already a bit outdated missing volume coupling, multi-scale, ...
I tried to also remove aspects to not increase the overall length.


fsimonis:
@uekerman Something that occasionally comes up during the workshop is the question of which units are used for time and data. It could be helpful to add the term **unit-less** in the intro to convey early that there is no limitation from preCICE.

content/index.html
```diff
@@ -32,13 +32,13 @@ <h2 class="subtitle no-anchor">The coupling library for partitioned multi-physic
     <div class="row">
       <div class="col-md-6">
         <p>
-          preCICE is an <strong>open-source coupling library</strong> for partitioned multi-physics simulations, including, but not restricted to fluid-structure interaction and conjugate heat transfer simulations.
+          preCICE is an <strong>open-source coupling library and ecosystem</strong> for general partitioned multi-physics and multi-scale simulations, including, surface coupling and volume coupling.
```
MakisH:
```suggestion
          preCICE is an <strong>open-source coupling library and ecosystem</strong> for general partitioned multi-physics and multi-scale simulations, including surface coupling and volume coupling.
```

MakisH: I would even make it:


```suggestion
          preCICE is an <strong>open-source coupling library and ecosystem</strong> for general partitioned multi-physics and multi-scale simulations, including surface and volume coupling.
```

(to me, this reads more easily and compactly, even if more vaguely)

MakisH (approved):
Looks very good and very much needed, thanks! The "and ecosystem" addition in the description is also very-very much appreciated. 🙏

The drop of "fluid-structure interaction and conjugate heat transfer" is also very much due.

There are also a few more places to update this information. Non-exhaustive list:

- [x] https://github.com/precice/.github/tree/main/profile/README.md
- [x] Description of https://github.com/precice/precice
- A few places where preCICE is listed as a package

Rendering:

Image

MakisH:
> @uekerman Something that occasionally comes up during the workshop is the question of which units are used for time and data. It could be helpful to add the term **unit-less** in the intro to convey early that there is no limitation from preCICE.

This is important, you are right. As it is a detail (the description does not imply dimensions anyway), should we maybe have a related "feature block" below?

See also an older related issue that was closed because it was rather impossible to address at scale: https://github.com/precice/precice.github.io/issues/31

fsimonis:
> This is important, you are right. As it is a detail (the description does not imply dimensions anyway), should we maybe have a related "feature block" below?
> 
> See also an older related issue that was closed because it was rather impossible to address at scale: #31

Thinking about it a bit more, I think this is more suitable for the [big picture](https://precice.org/docs.html) page.

The features need some improvement marketing wise. But that is for another day.

uekerman:
> A few places where preCICE is listed as a package

This one is still open.

PR #632, opened 2025-11-04
Update the acceleration configuration guide

Fujikawas:
This PR adds the minimum configurations for different acceleration methods.

### DIFF
--- a/content/docs/configuration/basics/configuration-acceleration.md
+++ b/content/docs/configuration/basics/configuration-acceleration.md
@@ -7,7 +7,7 @@ summary: "Mathematically, implicit coupling schemes lead to fixed-point equation
 
 To find out more about the mathematical background, please refer, for example, to [this paper](https://www.sciencedirect.com/science/article/pii/S0898122115005933).
 
-In preCICE, three different types of <tt>acceleration</tt> can be configured: `constant` (constant under-relaxation), `aitken` (adaptive under-relaxation), and two quasi-Newton variants (`IQN-ILS` aka. Anderson acceleration, `IQN-IMVJ` aka. generalized Broyden). Before looking at the details, we need to understand which data gets modified when.
+In preCICE, three different types of `acceleration` can be configured: `constant` (constant under-relaxation), `aitken` (adaptive under-relaxation), and two quasi-Newton variants (`IQN-ILS` aka. Anderson acceleration, `IQN-IMVJ` aka. generalized Broyden). Before looking at the details, we need to understand which data gets modified when.
 
 ## Coupling data and primary data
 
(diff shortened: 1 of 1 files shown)


content/docs/configuration/basics/configuration-acceleration.md
```diff
@@ -67,39 +67,105 @@ Constant under-relaxation with a factor of 0.5 can be a good choice for e.g. tur
...
+
 ```xml
 <acceleration:aitken>
   <data name="Displacements" mesh="StructureMesh"/>
   <initial-relaxation value="0.1"/>
 </acceleration:aitken>
 ```
 
-Aitken under-relaxation adapts the under-relaxation factor in every iteration based on current residuals of the defined primary data and we only need to define an initial relaxation factor `initial-relaxation`. An initial relaxation factor of 0.1 usually leads to a robust simulation.
+ An initial relaxation factor of 0.1 is recommended for strong coupling scenarios.
```
uekerman:
```suggestion
An initial relaxation factor of 0.1 is recommended for strongly coupled problems.
```

content/docs/configuration/basics/configuration-acceleration.md
```diff
@@ -67,39 +67,105 @@ Constant under-relaxation with a factor of 0.5 can be a good choice for e.g. tur
...
 
 Aitken under-relaxation can be a good choice for strong interaction with a fluid solver that does not fully converge in every iteration, or for compressible fluid solvers. For most cases, however, it is beneficial to change to a quasi-Newton scheme. Using Aitken under-relaxation is generally not recommended in combination with a parallel coupling scheme (refer to table 1 in [this paper](https://doi.org/10.1016/j.camwa.2015.12.025), where *Vec-Aitken* refers to our implementation of Aitken under-relaxation for parallel coupling schemes).
 
 ## Quasi-Newton schemes
 
+In preCICE, there are two quasi-Newton variants available: <tt>IQN-ILS</tt> (aka. Anderson acceleration) and <tt>IQN-IMVJ</tt> (aka. generalized Broyden). Both are good choices for strong interactions, while `IQN-ILS` is simpler to start with.
```
uekerman:
```suggestion
In preCICE, there are two quasi-Newton variants available: `IQN-ILS` (aka. Anderson acceleration) and `IQN-IMVJ` (aka. generalized Broyden). Both are good choices for strong interactions, while `IQN-ILS` is simpler to start with.
```

In general, I would avoid html as much as possible. Or is there a good reason to use it here that I am not seeing?
(more cases below)

uekerman (approved):
Thank you for adding this! Minor comments. Please go ahead and directly merge if done.

content/docs/configuration/basics/configuration-acceleration.md
Fujikawas: They are there since there were some before. I have replaced all of them now.

PR #638, opened 2025-11-23
Training: Add instructions for new VM image

MakisH:
After the 2025 Workshop, we decided to distribute the same image that we use for the live USB, such that people can create a VM and check it before coming to the training. This PR adds instructions regarding this, and removes the link to the demo VM. In the long run (ideally after Ubuntu 26.04 is released), these could be unified, and we could also provide an image for ARM CPUs.

I left out the dependencies for the FMI module, as we do not yet describe it. Notes:

```
The FMI module also needs:

- [preCICE FMI runner](tooling-fmi-runner.html) (`pip3 install fmpy==0.3.20 numpy==1.26.4 fmiprecice`)
```

### DIFF
--- a/content/community/community-training.md
+++ b/content/community/community-training.md
@@ -139,7 +139,7 @@ The mapping module also needs:
 
 - [ASTE](tooling-aste.html) (check by running `./precice-aste-run --help` from the ASTE build directory)
 
-The FSI workflow module needs:
+The FSI workflow module also needs:
 
 - OpenFOAM (openfoam.com): See the [Quickstart](quickstart.html) page (check with running `pimpleFoam -help`)
 - [OpenFOAM-preCICE](adapter-openfoam-get.html) (check with running the Quickstart tutorial)
--- a/content/community/community-training.md
+++ b/content/community/community-training.md
@@ -68,7 +68,8 @@ Find verification instructions at the end of each section.
 
 ### Provided system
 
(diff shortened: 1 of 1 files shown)


content/community/community-training.md
```diff
@@ -56,17 +56,71 @@ We parallelize the same Python codes used in the Basics module, we analyze the p
...
+- Ideally, four logical CPU cores (one is also fine).
+- At least 64MB of video RAM (more -> smoother graphics).
+- At least 25GB of storage.
+
+For VirtualBox, find these settings under the `System`, `Display`, and `Storage` categories.
+
+Further important settings:
+
+- The OpenFOAM adapter is installed under `~/OpenFOAM/ubuntu-v2406`. If you install (not "try" in a live session), move that directory to your user: `cd ~/OpenFOAM/ && mv ubuntu-v2406/ $USER-v2406`.
+- Set your keyboard layout: In the Ubuntu applications menu, type "keyboard layout". Select "Add input source". Remove the default one.
```
uekerman, line 99: when installing the Ubuntu version in the VM, there is anyway a dialogue for this.

uekerman (approved):

content/community/community-training.md
MakisH: But this is already needed in the installer or the try mode.

PR #649, opened 2025-12-16
Add user story of Matilde Fiore

### LINKED ISSUE #646: [Story]: Closed-loop simulations to analyze natural convection flow in deformed bundles
### Your name

Matilde Fiore, PhD

### Affiliation

von Karman Institute for Fluid Dynamics

### Contact details

matilde.fiore@vki.ac.be

### Title

Closed-loop simulations to analyze natural convection flow in deformed bundles

### Your story

Deformation of fuel pins in the core of nuclear reactors is the result of several concurrent phenomena and directly affects the peak temperatures reached in the cladding layers. In forced-convection regimes, the excess temperature induced by pin deformations can be reasonably estimated through numerical simulations that focus only on the active region of the assembly. Simulations of fuel bundles under natural convection are significantly more challenging, as the entire primary loop must be resolved in order to accurately capture buoyancy-driven flow dynamics. As the computational domain increases in size, mesh generation becomes increasingly demanding: it is necessary to limit the total number of cells while preserving good mesh quality and appropriately handling transition regions between zones of high and low spatial resolution. 
Using preCICE, we simplified the meshing process by partitioning the full domain into two subdomains: the active region of the bundle and the remaining part of the loop. The two subdomains have very different resolution and mesh structure (structured vs unstructured) and are handled by different solvers (chtMultiRegionFoam and buoyantPimpleFoam for the bundle and the loop, respectively). At the two shared interfaces, the coupling is enforced by means of custom boundary conditions and the exchange of derived fields, enabling a consistent, pressure-driven flow across the coupled domains.

### More info

_No response_

### Image

- [x] I will upload an image to be published on the website directly after opening this issue, or send it via email to @MakisH.

MakisH:
Closes #646.

With the submitted text, it looks like this:

Screenshot from 2025-12-16 20-17-39

With the two minor changes I suggest, it will look like this:

Screenshot from 2025-12-16 20-27-10

@matilde92 if you agree, I will apply the suggestions and merge.

I see that there is nothing to link to at the moment, we can always add something later.


collections/_testimonials/26-vki-closed-loop.md
```diff
@@ -0,0 +1,11 @@
+---
+title: "Closed-loop simulations to analyze natural convection flow in deformed bundles"
```
MakisH: The title is a bit long. I would suggest, instead:


```suggestion
title: "Closed-loop natural convection flow in deformed bundles"
```

collections/_testimonials/26-vki-closed-loop.md
```diff
@@ -0,0 +1,11 @@
...
+
+Using preCICE, we simplified the meshing process by partitioning the full domain into two subdomains: the active region of the bundle and the remaining part of the loop. The two subdomains have very different resolution and mesh structure (structured vs unstructured) and are handled by different solvers (chtMultiRegionFoam and buoyantPimpleFoam for the bundle and the loop, respectively). At the two shared interfaces, the coupling is enforced by means of custom boundary conditions and the exchange of derived fields, enabling a consistent, pressure-driven flow across the coupled domains.
```
MakisH: The "coupling is enforced by means of custom boundary conditions" is a bit confusing in preCICE terminology. I suggest, instead:


```suggestion
Using preCICE, we simplified the meshing process by partitioning the full domain into two subdomains: the active region of the bundle and the remaining part of the loop. The two subdomains have very different resolution and mesh structure (structured vs unstructured) and are handled by different solvers (chtMultiRegionFoam and buoyantPimpleFoam for the bundle and the loop, respectively). At the two shared interfaces we used custom boundary conditions and exchange derived fields, enabling a consistent, pressure-driven flow across the coupled domains.
```

MakisH:
Merging here to not keep this open too long, since the changes were rather trivial. Let me know if we should still tweak or add anything, anytime! :)

PR #660, opened 2026-01-02
Add Quickstart to the tutorials sidebar

MakisH:
The Quickstart is currently disconnected from the other tutorials. Someone that enters from a tutorial, might not notice it.

I wonder if we should actually just render the Quickstart among the other tutorials. That would be very easy: just change the `permalink` (and add a redirect from the old one) and relate it to the tutorials sidebar.


MakisH:
Since this is rather trivial, I am merging it to see how it will evolve and get feedback.

PR #672, opened 2026-01-04
Add environment config example for local build

### LINKED ISSUE #168: Documentation for pkg-config for local build
In https://precice.org/installation-source-installation.html, we describe how to set the `PKG_CONFIG_PATH` for an installed copy of preCICE.

https://github.com/precice/precice/pull/1238 added a new `libprecice.pc` file directly in the `build/` directory, as described in https://github.com/precice/precice/pull/1238#issuecomment-1108308991.

This would generally be useful also for users that just want to use preCICE without installing it to another path. In any case, we should document it.

However, I think we should now look at the bigger picture of how we introduce this option to the user. Just editing this page to add a hint is not enough, in my opinion.

MakisH:
In the [Building from source - Finding](https://precice.org/installation-source-finding.html), this PR adds an example of environment configuration for using preCICE directly from the build directory.

It takes into account and repeats all previous warnings.

Why: This is what many students need. I anyway send them my hard-coded example when they start.

Screenshot from 2026-01-04 23-25-59

The text is already there, only adding the example.

Also guards the rest of the paths with quotation marks to account for any spaces.

Closes #168

@fsimonis OK?


uekerman:
I also define `CPLUS_INCLUDE_PATH` and  `C_PATH`. Don't we then need these as well?

MakisH:
> I also define `CPLUS_INCLUDE_PATH` and `C_PATH`. Don't we then need these as well?

No, pkg-config and CMake already provide this information. On my system, I do not have these defined since a while, and I have not faced any issue with any component.

Merging this to avoid conflicts. If anything looks wrong, updates are easy.

Issue #639, opened 2025-11-25
Automatically-generated community metrics


MakisH:
A rather common question is "how many people use preCICE?". We could already visualize some facts on the website.

We currently have, on the [landing page](https://precice.org/):

- Number of stars
- [Number of citations](https://precice.org/#academic-at-heart-100-free-software) (currently broken, see #522)
- [Logowall](https://precice.org/#tried-and-tested-by-a-global-community) (updated manually)

We could also collect some of these facts in the community landing page: https://precice.org/community.html

We could embed:

- Analytics of the various repositories (core, tutorials) from GitHub. For example:
   - Number of contributors
   - Recent commit activity
   - Recent releases
   - Number of stars 
- Some analytics from [Plausible Analytics](https://plausible.io/precice.org) (analytics of our website)
- Download counts of the release artifacts. There are some third-party apps that get such data using the GitHub API ([example](https://somsubhra.github.io/github-release-stats/?username=precice&repository=precice)).
- Analytics of the Discourse forum:
   - Number of users
   - Number of threads

There is probably a lot more, and various implementation ways.

PR #684, opened 2026-01-19
Document HPC systems with preCICE pre-installed

MakisH:
Adds a table on https://precice.org/installation-special-systems.html

Reasons:

- Stop users from trouble early on
- Help admins find other admins that already managed
- Encourage more additions
- Get more usage data

Maintaining the table will be an issue, but so is the rest of this page as well, which mostly serves as a collection of clues.

This currently only includes one entry, we can probably add many more.

### DIFF
--- a/content/docs/installation/installation-special-systems.md
+++ b/content/docs/installation/installation-special-systems.md
@@ -16,6 +16,17 @@ The instructions may still be valuable for unlisted systems.
 We encourage users to actively contribute to this page! If your system is not listed, please feel encouraged to add instructions for it!
 {% endnote %}
 
+## HPC systems with preCICE pre-installed
+
+Before attempting to install preCICE yourself, check if it is already available on the system you are using.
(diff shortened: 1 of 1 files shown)


IshaanDesai:
The [SimTech Ehlers cluster](https://www.simtech.uni-stuttgart.de/labs/computecluster/) also has preCICE pre-installed, but I am not sure if it is worth mentioning, as it is very Stuttgart-specific.

fsimonis:
I like the text. The table on the other hand will be problematic as systems come and go.
We are already having issues keeping the distro instructions updated.

Unless your intention is croas-linking, I don't think it's a good idea.

MakisH:
Thanks for the detailed argumentation. I am also not sure how to best do this.

> Unless your intention is cross-linking

Yes, that's my trigger. I think it is useful in some occasions to be able to point to such a case (I wanted to refer to this in my thesis, could refer to from proposals, and maybe link from a page implementing #639), and I think that facilities/admins would appreciate it and would like to add their systems.

> The [SimTech Ehlers cluster](https://www.simtech.uni-stuttgart.de/labs/computecluster/) also has preCICE pre-installed, but I am not sure if it is worth mentioning, as it is very Stuttgart-specific.

Maybe this is more of a problem: that people will not think it is useful adding their system.

Maybe a better place could be a forum thread.

> I like the text.

Maybe we should still keep that without the table.

And the table definitely makes more sense with more entries and higher-tier systems.

PR #686, opened 2026-01-23
Remove extrapolation from cplschemes

fsimonis:
This has been removed with version 3.0.0.

https://github.com/precice/precice/pull/1738


MakisH (approved):
I see no further mentions.

PR #688, opened 2026-01-25
Add FEniCSx adapter site

NiklasVin:
The release of the FEniCSx adapter is currently prepared.
I left the structure and the contents of this site more or less identical to the one of the FEniCS adapter and just updated the corresponding information.
Additionally, I added a reference for the software update paper of the FEniCSx adapter and a _fenicsx_ tag to the original software paper such that it is listed under related literature.


IshaanDesai (commented):
I would suggest hosting the documentation in the adapter repository and linking to it. This can be done by updating submodules, via an Action workflow. Look at [this workflow](https://github.com/precice/micro-manager/blob/develop/.github/workflows/update-website.yml)

MakisH:
Documentation for rendering content from other repositories: https://precice.org/docs-meta-overview.html#rendering-content-from-external-repositories

NiklasVin:
Ok, thank you for the suggestions. I will close this PR and start over with the other approach.

PR #690, opened 2026-01-27
Add newsletter sign-up

fsimonis:
This PR adds the newsletter signup form to the landing page as part of the "join the community" section and on other pages in the footer.

## Landing page

Adding the form to the existing content looks very janky, so I created a "subsection".

image

## Footer

This doesn't show on the landing page. The form can only load once per page.

image

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,7 +3,7 @@
 
 {% if page.layout != "landing_page" %}
     <div class="row equal">
-      <div class="col-md-6" style="background:white;border-radius:1rem;padding:0 3rem;margin:2rem auto;">
+      <div class="col-md-4" style="background:white;border-radius:1rem;padding:0 3rem;margin:0 auto 50px auto;">
         <h2> The preCICE newsletter </h2>
         {% include newsletter.html %}
         </div>
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -10,7 +10,7 @@
     </div>
 {% endif %}
 
-    <div class="row equal">
+    <div class="row equal text-center">
       <div class="col-md-4">
         <nav aria-label="Footer left">
           <ul class="list-unstyled">
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,5 +1,15 @@
 <footer class="background-dark">
   <div class="container">
+
+{% if page.layout != "landing_page" %}
+    <div class="row equal">
+      <div class="col-md-6" style="background:white;border-radius:1rem;padding:0 3rem;margin:2rem auto;">
+        <h2> The preCICE newsletter </h2>
+        {% include newsletter.html %}
+        </div>
+    </div>
+{% endif %}
+
     <div class="row equal">
       <div class="col-md-4">
         <nav aria-label="Footer left">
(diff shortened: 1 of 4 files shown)


_includes/footer.html
MakisH, line 1: Spacing-wise, this is looks strange to me:

- While the current footer has left-aligned elements that follow the left boundary of the page, the form is centered. But since there is no clear reference point above or below, it looks like the elements are not aligned to each other anymore. 
- The footer has a significantly larger padding-top than the space between the three columns with links, which makes the form connect to these elements, instead.

Potential solution: introduce a "secondary footer" above the main one, in a different color (maybe light blue).

Potential extension to that: Have a two-column layout, similar to the one currently on the landing page.

Or just drop it if complicated (enough to have it on the landing page and the community channels).

content/index.html
MakisH, line 1: A problem with the current layout is that it ends up with a lot of empty space at the bottom-left of the section.

Instead of that, I would suggest:

- Making the form take the full length
- Squashing the newsletter description into one sentence, and remove the heading:

```
Subscribe to the preCICE newsletter for upcoming events, new publications, new features, and more important updates (four e-mails per year).
```

(starting with "events", to connect it to the previous part)

If this is complicated (I already tried a bit), then we could also just remove the photo and move it to the "Community" page (where it used to be).

MakisH (commented):
Good to have it on the landing page, and that is the right place.

We should also add it to https://precice.org/community-channels.html, where we could add more details and then link to that from different places.

_includes/footer.html
fsimonis: > While the current footer has left-aligned elements that follow the left boundary of the page, the form is centered. But since there is no clear reference point above or below, it looks like the elements are not aligned to each other anymore.

That's because they aren't.

I cannot control the padding of the iframe of the injected form and color wise it needs to be on a white background. So I need to put it in a white box. Not centering it looks weird.

> Potential solution: introduce a "secondary footer" above the main one, in a different color (maybe light blue).

I cannot just create a new light section in the footer because of the footer css and the background-light definitions.

I tried some ways and cannot get it look clean. This is the cleanest I managed. 

> Potential extension to that: Have a two-column layout, similar to the one currently on the landing page.

Tried this too and it needs too much text on the non-form column to create a somewhat balanced look. Too much for a footer at least.

> Or just drop it if complicated (enough to have it on the landing page and the community channels).

I think the visibility here is significantly higher than in the community channels.

Even if it looks jank at first, its better than nothing

content/index.html
fsimonis: > Making the form take the full length

On big screens, this looks super weird. It's a giant white bar. The width needs to be restricted somehow.

Replacing the picture looks way better. I used the list to balance out the height of the form.

image

_includes/footer.html
fsimonis: Adjusting the padding and shrinking it a bit helps a little bit

image

Centering the text should be enough

image

fsimonis: This looks a bit weird on mobile, but acceptable IMO

fsimonis:
> We should also add it to https://precice.org/community-channels.html, where we could add more details and then link to that from different places.

I'll open an issue for this as it isn't time-pressing and the integration may require some rewriting.

MakisH (approved):
Much better now, thanks for the updates! Both the landing page and the footer now look quite balanced.

I moved the deleted community photo to the community page.

PR #697, opened 2026-02-02
Fix inconsistent object name in Python example code

joargu:


MakisH (approved):
Good catch, thank you!

PR #701, opened 2026-02-03
Document precice-cli

### LINKED ISSUE #581: Documentation page for preCICE CLI
We don't seem to have one yet. It could be a separate page under `Tooling`, or a section under https://precice.org/tooling-overview.html

fsimonis:
This PR 

- Adds a page for the tooling-cli, linking to relevant repositories and subpages
- Updates the builtin tooling docs
- Adds the precice-cli to the tooling overview

Closes #581

### DIFF
--- a/_data/sidebars/docs_sidebar.yml
+++ b/_data/sidebars/docs_sidebar.yml
@@ -238,6 +238,10 @@ entries:
       url: /tooling-aste.html
       output: web, pdf
 
+    - title: The preCICE CLI
+      url: /tooling-cli.html
+      output: web, pdf
+
     - title: Built-in tooling
       url: /tooling-builtin.html
       output: web, pdf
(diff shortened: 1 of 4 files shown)


fsimonis:
I merged this to get the long overdue section on the website. We can iterate over the content now.

MakisH:
With a quick look, looks good. I would even suggest moving it above ASTE in the sidebar. This is a very nice tool (kudos!) that we should get at the hands of every user. I see why it fits next to the "Builtin tooling", maybe we need to move ASTE a bit down then.

uekerman:
We don't mention `precice-cli` here: https://precice.org/tooling-config-visualization.html
Would make sense there as well, no? I guess, we want to point (new) users always first to the CLI.

MakisH:
> We don't mention `precice-cli` here: https://precice.org/tooling-config-visualization.html Would make sense there as well, no? I guess, we want to point (new) users always first to the CLI.

Opened a PR to not forget this: https://github.com/precice/precice.github.io/pull/706

PR #703, opened 2026-02-03
Add hints to check the installation with precice-version

MakisH:
Adds hints in:

- Installation overview, below the graph: https://precice.org/installation-overview.html
- Building from source - installation: https://precice.org/installation-source-installation.html
- Training preparation: https://precice.org/community-training.html#individual-dependencies

Anywhere else to add it?


uekerman (approved):
Looks good. I didn't look long,  but I think you cover all meaningful places.

PR #720, opened 2026-02-20
Fix news card title truncation

### LINKED ISSUE #699: Latest news titles cut off between small mobile and large desktop breakpoints
Image

Image

tanishqio:
## Fix responsive truncation of news card titles

Closes #699

---

This PR addresses #699 by improving how news card titles are truncated across different viewport sizes.

Previously, titles were limited to **two lines regardless of screen width**, which caused overly aggressive truncation on tablet and mobile devices. On narrower viewports, this reduced readability and cut off titles prematurely.

---

### **Approach**

Introduced responsive `-webkit-line-clamp` values using media queries to allow more vertical space for titles on smaller screens:

- **Desktop (> 992px):** 2 lines  
- **Tablet (768px – 992px):** 3 lines  
- **Mobile (< 576px):** 4 lines  

This preserves a compact layout on larger screens while improving readability on smaller devices.

Additionally, the title color was explicitly set to `#000` so the ellipsis (`...`) visually matches the title text.

---

### **Why This Approach?**

- Improves readability on smaller viewports  
- Maintains consistent layout density on desktop  
- Uses a CSS-only solution (no JavaScript required)  
- Aligns with the existing responsive breakpoint structure  

---

### **Changes**

- Updated `css/landing-page.css`
- Added media queries for `.news-card h4`
- Adjusted responsive `-webkit-line-clamp` values
- Explicitly set title color to black

---

### **Testing**

Tested locally by:

- Resizing the browser viewport  
- Using DevTools device emulation  
- Verifying title rendering at different breakpoints  

The truncation now behaves consistently and improves title visibility across devices.

---

### **Screenshots**

image
image
image

### DIFF
--- a/css/landing-page.css
+++ b/css/landing-page.css
@@ -270,16 +270,16 @@ div.logo-holder > img {
   text-overflow: ellipsis;
   display: -webkit-box;
   -webkit-line-clamp: 2;
-  line-clamp: 2; 
+  line-clamp: 2;
   -webkit-box-orient: vertical;
-  height: 2.8em; 
+  color: #000;
 }
 
 .news-card p:first-of-type {
   overflow: hidden;
   text-overflow: ellipsis;
   display: -webkit-box;
-  -webkit-line-clamp: 5; 
+  -webkit-line-clamp: 5;
   line-clamp: 5;
   -webkit-box-orient: vertical;
 }
@@ -301,5 +301,20 @@ div.logo-holder > img {
   }
   #news-container .col-md-4 {
     flex: 1 1 100%;         
+  }
+}
+
+/* Responsive title line-clamp for news cards */
+@media (max-width: 992px) {
+  .news-card h4 {
+    -webkit-line-clamp: 3;
+    line-clamp: 3;
+  }
+}
+
+@media (max-width: 576px) {
+  .news-card h4 {
+    -webkit-line-clamp: 4;
+    line-clamp: 4;
   }
 }
--- a/css/landing-page.css
+++ b/css/landing-page.css
@@ -312,7 +312,7 @@ div.logo-holder > img {
   }
 }
 
-@media (max-width: 576px) {
+@media (max-width: 768px) {
   .news-card h4 {
     -webkit-line-clamp: 4;
     line-clamp: 4;


css/landing-page.css
```diff
@@ -303,3 +303,18 @@ div.logo-holder > img {
     flex: 1 1 100%;         
   }
 }
+
+/* Responsive title line-clamp for news cards */
+@media (max-width: 992px) {
```
MakisH, line 308: Why the boundaries specifically at 992px and 576px? How do these compare to other boundaries we currently have in this or other relevant CSS files?

css/landing-page.css
```diff
@@ -270,16 +270,16 @@ div.logo-holder > img {
   text-overflow: ellipsis;
   display: -webkit-box;
   -webkit-line-clamp: 2;
-  line-clamp: 2; 
+  line-clamp: 2;
   -webkit-box-orient: vertical;
-  height: 2.8em; 
+  color: #000;
```
MakisH, line 275: Self-note: This seems to indeed be needed. Without it, the dots in the title are gray.

MakisH (commented):
Thank you for this bugfix. As a reviewer, I also appreciate the screenshots. I tested it locally, and it does what it is supposed to do.

Since I understand that this is coming from a GSoC (and thus, learning) context: I would have probably made the PR description shorter (it is currently much longer than the code diff itself). That's something that takes practice, of course, and I think the guiding principle is writing from a personal perspective, like explaining to a peer. Tools might provide more details than needed, making it difficult to quickly spot the main point.

There are a few minor formatting changes as well (removing some unnecessary spaces). Since these are valid, only spaces, and only in the respective classes, that's fine for this PR.

See also the original issue (#699) for my reasoning for this PR vs #721.

css/landing-page.css
tanishqio: I chose 992px and 576px based on common responsive breakpoints. After checking the existing CSS, I see that 992px and 768px are already used, but 576px isn’t. I’ll switch to 768px to stay consistent with the current breakpoints.

tanishqio:
Thanks for the feedback and for testing it locally, I appreciate it.
I’ve updated the media queries to use the existing 768px breakpoint instead of introducing 576px.
I’ll keep future PR descriptions shorter and more to the point.

MakisH (approved):
Thank you to you and to everyone else that also tried to resolve this issue. Looks good enough to merge already, happy to review any other PR might come.

PR #726, opened 2026-02-22
Add jekyll-mermaid to conditionally render Mermaid graphs

### LINKED ISSUE #552: Support for mermaid graphs
As found in https://github.com/precice/precice.github.io/pull/498 ([page rendering](https://precice.org/configuration-coupling-multi.html)), [mermaid graphs](https://mermaid.js.org/) are currently not rendered.

This code:

```mermaid
graph LR;
    A -- first to second --> B;
    B -- first to second --> C;
    C -- first to second --> A;
```

leads to:

Image

We probably need to add a JS library.

hanzalahwaheed:
Fixes https://github.com/precice/precice.github.io/issues/552

# Feature:

Added mermaid.js support for the docs.

# Approach

Mermaid can only be loaded on the html layer. We conditionally load it only on the pages where required as it is quite heavy (approx. 800KB).

# Changes

## Before

image

## After

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,8 +1,5 @@
 repository: precice/precice.github.io_future
 
-mermaid:
-  src: 'https://cdn.jsdelivr.net/npm/mermaid@11.12.3/dist/mermaid.min.js'
-
 output: web
 # this property is useful for conditional filtering of content that is separate from the PDF.
 
@@ -263,6 +260,9 @@ algolia:
   index_name:     jekyll
   search_only_api_key: 760ed6be3e165d08b4a798ac4aa82fe4
   nodes_to_index: 'p,code,table' # html tags to include in search index
+
+mermaid:
+  src: 'https://cdn.jsdelivr.net/npm/mermaid@11.12.3/dist/mermaid.min.js'
 
 plugins:
   - jekyll-redirect-from
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,8 @@
 repository: precice/precice.github.io_future
 
+mermaid:
+  src: 'https://cdn.jsdelivr.net/npm/mermaid@11.12.3/dist/mermaid.min.js'
+
 output: web
 # this property is useful for conditional filtering of content that is separate from the PDF.
 
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ group :jekyll_plugins do
   gem 'jekyll-sitemap'
   gem 'jekyll-redirect-from'
   gem "jekyll-last-modified-at"
+  gem 'jekyll-mermaid'
 end
 
 gem 'jemoji'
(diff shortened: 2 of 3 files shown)


hanzalahwaheed:
hi @MakisH
please take a look at this. I've tried to go with the most efficient approach and following a similar commit discipline. happy to make any more changes :)

_includes/mermaid.html
```diff
@@ -0,0 +1,13 @@
+<script src="https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.min.js"></script>
```
MakisH: Just wondering: What is this CDN? Why use this one than anything else? What would the official way to get Mermaid be?

Not that the same questions does not apply to other scripts we already use, but now I can ask the question. 😅

_includes/mermaid.html
```diff
@@ -0,0 +1,13 @@
...
+    document.querySelectorAll('pre code.language-mermaid').forEach(function(elem) {
+      var div = document.createElement('div');
+      div.className = 'mermaid';
+      div.textContent = elem.textContent;
+      elem.closest('pre').replaceWith(div);
+    });
+    mermaid.initialize({ startOnLoad: false });
+    mermaid.run();
+  });
+</script>
```
MakisH: Also wondering: Is this coming from the mermaid documentation? (reference?) Or from any other source?

MakisH (commented):
Thank you for the contribution, looks indeed very nice, and it will be very useful.

I like that `mermaid.html` is a separate file, and I like that only specific pages load it.

I tested it locally and it does what it should.

_includes/mermaid.html
hanzalahwaheed: looked into this

- used the CDN cause it was being done in similar places. also it is the easiest way to add mermaid.
- another way could be to use the `jekyll-mermaid` plugin. should I go with this approach?

on second thoughts, maybe going with the plugin would be cleaner but it also uses the CDN

_includes/mermaid.html
hanzalahwaheed: this was a hack to parse mermaid renders in html. let me try the plugin approach to see it we can do this better.

_config.yml
```diff
@@ -1,5 +1,8 @@
 repository: precice/precice.github.io_future
 
+mermaid:
+  src: 'https://cdn.jsdelivr.net/npm/mermaid@11.12.3/dist/mermaid.min.js'
```
hanzalahwaheed: This is fixed to a static version `v11.12.3` now.

hanzalahwaheed:
yep!
- the `jekyll-mermaid` approach is much cleaner. 
- this fetches mermaid only on the required pages as well (gem only injects where the tag is used).
- no need of a separate `mermaid.html` as well

thanks for the kind pointers. learnt something new. do tell if this is alright :))

_config.yml
MakisH: No objection to having a pinned version. We use other dependencies with pinned versions as well.

_includes/mermaid.html
MakisH: Indeed, this looks cleaner. The downside is that it is bound to Jekyll, but this is what we have right now, so that's fine.

MakisH (approved):
Thank you for the contribution, looks good to me. We can revisit it later if/when we change the framework.

PR #732, opened 2026-02-22
Fix/news section sorting 402

Wdhanashri:
Sort news by publication date and make full card clickable                                                                                                                      
- Sort by created_at instead of last_posted_at (remaining item from #628)                                                                                                                                 
- Move date inside <a> tag so entire card is clickable                                                                                                                                                    
- Add created_at field to fetch-news.py for sort support
- Remove debug console.log                                                                                                                                                                                 - Fix typo in CSS comment
Addresses #402

### DIFF
--- a/css/landing-page.css
+++ b/css/landing-page.css
@@ -89,7 +89,7 @@ div.image-holder.usp > img {
 }
 
 /* Adapter section */
-/* adapter class also used for news in the langing page */
+/* adapter class also used for news in the landing page */
 a.adapter {
   text-decoration: none;
 }
@@ -263,6 +263,10 @@ div.logo-holder > img {
   padding: 20px;
   margin-bottom: 20px;
   box-shadow: 0 2px 4px rgba(0,0,0,0.05);
+}
+
+.news-card a.adapter {
+  display: block;
 }
 
 .news-card h4 {
--- a/css/landing-page.css
+++ b/css/landing-page.css
@@ -265,8 +265,15 @@ div.logo-holder > img {
   box-shadow: 0 2px 4px rgba(0,0,0,0.05);
 }
 
-.news-card a.adapter {
+.news-card a.news-link {
   display: block;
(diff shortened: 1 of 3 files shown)


css/landing-page.css
```diff
@@ -265,21 +265,25 @@ div.logo-holder > img {
   box-shadow: 0 2px 4px rgba(0,0,0,0.05);
 }
 
+.news-card a.adapter {
```
MakisH: Since you are touching this, maybe you could introduce a different class than `adapter`, since that was originally meant for something else.

js/news-collect.js
```diff
@@ -12,29 +10,28 @@ document.addEventListener("DOMContentLoaded", async function () {
 
     if (!topics.length) throw new Error("No topics found");
 
-    topics.sort((a, b) => new Date(b.last_posted_at) - new Date(a.last_posted_at));
+    // Sort by publication date instead of last activity
+    topics.sort((a, b) => new Date(b.created_at || b.last_posted_at) - new Date(a.created_at || a.last_posted_at));
```
MakisH, line 13: Why the OR `||` operator? The comment should also not be needed, since the "last activity" is not there anymore, right?

MakisH (commented):
Thank you for the contribution, @Wdhanashri! Useful cleanup as well.

A screenshot comparing the current version to this one:

Screenshot from 2026-02-22 20-48-38

js/news-collect.js
Wdhanashri: The || fallback is just temporary since created_at isn’t in news.json yet. It should be added after the next CI run of fetch-news.py, and once that happens the fallback won’t have any effect.

css/landing-page.css
Wdhanashri: Replaced adapter with a dedicated news-link class with its own styles.

MakisH (approved):

Issue #472, opened 2024-11-13
Possible migration from Jekyll to Hugo


fsimonis:
Whilst looking for alternatives to Jekyll I noticed that Hugo has many features that makes it very suitable for a replacement in our case.

Documentation under https://gohugo.io/documentation/

I gather Pro, contra, and unclear points underneath:

### Installation

Jekyll and its plugins are based on ruby.

Hugo is a linked executable which can be downloaded and installed directly from their repo.

### Subprojects

Subprojects/including other repositories are not supported by Jekyll.
The current workaround is a custom plugin that imports data from submodules.

Hugo supports additional content directories to cover submodules and even allows exporting data from external repositories at build-time using [hugo module](https://gohugo.io/hugo-modules/use-modules/).

We currently use `docs` folders in our repositories, so we would need to [mount them](https://gohugo.io/hugo-modules/configuration/#module-configuration-mounts) into the content directory. 

I am not sure if this requires changes in the imported repositories though.

### Menus

Currently handled using `sitebar` data and custom logic and templating.

Hugo has build-in support for [menus](https://gohugo.io/content-management/menus/).
Sites can define where they are located and menus can be rendered differently in various places.
We only would have to customize the rendering template.

### Git info

Currently handled using an unmaintained plugin which provides the last-modified date.

Hugo has built-in support for [git info](https://gohugo.io/methods/page/gitinfo/) including last commit of a page, author and also all authors of a page.

### SASS

Currently handled by the `jekyll-sass-converter` which is a direct dependency of `jekyll` and suffers from the problem in #471

Hugo directly uses dart-sass, which is essentially what the newer `google-protobuf` version of the converter does.

### Algolia

Currently handled by the abandoned `jekyll-algolia` gem, which has a flaky dependency and produces overly long records.

The Hugo website itself uses Algolia though, so there is a way to index hugo projects in some way.

### Paper citations

Currently extracted from Google Scholar which often fails due to rate limiting.

Hugo supports [external data source](https://gohugo.io/content-management/data-sources/), meaning we can keep this information updated somewhere else and import it when needed.

### Image handling

Currently there is no asset pipeline. Image will be included in the website as is.

Hugo has an asset pipeline that can resize images to improve the experience on mobile.

### Basic SCHEMA metadata

We currently generate some schema metadata using jekyll templates.
Hugo renders some basic schema metadata by default.

### Template engine

Jekyll uses the Liquid template engine.

Hugo uses Go templating, which requires porting.

[...]

PR #736, opened 2026-02-22
feat: add dark mode toggle button

aayushi-singhhh:
- Add css/dark-mode.css with full dark mode overrides (backgrounds, text, navbar, panels, code, tables, footer, strong/emphasis yellow)
- Add js/dark-mode.js for theme toggle and animated stats counter
- Update _includes/head.html to load dark-mode CSS/JS and prevent flash-of-unstyled-content on page load
- Add dark mode toggle button to _includes/topnav.html
- Add stats banner and feature cards sections to content/index.html

https://github.com/user-attachments/assets/6f711bed-db47-4be7-b8a0-4848fb32ad95

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -104,3 +104,13 @@ href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.7.2/css/all.min.css"
 
 <!-- Plausible Analytics -->
 <script async defer data-domain="precice.org" src="https://plausible.io/js/plausible.js"></script>
+
+<!-- Dark Mode: apply theme before paint to avoid flash -->
+<script>
+  (function(){
+    var t = localStorage.getItem('precice-theme') || 'light';
+    document.documentElement.setAttribute('data-theme', t);
+  })();
+</script>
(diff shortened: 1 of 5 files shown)


MakisH:
Thank you for the contribution, @aayushi-singhhh. It is interesting to see how far we are from supporting a dark theme, and the video makes that easier.

I don't see how this PR could achieve a proper dark theme state without a lot of additional work from both your side and my side. At the same time, in the context of a GSoC project (which I understand you are targeting), we are planning to transition to a new framework (see #472), which would give us a dark theme at a much lower effort. So, any work on adding a dark theme to the current framework would be eventually wasted.

I am closing this PR, and I think we are quickly running out of issues to fix, before even the GSoC applications open!

aayushi-singhhh:
Thank you for the clarification. Investing effort into dark mode on the current stack would not be ideal if the framework transition is planned. I’ll shift my focus toward the Bootstrap upgrade and exploring the Hugo migration feasibility instead. Could you guide me what issues I can target to solve please?

PR #737, opened 2026-02-22
Enable scrolling in search results dropdown

### LINKED ISSUE #403: Scrolling in Algolia (or other full results view)
One old issue we have is that one can only see the first few results in the Algolia search bar.
Not sure how to fix it, but it is an issue.

@chlorenz you already tried fixing this in the past, right?

Wdhanashri:
Changed overflow from hidden to auto on the search results container so users can scroll through all results instead of being cut off.

Fixes #403


MakisH (approved):
Thank you for another useful contribution! 😄 

That was quite a long-standing annoying issue with in the end such a small fix. 🙈 

I have confirmed that it works locally. I don't see anything related in the history that would show that this was previously tried and rejected.

Wdhanashri:
Thanks for verifying and for the kind words!

PR #740, opened 2026-02-22
feat: add copy-to-clipboard button on all code blocks

aayushi-singhhh:
Adds a 'Copy' button to every Rouge-highlighted code block across the entire documentation and tutorial sections of the site.

Changes
-------

js/copy-code.js (new file)
- Pure vanilla JS, zero dependencies, IIFE-scoped
- Targets all div.highlight > pre blocks (Rouge output) and bare pre > code blocks not wrapped by the highlighter
- Uses the modern async navigator.clipboard.writeText() API
- Falls back to a hidden textarea + document.execCommand('copy') for non-HTTPS pages and older browsers
- Button shows 'Copied!' (green) on success, 'Failed' (red) on error, auto-resets to 'Copy' after 2 seconds
- Button is disabled during the async copy to prevent double-clicks
- Guards against duplicate buttons if attachCopyButtons() is called twice

css/customstyles-precice.css (appended, ~65 lines)
- .copy-code-btn: absolutely positioned top-right of the code block, hidden (opacity:0) by default, fades in on div.highlight:hover
- Visible on :focus for keyboard navigation
- @media (hover: none): always visible on touch/mobile devices
- .copy-code-btn--success: green feedback (#d4edda / #28a745)
- .copy-code-btn--error:   red feedback   (#f8d7da / #dc3545)
- .copy-code-btn:disabled: default cursor while copy is in-flight

_includes/head.html (1 line added)
- Loads js/copy-code.js after customscripts.js on every page

Why this matters
----------------
preCICE documentation is heavily code-heavy (C++, Python, XML config, shell commands). Users currently must manually select and copy code, which is error-prone for long config blocks. This brings the site's DX in line with ReadTheDocs, GitHub, and MkDocs Material, all of which provide copy buttons as a baseline feature.

No dependencies added. No existing files modified beyond customstyles-precice.css (append-only) and head.html (1 line). Build output verified: jekyll build completes with zero errors/warnings.
Screenshot 2026-02-23 at 4 32 32 AM

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -80,6 +80,7 @@ href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.7.2/css/all.min.css"
 <script src="js/jquery.navgoco.min.js"></script>
 <script src="{{ "js/toc.js" }}"></script>
 <script src="{{ "js/customscripts.js" }}"></script>
+<script src="{{ "js/copy-code.js" }}"></script>
 
 <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
 <!-- WARNING: Respond.js doesn't work if you view the page via file:// -->
--- a/css/customstyles-precice.css
+++ b/css/customstyles-precice.css
@@ -376,3 +376,67 @@ details > summary::after {
 details[open] > summary::after {
   content: "▲";
 }
+
+/* ── Copy-code button ────────────────────────────────────────────────────────
+   Appears on hover over any Rouge-highlighted code block.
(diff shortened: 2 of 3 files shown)


MakisH:
Thank you for your contribution!

This adds some new scripts, which would not be needed once we move to a better framework (#472), which would give a better integrated button very easily.

Since this would now be additional code to maintain, and since it still would need some theming and further review to get it right, I think we should not focus our energy into this direction right now.

In any case, it's a good idea, and I hope you learned something while making it.

Issue #733, opened 2026-02-22
Search for ‘GSoC’ Returns Irrelevant XML Documentation Pages


MuhammadAashirAslam:
# Description

When using the search on the preCICE website, short queries like “GSoC” sometimes match unrelated XML code snippets or `<m2n:sockets>` examples. This happens because Algolia’s prefix matching and typo tolerance treat short words as potential matches even if they’re completely unrelated.  

For example, “gsoc” can match part of the word “sockets” in `<m2n:sockets>` due to 2-typo allowance and prefix matching, producing false positives at the top of search results.  

# Steps to Reproduce

1. Go to the preCICE website search bar.  
2. Enter **“GSoC”** as the query.  
3. Observe the results:  
   - XML reference pages or inline code blocks appear before the actual GSoC documentation page.  
   - The match comes from code snippets, not page titles or keywords.  

# Expected Behavior

- Queries should match **actual content pages** rather than short code tokens.  
- Pages like the GSoC documentation should appear first.  
- Irrelevant XML or code-heavy pages should not dominate the search results.  

### Screen Shot 

<img width="1899" height="974" alt="Image" src="https://github.com/user-attachments/assets/ef7554c0-ad3d-466c-a297-61c1d4956bf8" />

MuhammadAashirAslam:
So I was trying to solve this issue locally and found out that the problem was caused by Algolia's default **typo tolerance**. By default, it allows **1 typo for words with 4 or more characters**. That's why when I searched for `"gsoc"` (4 characters), it was matching parts of `"sockets"` inside `<m2n:sockets>` XML code blocks. Basically, Algolia was treating `"gsoc"` as a 2-typo match for `"soc"` in `"sockets"`, which is completely unrelated.

I confirmed this by checking the ranking info from the Algolia API:

```json
"_rankingInfo": {
    "nbTypos": 2,
    "nbExactWords": 0,
    "firstMatchedWord": 2000
}
```

Zero exact word matches and 2 typos — a pure false match coming from code content.

## Files Changed

- `_includes/algolia.html`
- `js/algolia-search.js`

Both files had the same change inside the `searchFunction`.

## Code Changes

```diff
  searchFunction: function(helper) {
      var searchResults = $('#search-results');
      if (helper.state.query === '') {
          searchResults.hide();
          return;
      }

+     // Prevent short queries like "gsoc" from matching unrelated code
+     helper.setQueryParameter('minWordSizefor1Typo', 5);
[...]

Issue #388, opened 2024-04-24
Updating the Algolia search index: Record is too big


MakisH:
Trying to update our Algolia search index, I got the following error:

```
Rendering to HTML (100%) |====================================================================================================================================================================================|
[✗ Error] Record is too big                                                      =====================================================================                                                        |
                                                                                 
The jekyll-algolia plugin detected that one of your records exceeds the 10.00 Kb 
record size limit.                                                               
                                                                                 
title:    Overview of adapters                                                   
url:      /adapters-overview.html                                                
size:     9.98 Kb                                                                
                                                                                 
Most probable keys causing the issue:                                            
   html (6.84 Kb), content (2.29 Kb), keywords (0.06 Kb)                         
                                                                                 
Complete log of the record has been extracted to:                                
   /home/gc/repos/precice/website/jekyll-algolia-record-too-big.log              
                                                                                 
This issue can be caused by malformed HTML preventing the parser to correctly    
grab the content of the nodes. Double check that the page actually renders       
correctly with a regular `jekyll build`.                                         
                                                                                 
You can also exclude the page generating this error from the indexing by editing 
the `files_to_exclude` key of your config.                                       
                                                                                 
If you think this is an error and your current Algolia plan should allow you to  
push records bigger than 10.00 Kb, you can change the `max_record_size` config   
option to increase the limit. Paid plans have a limit set to 20Kb, while free    
Community plans have it set to 10Kb.                                             
                                                                                 
The following documentation might help you:                                      
   - https://community.algolia.com/jekyll-algolia/options.html#files-to-exclude  
   - https://community.algolia.com/jekyll-algolia/options.html#nodes-to-index    
   - https://community.algolia.com/jekyll-algolia/options.html#max-record-size   
```

Log: [jekyll-algolia-record-too-big.log](https://github.com/precice/precice.github.io/files/15097763/jekyll-algolia-record-too-big.log) (9.98 Kb, on the 10 Kb limit)

Trying to reduce the size of the (borderline) `adapters-overview.html`, I get the same issue with another file: [jekyll-algolia-record-too-big.log](https://github.com/precice/precice.github.io/files/15097780/jekyll-algolia-record-too-big.log) (17.59 Kb, over the 10 Kb limit, corresponding to [this file](https://github.com/precice/precice.github.io/blob/052ee40530dd4dbc8565c6bda53f424cc6c6bfb0/pages/docs/dev-docs/dev-publications.md)), which probably is indeed a bit too complex.

[...]

PR #741, opened 2026-02-22
Improve short-query precision in Algolia search

Flamki:
Summary
Improve short-query precision in Algolia by applying a minimal query-time typo-threshold tweak in both Algolia entry points.

What changed
- `_includes/algolia.html`
- `js/algolia-search.js`

Both now set:
- `minWordSizefor1Typo = 5`
- `minWordSizefor2Typos = 9`

This is intentionally unconditional and minimal. Algolia applies these as per-word thresholds, so the behavior is naturally scoped to short words without extra helper logic.

Why
This reduces false positives for short queries (for example `gsoc` matching unrelated XML tokens) while avoiding broader search/index configuration changes.

Context
- The indexing/root-cause workflow typo has already been fixed separately in `#813`.
- This PR focuses only on query-time typo-tolerance behavior in the frontend search code.

Validation
- `pre-commit run --files _includes/algolia.html js/algolia-search.js` (passed)
- Attempted Docker build:
  - `docker run --rm -v "${PWD}:/srv/jekyll" -w /srv/jekyll jekyll/jekyll:4 bash -lc "bundle exec jekyll build"`
  - Could not run in this environment because Docker daemon is not available (`dockerDesktopLinuxEngine` pipe missing).

Related: #733

### DIFF
--- a/_includes/algolia.html
+++ b/_includes/algolia.html
@@ -9,6 +9,8 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+    helper.setQueryParameter('minWordSizefor1Typo', 5);
+    helper.setQueryParameter('minWordSizefor2Typos', 9);
     helper.search();
     searchResults.show();
   }
--- a/_includes/algolia.html
+++ b/_includes/algolia.html
@@ -9,6 +9,8 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+    // Raise typo thresholds so short words (<=4 chars) like "gsoc" require
+    // exact matches, preventing fuzzy hits against unrelated XML tokens.
     helper.setQueryParameter('minWordSizefor1Typo', 5);
     helper.setQueryParameter('minWordSizefor2Typos', 9);
     helper.search();
--- a/js/algolia-search.js
+++ b/js/algolia-search.js
@@ -9,6 +9,8 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+    helper.setQueryParameter('minWordSizefor1Typo', 5);
+    helper.setQueryParameter('minWordSizefor2Typos', 9);
     helper.search();
     searchResults.show();
   }
--- a/js/algolia-search.js
+++ b/js/algolia-search.js
@@ -9,6 +9,8 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+    // Raise typo thresholds so short words (<=4 chars) like "gsoc" require
+    // exact matches, preventing fuzzy hits against unrelated XML tokens.
     helper.setQueryParameter('minWordSizefor1Typo', 5);
     helper.setQueryParameter('minWordSizefor2Typos', 9);
     helper.search();


MuhammadAashirAslam:
Hi @Flamki, thanks for submitting the PR! I just have one question about the scope of the fix.

The `getStrictShortTokens` function only activates for **single-word alphanumeric queries between 3-5 characters**. What happens in a multi-word query like `"gsoc projects"` where the short token is part of a longer search?

Since `tokens.length !== 1` returns early with an empty array, `disableTypoToleranceOnWords` would receive `[]` and the typo tolerance would remain at default — meaning `"gsoc"` could still match `"sockets"` in that context.

Was this intentional to keep the fix narrow?

(Also can you attach the screenshot of your solution working locally)

Also can you see my PR #744

Flamki:
﻿Thanks for the careful review, great catch.

You were right: the earlier helper was too narrow for multi-word input. I updated the PR so short tokens are handled even when part of a longer query (for example `gsoc projects`).

Follow-up changes in this PR:
- Detect short alphanumeric tokens (3-5 chars) anywhere in the query, not only single-word queries.
- Use an Algolia-compatible query-time approach:
  - if a short token exists: `minWordSizefor1Typo=5`, `minWordSizefor2Typos=9`
  - otherwise: defaults `4` / `8`

This keeps the scope narrow (only short-token queries), avoids global index/config changes, and keeps XML search behavior for normal XML queries.

Validation:
- `pre-commit run --files _includes/algolia.html js/algolia-search.js`
- `docker run --rm -v "${PWD}:/srv/jekyll" -w /srv/jekyll jekyll/jekyll:4 bash -lc "bundle install && bundle exec jekyll build"`

Local screenshots:

1) `gsoc` (no noisy XML matches)

local-gsoc

2) `sockets` (XML results still available)

local-sockets

Flamki:
Also yes, I saw #744. The main difference is scope: #744 applies typo-threshold changes globally, while this PR applies stricter thresholds only when a short token is present in the query.

MuhammadAashirAslam:
Hey @Flamki , thanks for iterating on this! One thing I wanted to point out about the updated approach:

`minWordSizefor1Typo` and `minWordSizefor2Typos` are **per-word thresholds**, not per-query settings. Setting `minWordSizefor1Typo = 5` only affects words with 4 or fewer characters. If the query has no such words, the setting has no effect compared to the default of 4.

This means the conditional check via `hasStrictShortToken` produces **identical results** to always setting the values. For example, searching `"configuration"` (13 chars) gets 2 typos regardless of whether `minWordSizefor1Typo` is 4 or 5, because 13 > 5 either way.

So the helper function adds ~20 lines of logic that doesn't change any search behavior compared to the simpler unconditional approach.

I think the minimal 2-line version keeps things cleaner and easier to maintain. Happy to discuss if I'm missing something though! 🙂

Flamki:
@MakisH @MuhammadAashirAslam 
I simplified this PR by removing the helper/conditional logic entirely.

It now unconditionally sets `minWordSizefor1Typo = 5` and `minWordSizefor2Typos = 9` at query time in both `_includes/algolia.html` and `js/algolia-search.js`.

I also added a short inline comment in both files to explain why these values are used (to reduce short-query false positives like `gsoc` matching unrelated XML tokens).

Note: the indexing root-cause fix (typo in `update-algolia.yml`) was already merged separately in #813, so this PR only handles the typo-tolerance side.

Validation update: I ran a full local Jekyll build in Docker.

Command:
`docker run --rm -v "${PWD}:/srv/jekyll" -w /srv/jekyll jekyll/jekyll:4 bash -lc "bundle install && bundle exec jekyll build"`

Result:
`bundle exec jekyll build` completed successfully (`done in 27.344 seconds`).

MuhammadAashirAslam:
Hey @Flamki I see you have implemented exactly the same solution as my PR but the issue was something else (already working on it and found the root cause). This is something for after that so for now you can look into other issues. Thanks 😊

Flamki:
@MakisH @MuhammadAashirAslam thanks! 
Since #733 was resolved and closed by #813, this PR is no longer addressing the primary issue.

It now serves as a small hardening improvement: adding query-time typo threshold tuning in _includes/algolia.html and js/algolia-search.js (minWordSizefor1Typo=5, minWordSizefor2Typos=9) to reduce fuzzy false positives for short tokens.

I’m happy to keep this as an optional follow-up enhancement, or close it if you prefer to keep things minimal post-#813. Let me know your preference.

MakisH:
Will look into this after we manage to update the Algolia record again (see #388).

Let's see if this is still needed then (happy to merge it if yes).

MakisH:
Regarding the original issue, this has now been solved. I would prefer to leave the settings as-is for now, unless there is a separate good reason to change them.

Thanks a lot for contributing anyway!

Flamki:
Thank you for the clarification and for taking the time to review this thoroughly.

That makes sense — since the original issue has now been resolved and there isn’t a strong need to adjust the typo thresholds at this point, I’m happy to leave the current settings unchanged and close this PR.

I appreciate the feedback and the discussion. It was helpful to dig into how Algolia’s per-word typo thresholds behave in practice.

Thanks again, and I’ll look into other issues where I can contribute.

PR #744, opened 2026-02-23
Fix irrelevant search results for short queries like GSoC

### LINKED ISSUE #733: Search for ‘GSoC’ Returns Irrelevant XML Documentation Pages

MuhammadAashirAslam:
Fixes #733

### Problem

Short queries like `gsoc` were returning irrelevant XML reference results due to Algolia's default typo tolerance. For example, `gsoc` was fuzzy-matching `sockets` in XML code content.

### Fix

Added stricter typo tolerance parameters at query time in includes/algolia.html and js/algolia-search.js, so words with 4 or fewer characters require exact matches. This is a client-side-only change with no indexing, config, or content modifications.

> [!NOTE]
> Making "GSoC" fully discoverable would additionally require adding `keywords` to `searchableAttributes` in config.yml and re-indexing.

### Screenshots

**Before**

image

**After**

image

### DIFF
--- a/_includes/algolia.html
+++ b/_includes/algolia.html
@@ -9,6 +9,13 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+
+    // Increase minimum word sizes for typo tolerance to prevent
+    // short queries like "gsoc" from matching unrelated code tokens
+    // (e.g., "gsoc" was matching "sockets" via prefix with 2 typos)
+    helper.setQueryParameter('minWordSizefor1Typo', 5);
+    helper.setQueryParameter('minWordSizefor2Typos', 9);
+
     helper.search();
     searchResults.show();
   }
--- a/js/algolia-search.js
+++ b/js/algolia-search.js
@@ -9,6 +9,13 @@ const algoliaSearch = instantsearch({
       searchResults.hide();
       return;
     }
+
+    // Increase minimum word sizes for typo tolerance to prevent
+    // short queries like "gsoc" from matching unrelated code tokens
+    // (e.g., "gsoc" was matching "sockets" via prefix with 2 typos)
+    helper.setQueryParameter('minWordSizefor1Typo', 5);
+    helper.setQueryParameter('minWordSizefor2Typos', 9);
+
     helper.search();
     searchResults.show();
   }


MakisH:
Thank you for your contribution. However, I am confused. Testing locally, it does not seem to have any effect, and the screenshots don't demonstrate the expected result.

General note: the PR description is quite long for what the change is (I guess suggested by some tool). This makes it tricky to see what is going on.

I see that there is a similar discussion in #741. I am not going to interact with these PRs for now, and let the discussion converge.

MuhammadAashirAslam:
Hi @MakisH, I was trying to solve this and came to know there is a much easier solution. First of all the "Google Summer of Code" page is not indexed with Algolia, searching for it returns nothing. You would need to re-index it with bundle exec jekyll algolia using the Admin API key. As for the XML issue, just changing the heading from Google Summer of Code to Google Summer of Code (GSoC) would fix it since headings is already a [default searchable attribute](https://github.com/algolia/jekyll-algolia/blob/master/lib/jekyll/algolia/configurator.rb#L20-L25) in jekyll-algolia, so "GSoC" would match directly instead of fuzzy-matching XML tokens

MakisH:
> Hi @MakisH, I was trying to solve this and came to know there is a much easier solution. First of all the "Google Summer of Code" page is not indexed with Algolia, searching for it returns nothing. You would need to re-index it with bundle exec jekyll algolia using the Admin API key. 

Hm, the index is updated nightly (and has already been updated many times since the page was added), but it is indeed not showing up. Any other ideas?

> As for the XML issue, just changing the heading from Google Summer of Code to Google Summer of Code (GSoC) would fix it since headings is already a [default searchable attribute](https://github.com/algolia/jekyll-algolia/blob/master/lib/jekyll/algolia/configurator.rb#L20-L25) in jekyll-algolia, so "GSoC" would match directly instead of fuzzy-matching XML tokens

Good point, feel free to extend it.

MuhammadAashirAslam:
@MakisH I think I have found the culprit. In [update-algolia.yml](https://github.com/precice/precice.github.io/blob/master/.github/workflows/update-algolia.yml) line 25, the variable is named LASTEST_COMMIT but line 27 references LATEST_COMMIT. This typo means TIME_DIFF is computed against an empty variable, which could cause the indexing step to be skipped unexpectedly. Should I open a new PR to fix this?

MakisH:
Yes, please!

Issue #538, opened 2025-06-24
Content files restructure for compatibility with Hugo


fsimonis:
The hugo migration will become significantly less hacky if we bring the current content into a hierarchical format that reflects the page structure.
Moving them around in jekyll should not be an issue, as we use permalinks in all pages.

To use the navigation features, we should use the navigation hierarchy as the content structure with folders:

- quickstart
- docs
- tutorials
- community
- about

IIRC some things cannot be directly mapped to the hierarchical structure, such as https://precice.org/docs.html which should be docs/index.html but really is docs/fundamentals/index.html.

Furthermore, some sections in the docs don't have an overview page.

Finally, it is still unclear to me how to handle the nested sections such as https://precice.org/configuration-mapping.html which should be docs/configuration/basics/mapping. Such subsections may need some overview pages too.

fsimonis:
The structure we should aim for is shown here: https://gohugo.io/content-management/organization/#organization-of-content-source

MakisH:
Are you only talking about moving around files, or also changing the structure of the rendered content?

fsimonis:
For preparing the migration I am talking about changing the file layout of the sources only.

The consistent use of permalinks should result in identical rendering of the page.

fsimonis:
I did an initial restructuring in #580.

There are still some todos:
* the filenames need to be cleaned up as they will be the page names in the output. Prefixes need to be removed.
* sections need `_index.X` files. These can be empty or contain the full overview, but they need to exist for hugo to recognise folders as sections.

hanzalahwaheed:
@MakisH hey! working on this will build crucial context for the coming migration. is this worth picking up right now? will require some assistance here

MakisH:
@hanzalahwaheed this is something that we wanted to have for the GSoC idea. We also expected to work in the summer on this, and did not expect such an influx of contributions while there is still snow outside. Neither me nor @fsimonis is ready to supervise this or any other large project before summer.

hanzalahwaheed:
@MakisH  Yeah, that makes sense. Sorry to be a pain haha!

In the meantime, I’ll keep looking for smaller improvements and places where I can help build context. Also happy to assist in other ways, maybe helping with issue triage, improving contributor docs, or organising things so handling the increased contributor traffic is easier.

Some obvious suggestions:
- `CONTRIBUTING.md` in the repo can be introduced.
- `help-wanted` tags can be added for issues open for contribution.
[...]

PR #752, opened 2026-02-23
Rename documentation files to remove redundant prefixes (#589)

pratapsisodiya:
Resolves #589

### What this does
This PR performs a major structural cleanup of the `content/docs/` directory by removing redundant folder prefixes from filenames (e.g., `installation/installation-overview.md` -> `installation/overview.md`).

### Why this is high impact
- **Hugo Migration Readiness:** This satisfies a critical requirement for the Hugo migration project (#588), moving the documentation into a clean hierarchical format.
- **Improved Maintainability:** Removes redundant naming that made the filesystem difficult to navigate.
- **Zero-Risk Transformation:** Since every markdown file already uses a static `permalink` in its frontmatter, the actual website URLs remain unchanged. No existing links will break.

### Implementation Details
- Used `git mv` across the following directories to preserve full file history:
  - `configuration/`, `couple-your-code/`, `installation/`, `running/`, `tooling/`, `dev-docs/`, `fundamentals/`.
- Verified the Jekyll build locally to ensure integrity.

PR #754, opened 2026-02-23
Cleanup docs source filenames for Hugo migration (batch 1)

Flamki:
﻿Summary
Follow-up for #538: clean source filenames by removing section prefixes in selected docs sections.

Scope of this batch
- `content/docs/fundamentals/*`
- `content/docs/running/*`
- `content/docs/tooling/*`
- `content/docs/docs-meta/*`

What changed
- Renamed files from prefixed names (for example `fundamentals-overview.md`) to section-local names (for example `overview.md`).
- No content edits.
- No permalink/frontmatter changes.

Why this is safe
- Jekyll output paths remain unchanged because existing explicit permalinks are kept.
- This incrementally prepares Hugo-compatible naming without changing rendered URLs.

Validation
- `pre-commit run --files <renamed files>` (passed)
- `docker run --rm -v "${PWD}:/srv/jekyll" -w /srv/jekyll jekyll/jekyll:4 bash -lc "bundle install && bundle exec jekyll build"` (passed)

Relates to #538
Parent context: #472


MakisH:
As I also wrote in https://github.com/precice/precice.github.io/pull/752 that also addresses this:

Thank you for the contribution. We can only review this later, as this is work intended for the framework migration project and capacity right now is limited.

Not sure right now if this is a good idea.

MakisH:
Closing for the same reason as https://github.com/precice/precice.github.io/pull/752

PR #755, opened 2026-02-23
Fix dead code in VersionBlock#released? — add missing return

### LINKED ISSUE #745: VersionBlock#released? guard clause does not return (dead code)
## Description

The early-exit guard for empty `@version` does not return, so the method always falls through to the `Gem::Version` comparison.

```ruby
if @version.nil? || @version.empty?
  true  # value is discarded
end
```

This contradicts the comment and makes the guard dead code.

There is no current behavior change, because
Gem::Version.new("") evaluates to 0, so the comparison still returns true.

### Fix

```Add an explicit return:
if @version.nil? || @version.empty?
-  true
+  return true
end

```

Happy to open a PR if this is acceptable @MakisH .

SnehaChaursia:
close #745 
The early-exit guard for an empty `@version` does not return, so the method always falls through to the `Gem::Version` comparison.
```ruby
if @version.nil? || @version.empty?
  true  # value is discarded — method continues
end
```
This contradicts the comment and makes the guard dead code.

There is no current behaviour change, because Gem::Version.new("") evaluates to 0, so the comparison still returns true.


MakisH (approved):
Thanks!

PR #757, opened 2026-02-23
Fix homepage buttons missing target=_blank and Github typo

### LINKED ISSUE #749: Homepage buttons open GitHub in the same tab instead of a new one
### What's wrong
The Latest release and Star on GitHub buttons on the homepage are missing target="_blank", so clicking them navigates the user away from precice.org. Every other external link in the navbar (Discourse, LinkedIn, Mastodon, YouTube) already opens in a new tab — these two are inconsistent.

Also a small typo: the button says "Star on Github" instead of "Star on GitHub".

If you allow I would like to resolve this small issue @MakisH  .

SnehaChaursia:
The 'Latest release' and 'Star on GitHub' buttons are dynamically generated in js/github-queries.js. Both link to external GitHub URLs but were missing target="_blank" and rel="noopener noreferrer", causing clicks to navigate users away from precice.org.

All other external links in the navbar (Discourse, LinkedIn, Mastodon, YouTube) already open in a new tab. This change makes the two CTA buttons consistent with the rest of the navigation.

Also fixes a capitalization typo: 'Star on Github' -> 'Star on GitHub'.
closes #749


MakisH (approved):
Thanks!

PR #758, opened 2026-02-23
perf: lazy-load logo wall images (#751)

### LINKED ISSUE #751: Logo wall images are not lazy-loaded (67 off-screen requests on page load)
## What's wrong
The homepage renders ~67 partner/institution logos via a Liquid loop (`content/index.html`, line ~518).  
The logo wall is near the bottom of a long page, but all image requests fire immediately on load — none of the `` tags use `loading="lazy"`.

This causes 67 HTTP requests to compete with above-the-fold resources (hero image, CSS, JS), which can slow the initial page load, especially on mobile.

## Fix
A single attribute change in the template applies to all logos:

```diff
<img class="img-responsive center-block"
     src="images/logo/{{logo.logo}}"
    alt="{{logo.name}}, {{logo.country}}">
    alt="{{logo.name}}, {{logo.country}}"
    loading="lazy">
```

Happy to open a PR with this change if this sounds good. @MakisH

praaatap:
Closes #751

## What

Adds `loading="lazy"` to the ~67 logo wall `` tags on the homepage.

## Why

All logo images fire HTTP requests immediately on page load, even though the logo wall is near the bottom of a long page. This competes with above-the-fold resources (hero, CSS, JS) and hurts initial load performance — especially on mobile.

## How

Single attribute addition in [content/index.html](cci:7://file:///d:/top/gsoc/precice.github.io/content/index.html:0:0-0:0):

```diff
- <img class="img-responsive center-block" src="images/logo/{{logo.logo}}" alt="{{logo.name}}, {{logo.country}}">
+ <img class="img-responsive center-block" src="images/logo/{{logo.logo}}" alt="{{logo.name}}, {{logo.country}}" loading="lazy">

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -515,7 +515,7 @@ layout: landing_page
         <div class="col-md-3 col-sm-3 col-xs-3 text-center"> <!--  vertical-align -->
           <a href="{{logo.link}}" class="logo no-icon">
             <div class="logo-holder vertical-align "> <!-- center-block -->
-              <img class="img-responsive center-block" src="images/logo/{{logo.logo}}" alt="{{logo.name}}, {{logo.country}}">
+              <img class="img-responsive center-block" src="images/logo/{{logo.logo}}" alt="{{logo.name}}, {{logo.country}}" loading="lazy">
             </div>
             <p class="institution">{{logo.name}}</p>
             <p class="country">{{logo.country}}</p>


MakisH:
Thanks for the contribution! https://github.com/precice/precice.github.io/pull/759 also implemented the same essentially at the same time, with the same way, from the person that reported the issue.

In any case, good that it is fixed now!

pratapsisodiya:
Got it @MakisH

PR #763, opened 2026-02-23
Porting guide v2-v3: Fix getMaxTimeStepSize() typo

carme-hp:
`participant::getMaxTimeWindowSize(`) is not a method in preCICE. I believe this was meant to be `participant::getMaxTimeStepSize() `as in the example above.


MakisH (approved):
Good catch! This seems to be the only instance of this.

PR #771, opened 2026-02-23
Fix code placed after <body>

praaatap:
### Description
This PR fixes invalid HTML structure in [_layouts/default.html](cci:7://file:///d:/top/gsoc/precice.github.io/_layouts/default.html:0:0-0:0) and [_layouts/landing_page.html](cci:7://file:///d:/top/gsoc/precice.github.io/_layouts/landing_page.html:0:0-0:0). 

Previously, the [google_analytics.html](cci:7://file:///d:/top/gsoc/precice.github.io/_includes/google_analytics.html:0:0-0:0) include block was placed *after* the closing `</body>` tag (between `</body>` and `</html>`). This is invalid HTML, as all visible and functional blocks (like analytics scripts) must be contained within the `<body>` or `<head>` tags.

This PR moves the block directly *above* the closing `</body>` tag to ensure the output is valid HTML and conforms to standard web best practices.

### Fixes
Fixes #<ISSUE_NUMBER_HERE>
*(If you did not make an issue for this, you can just delete this line).*


_layouts/landing_page.html
MakisH, line 1: I would avoid touching the formatting in unrelated files, as this might introduce unnecessary conflicts. In any case, this is only one file in this PR, and the formatting changes make sense, so that's fine.

MakisH (approved):
Thanks! Does not have any effect, since we don't use Google Analytics, but this looked anyway wrong.

PR #776, opened 2026-02-23
Fix broken markdown link in open projects page

MuhammadAashirAslam:
The [documentation] link was missing parentheses around the URL, rendering it as plain text instead of a clickable link. 

image


MakisH (approved):
I would like to be able to see that this was a planted easy bug to fix, but unfortunately it was just an honest typo. Strange that markdownlint did not catch it.

Thanks!

PR #773, opened 2026-02-23
Fix unclosed HTML tags on landing page

MuhammadAashirAslam:
Fix three HTML tag typos in the landing page (`content/index.html`):

- Lines 39 and 42: `<p>` → `</p>` (unclosed paragraph tags)
- Line 377: Removed stray `</p>` with no matching opening tag

Solves #772

PR #767, opened 2026-02-23
fix(ui): correct malformed HTML comment in topnav.html

praaatap:
### Description
This PR fixes a minor typo in the very first line of [_includes/topnav.html](cci:7://file:///d:/top/gsoc/precice.github.io/_includes/topnav.html:0:0-0:0). 

The file started with `<a!-- Navigation -->` instead of the standard `<!-- Navigation -->`. Since `<a!--` is not a valid HTML comment syntax, browsers might interpret it as an unclosed or broken `<a>` element. 

This is a single-line cleanup to ensure correct HTML syntax.

Fixes #766

PR #780, opened 2026-02-23
Fix broken HTML tags in landing page and topnav

Wdhanashri:
Found a few minor HTML issues while working on the codebase:
1. Unclosed <p> tags in content/index.html (lines 39, 42)
            <p>       
          ...
          <p>         
Browsers auto-correct this, but it's invalid HTML.

2. Broken comment tag in _includes/topnav.html (line 1)
  <a!-- Navigation -->   should be 
  <a!-- is parsed as an invalid element, not a comment.

### DIFF
--- a/_includes/topnav.html
+++ b/_includes/topnav.html
@@ -1,4 +1,4 @@
-<a!-- Navigation -->
+<!-- Navigation -->
 <nav class="navbar navbar-inverse navbar-fixed-top" aria-label="Main">
     <div class="container topnavlinks">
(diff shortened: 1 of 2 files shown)


MuhammadAashirAslam:
@Wdhanashri  i think point 2 was solved by #767 and also point 1 was solved by #773 (may be you have not pulled the recent changes)

Wdhanashri:
@MuhammadAashirAslam Thanks for pointing it out.

Issue #404, opened 2024-05-03
Wrong "updated on" date for imported pages


fsimonis:
Submodules are still shallow-cloned, leading to incorrect "Updated" status of imported pages.

This is a bug in the action itself.

MakisH:
This issue still holds: all tutorials show the same "updated" date as the last commit to the website. Example: https://precice.org/tutorials-aste-turbine.html

Related issue in the checkout action: https://github.com/actions/checkout/issues/1124

Aayushman-nvm:
This should be resolved in PR #783. @MakisH  could you please review? Happy to make any adjustments if needed.

PR #783, opened 2026-02-23
fix: fetch full history for submodules to fix incorrect last-modified dates

Aayushman-nvm:
Pages from submodules were showing incorrect last-modified dates because actions/checkout does a shallow clone of submodules even when fetch-depth: 0 is set. This caused all submodule pages to show the same wrong date.

Fix: added a step to unshallow all submodules after checkout so Jekyll can 
correctly determine the last commit date for each file.

Closes: Issue #404 

Screenshots:
Before -
image

After -
image


Aayushman-nvm:
Still working on it

Aayushman-nvm:
@MakisH For this issue i might've to modify vendor files... hook.rb to be exact
path: /vendor/bundle/ruby/3.3.0/gems/jekyll-last-modified-at-1.3.2/lib/jekyll-last-modified-at/hook.rb

is that fine?

MakisH:
Thanks for the contribution! That's indeed an annoying bug...



> @MakisH For this issue i might've to modify vendor files... hook.rb to be exact path: /vendor/bundle/ruby/3.3.0/gems/jekyll-last-modified-at-1.3.2/lib/jekyll-last-modified-at/hook.rb
> 
> is that fine?

Doesn't sound right: `/vendor/` is auto-generated and not even in this repository. Modifying it could cause all kinds of conflicts.

Aayushman-nvm:
@MakisH haha true but i think u might wanna check this screenshot out:

deployed site:
image

My local site:
image

I made a slight change... just a single line change and now the dates are displayed as 14th jan... if u could clarify if 14th jan is ri8 then i think i might've fixed the issue

Aayushman-nvm:
for a context... I wrote a script (inject_dates.sh) that goes into each submodule, gets the real last-commit date for every markdown file using git, and injects it as last_modified_at into the frontmatter before Jekyll builds.

then changed = to ||= in hook.rb so the plugin respects existing frontmatter values instead of always overwriting with its own (broken for submodules) git detection.

added two new steps in deploy.yml... one of them is not yet commited

Aayushman-nvm:
@MakisH here's what ive done so far:

inject_dates.sh, a new file im adding to the repo root. This is a shell script that before Jekyll builds, goes into each submodule, finds every markdown file, gets its real last git commit date, and writes it into the file's frontmatter as last_modified_at.

.github/workflows/deploy.yml, in this file ive added 3 new steps: unshallow submodules so full git history is available, run inject_dates.sh, and patch the plugin's hook.rb to respect frontmatter instead of overwriting it.

The reason I patched patch = to ||= in hook.rb is that the plugin runs git log from the top level git directory, which has no history for files inside submodules so it falls back to filesystem mtime (always today's date). By changing to ||=, the plugin still works as before for all regular pages, but respects the last_modified_at we inject via inject_dates.sh for submodule files where the plugin's git detection fails. Pages that don't have last_modified_at in their frontmatter are completely unaffected.

Aayushman-nvm:
In case you’re going to test and review this branch locally, you need to run the following scripts in sequence:

1. Run the date injection script:
 ```bash
bash inject_dates.sh
```
2. Patch the jekyll-last-modified-at hook:
```bash
 sed -i 's/item\.data\["last_modified_at"\] = Determinator/item.data["last_modified_at"] ||= Determinator/' \
  vendor/bundle/ruby/*/gems/jekyll-last-modified-at-*/lib/jekyll-last-modified-at/hook.rb
 ```
3. Start the Jekyll server:
```bash
bundle exec jekyll serve -l
```

We don’t need to worry about running these commands manually once the code is merged, they’re already configured in the deploy.yml files.

fsimonis:
Hi, I have some questions:

1. Why do you unshallow each submodule instead of replacing the checkout action with a standard deep git clone? At this point we just need a simple clone and disable all the features that the action does.
2. Why doesn't jekyll-last-modified-at detect the correct times now? The information is in git.
3. What is the point of patching files and jekyll-dependencies for a feature that will be removed as part of the GSoC project?

Aayushman-nvm:
Hi @fsimonis :
Answering ur questions -

1:- The workflow already used actions/checkout consistently for all three checkouts. Replacing just the website checkout with a plain git clone while keeping the action for the other two would be inconsistent. As a contributor, I wanted minimal changes to the existing working CI rather than restructuring it. The unshallow step was the least invasive addition to make submodule history available. I'm happy to replace it with a plain deep clone if you prefer.

2:- Even though the history is in git, the plugin architecturally can't reach it. Looking at `determinator.rb`, it runs `git log` with `--git-dir` set to `git.top_level_directory` the parent repo's `.git` directory (This repo). From the parent repo's perspective, a submodule is just a commit pointer, not a folder with individual file histories. So when the plugin computes `relative_path_from_git_dir` for say `imported/tutorials/quickstart/README.md`, it looks for that path in the parent repo's git history... finds nothing... and falls back to `mtime` which is always today in CI. The information is in git but inside each submodule's own `.git`, which the plugin never looks into.

3:- My thought was the migration to Hugo will probably take some time and the Jekyll site will remain in production the entire time. Hugo will be developed in a separate branch and only replace Jekyll once it's a complete and tested replica, after which Bootstrap and UI work still follows. Until that point all pages will show wrong dates and users will see that... and for a documentation site that's not great. The project will start in end of May with main coding work beginning in May / June, so users would be seeing wrong dates for several months while the Jekyll site is still live.
I also looked into whether this approach carries over to Hugo and it doesn't. Hugo with `enableGitInfo = true` handles submodule dates natively without any of these workarounds, so this fix is intentionally Jekyll-only and throwaway, meant only to cover the gap between now and when the Hugo migration goes live in production... keeping what users see on the site in mind. That said, the call is completely yours.

.github/workflows/deploy.yml
```diff
@@ -72,6 +80,19 @@ jobs:
...
+        run: |
+            set -euo pipefail
+            shopt -s nullglob
+            hook_files=(vendor/bundle/ruby/*/gems/jekyll-last-modified-at-*/lib/jekyll-last-modified-at/hook.rb)
+            if [ "${#hook_files[@]}" -ne 1 ]; then
+              echo "Expected exactly one jekyll-last-modified-at hook.rb file, found ${#hook_files[@]}:" >&2
+              printf '  %s\n' "${hook_files[@]}" >&2
+              exit 1
+            fi
+            sed -i 's/item\.data\["last_modified_at"\] = Determinator/item.data["last_modified_at"] ||= Determinator/' "${hook_files[0]}"
```
MakisH: Is this patch something you found in some documentation/issue, or did some tool suggest it? Does a newer version or other GitHub Action fix the issue?

I see that the respective project has been archived, so contributing to upstream would not work. In that case, patching is fine, but it should be explained in the comment.

Some comments above the step would help understand what is going on and why this is needed.

.github/workflows/deploy.yml
```diff
@@ -35,6 +35,14 @@ jobs:
           path: website
           fetch-depth: 0 # full history required to determine last edit
 
+      - name: Fetch full history for submodules
+        working-directory: website
+        run: git submodule foreach --recursive "git fetch --unshallow --all 2>/dev/null || git fetch --all"
+
```
MakisH, line 42: I am fine with this trick, as long as there is a comment above it explaining the intention.

MakisH (changes requested):
Thank you for the follow-up. Main point from my side would be that the respective hacks would need some comments, so that they make more sense in the future.

Given that we intend to modify this infrastructure a lot in a few months, I do not mind getting the hacks merged. The [jekyll-last-modified-at](https://github.com/gjtorikian/jekyll-last-modified-at) project is anyway officially abandoned and [the wrong times issue seems to be known](https://github.com/gjtorikian/jekyll-last-modified-at/issues/106) (if not a different one).

It should be clear, though, that most/all of this code would soon be removed again. But since the work has already been invested and I do not see any harm, I would be fine getting a temporary fix merged.

By the way, did you consider directly contributing to https://github.com/actions/checkout/issues/1124 ? That would have an impact to more projects.

MakisH:
Note that pre-commit fails: https://github.com/precice/precice.github.io/actions/runs/22350276896/job/66926818293

Aayushman-nvm:
@MakisH my bad I never got any notification of this review/comment, ill get working on this right away

.github/workflows/deploy.yml
```diff
@@ -72,6 +81,22 @@ jobs:
...
+        run: |
+            set -euo pipefail
+            shopt -s nullglob
+            hook_files=(vendor/bundle/ruby/*/gems/jekyll-last-modified-at-*/lib/jekyll-last-modified-at/hook.rb)
+            if [ "${#hook_files[@]}" -ne 1 ]; then
+              echo "Expected exactly one jekyll-last-modified-at hook.rb file, found ${#hook_files[@]}:" >&2
+              printf '  %s\n' "${hook_files[@]}" >&2
+              exit 1
+            fi
+            sed -i 's/item\.data\["last_modified_at"\] = Determinator/item.data["last_modified_at"] ||= Determinator/' "${hook_files[0]}"
```
fsimonis: @Aayushman-nvm Instead of patching this here, could you open a PR in the repo of the `jekyll-last-modified-at` gem and then use the Gemfile of this repo to point to this PR?

inject_dates.sh
fsimonis, line 1: @Aayushman-nvm as you are essentially overwriting `last_modified_at` here, this leads me to an idea.

Could you instead use a minimal jekyll plugin to apply this to imported files? This should be pretty straight forward, as all files in `imported/` need to be patched.

MuhammadAashirAslam: @fsimonis i think you should see #856 i have implemented the same solution you are recommending

Aayushman-nvm: > @Aayushman-nvm as you are essentially overwriting `last_modified_at` here, this leads me to an idea.
> 
> Could you instead use a minimal jekyll plugin to apply this to imported files? This should be pretty straight forward, as all files in `imported/` need to be patched.

Yep for sure... I'm on it

Aayushman-nvm:
@fsimonis Ive addressed the reviews:

Replaced the previous solution (shell script + sed patch on the gem) with a native Jekyll generator plugin `_plugins/imported_dates.rb`

- Scoped strictly to `imported/`
- Batches git calls per submodule root instead of spawning one process per file, significantly faster on large submodules
- Uses Open3.capture3 with argv array, no shell injection risk from filenames
- Computes file paths relative to each submodule's git root, correctly handles files in subdirectories, not just submodule root-level files
- Uses stable Jekyll API (site.pages + site.documents) throughout
- Respects explicitly set last_modified_at frontmatter, never overwrites it
- Skips files with no git history with a named warning instead of silently falling back to unreliable filesystem mtime
- Removes the dependency on jekyll-last-modified-at gem entirely, no gem patching, no sed, no shell scripts

Aayushman-nvm:
As for testing:

Verified submodule has full git history:

```bash 
git -C imported/tutorials log -1 --format="%ci" -- volume-coupled-flow/README.md
# 2026-02-03 12:37:38 +0100
```

Confirmed plugin produces matching date with no warnings:

```bash
bundle exec jekyll build 2>&1 | grep -i "ImportedLastModifiedAt"
# (no output, clean build)
```
Confirmed frontmatter override is respected added `last_modified_at: 2026-03-17 00:00:00` to `volume-coupled-flow/README.md`, rebuilt, audit showed `2026-03-17` despite git having `2026-02-03`. Removed after confirming.

.github/workflows/deploy.yml
MakisH: @fsimonis the `jekyll-last-modified-at` plugin seems to have been abandoned, see https://github.com/precice/precice.github.io/pull/783#pullrequestreview-3942439390

MakisH:
@Aayushman-nvm as discussed in #856, I appreciate the effort, but this solution still looks rather confusing to me.

Since you asked about Copilot in https://github.com/precice/precice.github.io/pull/856#issuecomment-4076262613, I requested another Copilot review. I don't know what more it can do, but I have looked at both PRs and tried to understand them. Both seem to work, but I found the solution in #856 easier to understand.

I would say let's not invest more effort on this one. A few learning outcomes have been achieved anyway, despite merged or not.

Aayushman-nvm:
@MakisH i understand, its all good 😃

PR #784, opened 2026-02-23
fix: restore focus outline on navbar links for accessibility

Sahaj0209:
Problem : Keyboard users navigating the site with the Tab key had no visible focus indicator on navbar links. Users cannot see which element is currently focused .
Root Cause : In "css/customstyles.css", the ".nav" had "outline: none" which suppressed the browser's default focus ring on all navigation links .
Fix : Commented "outline: none" from the ".nav"  in "css/customstyles.css"
Testing : Open the site and press Tab to navigate through navbar links . A visible focus ring now appears on each focused element . Links elsewhere on the site are unaffected .
fix-nav


MakisH:
Thanks but... how do I even focus the navbar in Firefox/Chromium? I can only do that with the mouse (and I see the difference).

MuhammadAashirAslam:
Hi @MakisH, I think this was only verified on Chrome-based browsers. I’ve opened a PR that fixes the issue globally across browsers #797

PR #786, opened 2026-02-23
fix: moment.js loaded globally but only used in search

### LINKED ISSUE #785: moment.js loaded globally but only used in search
### Page URL

https://precice.org

### What is wrong?

`moment.js` (~17KB) is loaded unconditionally on every page of the website via [_includes/head.html](cci:7://file:///d:/precice/precice.github.io/_includes/head.html:0:0-0:0) (line 101), but it is only ever used in one place: [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0) (line 20), to format a date string in search results.

**Current behaviour:** Every visitor downloads `moment.js` regardless of whether they interact with the search widget at all.

**Expected behaviour:** `moment.js` should only be loaded as a dependency of the search widget ([algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0)), not as a global resource in `<head>`.

**Proposed fix:**
- Remove the `<script src="...moment.min.js">` tag from [_includes/head.html](cci:7://file:///d:/precice/precice.github.io/_includes/head.html:0:0-0:0)
- Add it at the top of [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0), directly before the inline script that consumes it

This is a one-file-change with zero risk of breakage since `moment` has exactly one call site.

Parth-Garg2629:
## Summary

`moment.js` (~17KB) was loaded unconditionally on **every page** via [_includes/head.html](cci:7://file:///d:/precice/precice.github.io/_includes/head.html:0:0-0:0), but it is only used in [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0) (line 20) to format dates in search results.

## Changes

- **Removed** `<script src="...moment.min.js">` from [_includes/head.html](cci:7://file:///d:/precice/precice.github.io/_includes/head.html:0:0-0:0)
- **Added** it at the top of [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0), directly before the inline script that calls `moment.unix(...).format(...)`

## Why

This removes a hidden global dependency and co-locates the library with its only consumer. It also prepares the codebase for conditionally loading the search widget (and its deps) in the future.

## Testing

- Verified search results still display correctly with formatted dates
- `moment.js` is still loaded as part of the search widget include

Closes #785


fsimonis:
See issue

PR #794, opened 2026-02-23
Feat: Sequence Diagram SVGs into Mermaid renders

### LINKED ISSUE #792: Feature: Use mermaidjs for sequence diagrams?
Now that we have mermaidjs support, maybe some of the `sequence diagrams` can be visualised better with these as mermaid has native support for sequence diagrams.

# Advantages of mermaidjs over raw images/SVGs (for sequence diagrams):
1. Mermaid diagrams are much easier to maintain since they are written in plain text, whereas SVGs require manual editing or regenerating from external tools.

2. Contributors can update Mermaid diagrams directly in the Markdown files, which lowers the barrier to contribution compared to editing and exporting SVG files.

3. Changes to Mermaid diagrams are clearly visible in Git diffs, while SVG diffs are mostly coordinate and XML changes that are hard to review.

4. Mermaid removes the need to manage separate static image files, keeping the diagrams co-located with the documentation and simplifying the repository structure.

5. Hugo and GitHub both support Mermaid well, so diagrams render automatically without requiring additional export steps or asset handling.

6. Require separate SVGs for Dark mode (noticed this when I pasted the SVG here).

# Created a quick POC to for the maintainers to consider.

> Reference: https://precice.org/couple-your-code-coupling-flow.html
## SVG

Image

## mermaidjs render

```mermaid
sequenceDiagram
    participant FluidSolver
    participant SolidSolver

    Note over FluidSolver,SolidSolver: Initialization
    FluidSolver->>FluidSolver: precice.initialize()
    SolidSolver->>SolidSolver: precice.initialize()

    SolidSolver->>FluidSolver: send mesh (SolidMesh)
    FluidSolver->>FluidSolver: receive mesh

    loop Coupling timestep

        Note over FluidSolver: Solve fluid timestep
        FluidSolver->>FluidSolver: solveTimeStep(...)
        FluidSolver->>FluidSolver: precice.advance()

        Note over FluidSolver: Map write Forces\nFluidMesh → SolidMesh
        FluidSolver->>SolidSolver: send Forces

        Note over SolidSolver: Waiting for Forces
        SolidSolver->>SolidSolver: receive Forces
        SolidSolver->>SolidSolver: solveTimeStep(...)
        SolidSolver->>SolidSolver: precice.advance()

        SolidSolver->>FluidSolver: send Displacements

        Note over FluidSolver: receive Displacements
        FluidSolver->>FluidSolver: Map read Displacements\nSolidMesh → FluidMesh

    end

    Note over FluidSolver,SolidSolver: Finalization
    FluidSolver->>FluidSolver: precice.finalize()
    SolidSolver->>SolidSolver: precice.finalize()
```

hanzalahwaheed:
Fixes https://github.com/precice/precice.github.io/issues/792

I've tried to convert the diagrams into mermaid renders. Here are the tow outputs that I have. Need help choosing the better one.

# serial coupling diagram

> reference: https://precice.org/couple-your-code-coupling-flow.html

## original screenshot

image

## mermaid render A

This is a more direct one-to-one mapping of the diagram.

image

## mermaid render B

This is a more semantic render.

image

@MakisH which one do you think is a better render? I'm inclined towards A but the loop box of B is also nice.


MakisH:
Thank you @hanzalahwaheed, this is useful. We will eventually review it and see how it might work together with other documentation as well.

MakisH:
Overall, I think this is nice for new graphs, but I am not yet convinced of the value that the Mermaid version of this figure would add. I would suggest not investing more work on this at the moment, let's just use this PR to evaluate the option.

Some comments, in arbitrary order:

- The SGV (made with TikZ) also has some colors, which point to the two participants. Can we color boxes or text differently?
- In "Approach A", the two participants seem like they are taking turns in calling `initialize()`. This is not correct: up to that point, they are decoupled. Similarly, in "Approach B", the calls to `initialize()` and `finalize()` seem to be one after another.
- In "Approach B", I like the loop box.
- In the SVG, the indentation relates to the different software layers, and it is nice to have.
- An advantage of the SVG version is that it can easily be exported and used in a PDF document. How would we integrate Mermaid graphs there?

I tend towards not migrating this figure to Mermaid, but simpler/new figures would be fine to have directly in Mermaid. @fsimonis what do you think?

MakisH:
Closing due to inactivity, and since this is far from a state to get merged.

In any case, I think this is a nice proof-of-concept, and thank you for contributing! The main technical work in #726 was already very useful.

PR #795, opened 2026-02-24
Add Escape-key support to Algolia search dropdown

tanishqio:
This PR adds Escape-key support to the Algolia search dropdown.

Currently, the dropdown can only be dismissed by clicking outside the search area. Pressing Escape has no effect.

This change adds a small `keydown` listener in `_includes/algolia.html` to:
- Hide the search results dropdown when Escape is pressed
- Remove focus from the search input (if focused)

Scope:
- Single file change
- No layout or dependency modifications
- No visual changes
- Existing mouse-based behavior remains unchanged

This aligns the search behavior with common UX expectations and improves keyboard accessibility.


MakisH:
Thanks for the contribution! I tested it locally and it does what it should.

Consider Copilot's feedback for what it is.

MakisH (approved):

PR #797, opened 2026-02-24
Style: restore focus outline on navbar links globally for accessibility

MuhammadAashirAslam:
Fix missing focus outline in Firefox and Chromium by suppressing the global

```
*:focus {
  outline: none;
}
```

rule in customstyles-precice.css.

This restores visible keyboard focus when navigating with Tab.

**Result**

Focus box now appears correctly in Firefox and Chromium browsers.

image


MakisH (approved):
That's useful, thanks!

PR #800, opened 2026-02-24
docs: use markdown formatting instead of raw html tags

praaatap:
docs: use markdown formatting instead of raw html tags

## Description
This PR replaces instances of raw HTML `<i>` tags with standard Markdown formatting (`*` for italics) inside `.md` content files.

Mixing raw HTML inside Markdown documents makes the raw source code harder to read and breaks common Markdown linting rules. 

## Changes Made
- Converted `<i>` to `*` for italics in `content/community/workshops/precice-workshop-2023.md`.
- Verified that no other content files are relying on raw HTML `<b>` or `<i>` tags where standard Markdown would suffice.

## Testing
- [x] Confirmed the pages render formatting identically in standard Markdown viewers.


MakisH:
Thank you, but this is wrong: this is inside an HTML tag, and the suggested form does not render correctly.

PR #790, opened 2026-02-23
Add keyboard shortcut to focus algolia search input box.

hanzalahwaheed:
Fixes https://github.com/precice/precice.github.io/issues/748

# Screenshot

image

PR #801, opened 2026-02-24
refactor: consolidate Algolia search into external js/algolia-search.js

### LINKED ISSUE #796: Refactor Algolia search to use external js/algolia-search.js (remove duplicate inline implementation)
## Problem

The Algolia search logic currently exists in two places:

- `_includes/algolia.html` — active inline implementation (used in production)
- `js/algolia-search.js` — unused external implementation

This results in duplicate code and creates confusion about which version is canonical.

## Why this matters

- Violates single source of truth (logic exists in two files)
- Makes future changes error-prone
- Inline JS cannot be browser-cached
- Increases complexity inside layout templates
- The external JS file currently appears unused

## Proposed solution

- Keep `_includes/algolia.html` only for passing Liquid configuration variables
- Load `js/algolia-search.js` as the single implementation file
- Remove the duplicate inline logic

This would:

- Consolidate the implementation
- Enable browser caching
- Simplify templates
- Reduce maintenance risk

No visual or functional changes are intended.

---

Since `_includes/algolia.html` was recently modified (Escape-key improvement), I wanted to open this issue first before proceeding with any structural refactor.

Parth-Garg2629:
## Problem

The Algolia search logic existed in two near-identical places:
- [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0) — active inline implementation (~135 lines of JS in a template)
- [js/algolia-search.js](cci:7://file:///d:/precice/precice.github.io/js/algolia-search.js:0:0-0:0) — unused duplicate

## Solution

- [_includes/algolia.html](cci:7://file:///d:/precice/precice.github.io/_includes/algolia.html:0:0-0:0) is now a thin config bridge only — it exposes Jekyll Liquid variables as `window.ALGOLIA_CONFIG` and loads the external JS file
- [js/algolia-search.js](cci:7://file:///d:/precice/precice.github.io/js/algolia-search.js:0:0-0:0) is now the single implementation file — reads config from `window.ALGOLIA_CONFIG`, pure vanilla JS with no Liquid tags

## Benefits

- ✅ Single source of truth for search logic
- ✅ External JS is browser-cacheable (inline scripts are not)
- ✅ Templates stay clean — no logic mixed into layout includes
- ✅ No YAML front matter tricks needed — config passes cleanly via a global

## No functional changes

Search behaviour, styling, and UX are identical to before.
The 167 lines removed = the duplicate. The code that was identical in both files now lives in one place (
algolia-search.js). Nothing was deleted that wasn't already covered by the external file.

Closes #796


MakisH:
Thank you for the contribution. There have been changes in `algolia.html` via #790 and #795, leading to conflicts here that would need to be resolved. In the meantime there has also been an alternative solution in #823.

With a first look at this PR, I think it is still strange that it keeps both files. In any case, this will be cleaned up in the context of the framework migration.

Since the effort/iterations required to bring this PR in a mergeable state would be higher than the benefit it would bring, I am closing this now.

The solution is much simpler: just remove the unused file. I did that now in https://github.com/precice/precice.github.io/pull/851

PR #804, opened 2026-02-24
refactor: move Algolia search box styles from JS to CSS

Parth-Garg2629:
The Algolia search box was styled via JS (`elem.style.*` calls) in `algolia-search.js`, with a comment acknowledging this was a workaround. The equivalent CSS was already written — just commented out in the same file.

This PR moves those styles into the existing Search section of `css/customstyles-precice.css` and removes the JS style injection and the dead commented CSS block.

No visual changes — styles are identical.


MakisH:
Since there are a few PRs touching the Algolia script, conflicts are accumulating. Since the box will anyway be replaced when we change the framework, I think that reviewing and iterating on related PRs (especially anything related to moving source code between files / non-functional) would need more effort reviewing than it would bring long-term benefit.

Thank you for the contribution anyway.

Issue #586, opened 2025-09-04
Sidebar elements from imported repositories


MakisH:
We have some imported repositories (e.g., `tutorials`, `preeco-orga`, `openfoam-adapter`), which define their own documentation pages. We sometimes add new pages in their documentation, but we then need yet another PR here to add them to the respective sidebar, which we often forget to do (or is quite tedious).

It would be great to collect sidebar elements also from these imported repositories.

For example: in the tutorials sidebar, we could host the sidebar file in the tutorials repository and import it from there.

Another example: we have the OpenFOAM adapter in the docs sidebar. We could host a YAML file in each adapter repository, and stitch all of them together, in alphabetical order, in the tutorials sidebar file.

A solution could be with jinja2 templates and a Python script that stitches together different sidebar yaml files.

fsimonis:
As part of the hugo migration, the complete sidebar(s) should be autogenerated.

Aayushman-nvm:
Hi @MakisH, I’ve opened a PR (#810 ) that partially addresses this issue. Would appreciate your feedback whenever you get a chance.

PR #810, opened 2026-02-25
Automate tutorial sidebar generation

Aayushman-nvm:
## Automate tutorial sidebar generation

Replaces the manually maintained `tutorial_sidebar.yml` with a Liquid-based auto-discovery system (related to #586 - partial fix).

### Problem

Adding a tutorial required manually updating `tutorial_sidebar.yml`, creating duplication and cross-repo dependency between the main site and the tutorials submodule.

### Solution

- **_includes/sidebar.html**  
  Dispatches to a dedicated tutorial sidebar include when `tutorial_sidebar` is requested. Other sidebars remain unaffected.

- **_includes/tutorial_sidebar.html**  
  Auto-builds the sidebar using `site.pages` filtered to `imported/tutorials/`.  
  Renders:
  - **Introduction** (hardcoded — lives in main repo)
  - **Basic cases** (configured via data file, link-only behavior preserved)
  - **All tutorials** (auto-discovered, alphabetically interleaved flat pages and groups)

- **_data/tutorial_sidebar_config.yml**  
  Centralizes editorial decisions: featured tutorials, subgroup definitions, display order, short title overrides.  
  Only this file requires manual updates.

- **_data/sidebars/tutorial_sidebar.yml**  
  Preserved for reference but no longer used.

### Result

- New tutorial added → appears automatically in **All tutorials**
- Featuring/grouping changes → edit `_data/tutorial_sidebar_config.yml` only
- No additional front matter required in tutorial READMEs
- All other sidebars remain unchanged

### Hugo Migration

The configuration file structure is directly portable.  
Template logic maps cleanly to Hugo (`range where`, `GroupByParam`).  
No changes required to tutorial front matter.

### Testing

- Renamed and deleted `tutorial_sidebar.yml` → sidebar still renders
- Verified layout matches hosted site

### Note

- The sidebar for other repos can be automated similarly while being completely Hugo migration safe and will close the issue completely. This refers to issue #586.
- Short names present in tutorial_sidebar.yml couldn't be replicated, its only possible if all the README.md of subfolders in imported/tutorials has those short names in its front matter.

### Comparison

**Hosted site**

Hosted site sidebar

**Local site**

Local site sidebar

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,74 +1,119 @@
-{% assign sidebar = site.data.sidebars[page.sidebar].entries %}
+{% if page.sidebar == 'tutorial_sidebar' %}
+  {% include tutorial_sidebar.html %}
+{% else %}
+  {% assign sidebar = site.data.sidebars[page.sidebar].entries %}
 
-<nav aria-label="{{sidebar[0].product}}">
-<ul id="mysidebar" class="nav">
-  <li class="sidebarTitle">{{sidebar[0].product}} {{sidebar[0].version}}</li>
-  {% for entry in sidebar %}
-  {% for folder in entry.folders %}
-  {% if folder.output contains "web" %}
-  {% if folder.url %}
-  <!-- Parent menu item has url and no folderitems -->
-  {% if page.url == folder.url %}
-  <li class="active">
-    <a title="{{folder.title}}" href="{{folder.url | remove: "/"}}">{{folder.title}}</a>
-  </li>
-  {% else %}
-  <li>
-    <a title="{{folder.title}}" href="{{folder.url | remove: "/"}}">{{folder.title}}</a>
-  </li>
-  {% endif %}
-  {% else %}
-  <!-- Parent menu item has folderitems -->
-  <li>
-    <a title="{{folder.title}}" href="#">{{folder.title}}</a>
-    <ul>
-      {% for folderitem in folder.folderitems %}
-      {% if folderitem.output contains "web" %}
-      {% if folderitem.external_url %}
-      <li><a title="{{folderitem.title}}" href="{{folderitem.external_url}}" target="_blank" rel="noopener">{{folderitem.title}}</a></li>
-      {% elsif page.url == folderitem.url %}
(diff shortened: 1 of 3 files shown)


_includes/sidebar.html
```diff
@@ -1,74 +1,119 @@
...
-      {% elsif page.url == folderitem.url %}
-      <li {% if folderitem.link-only == nil or folderitem.link-only == false %} class="active" {% endif %} ><a title="{{folderitem.title}}" href="{{folderitem.url | remove: "/"}}">{{folderitem.title}}</a></li>
-      {% elsif folderitem.type == "empty" %}
-      <li><a title="{{folderitem.title}}" href="{{folderitem.url | remove: "/"}}">{{folderitem.title}}</a></li>
+  <nav aria-label="{{sidebar[0].product}}">
+    <ul id="mysidebar" class="nav">
+      <li class="sidebarTitle">
+        {{ sidebar[0].product }}
+        {{ sidebar[0].version }}
+      </li>
```
MakisH, line 11: This part seems to be unfinished (generic names).

MakisH (commented):
Thank you for the contribution! Seems to be working overall, but it is also clearly complicated. When we change the framework, it might me much simpler (https://github.com/precice/precice.github.io/issues/472).

I am also wondering why the `preCICE system tests` appears in the tutorials sidebar. This is fetched from the tutorials repository, but it is not a tutorial:

Image

The ASTE wind turbine tutorial is currently a bit long in the sidebar, while previously we had a custom name. How would you approach a custom name here?

This seems already like a lot of work. I would avoid investing more effort here unless it is clear that we cannot get a simpler solution with Hugo.

MakisH:
Actually, there is a bigger issue: I still need to manually add new tutorials into `_config.yml`, which means that this is not a solution to the issue of having to contribute to two repositories to add a new tutorial. That is a more fundamental issue.

Until we solve the automatic import, I think the automatic sidebar generation is not really helpful and might make the process more confusing.

Nice concept demonstration, but I would say let's not invest more effort here at the moment.

PR #811, opened 2026-02-25
Cleanup/remove duplicate algolia js

Sahaj0209:
Problem: Algolia search logic existed in two places: "_includes/algolia.html" (active) and "js/algolia-search.js" (unused). This violated single source of truth and added confusion about which version was canonical.
Fix: Deleted "js/algolia-search.js" as it is identical to and superseded by "_includes/algolia.html", which is the only active implementation included via "_includes/topnav.html".
Verified: "js/algolia-search.js" is not referenced anywhere in the code base:


MakisH:
Thank you. The same is already addressed by https://github.com/precice/precice.github.io/pull/801/changes

This PR actually does not seem to do what it claims.

PR #812, opened 2026-02-25
Fix/feed xml undefined variable

Sahaj0209:
Problem: In "feed.xml" line 27, the loop iterates with variable tag but the template referenced the undefined variable cat, meaning category tags were always rendered empty.
Fix: Replaced {{ cat | xml_escape }} with {{ tag | xml_escape }} to match the loop variable.


MakisH:
Thank you. There is already another open PR on this: https://github.com/precice/precice.github.io/pull/803

Note that this PR also includes some unrelated changes.

PR #798, opened 2026-02-24
Fix(html): Fix structural typos

MuhammadAashirAslam:
# HTML Typo Fixes Summary

- **`_includes/custom/usermapcomplex.html`**: Added a missing closing `</p>` tag on line 74. The paragraph was previously unclosed and enveloping a `<ul>`.
- **`_layouts/default_print.html`**: Removed a redundant, duplicated `<html>` tag at the top of the file, preserving the correct `<html lang="en">` tag intact.
- **`content/index.html`**: Removed an extra, stray closing `</div>` tag that didn't have a matching opening tag.

PR #815, opened 2026-02-26
Fix: remove extra closing div in community section

praaatap:
### I noticed an extra </div> tag at the end of the "Join the community" section that didn't need to be there. I've removed it to keep the HTML clean and make sure the page structure stays solid.

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -382,7 +382,7 @@ layout: landing_page
           {% include newsletter.html %}
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


MuhammadAashirAslam:
Hey @praaatap , i have already solved this issue (#798 ) . I think you haven't pulled the new changes

pratapsisodiya:
got it @MuhammadAashirAslam

PR #793, opened 2026-02-23
Fix XSS via unsanitized innerHTML in JS files

vagxrth:
## Summary
  - Replace `innerHTML` with safe DOM APIs (`createElement`, `textContent`, `setAttribute`) in three JS files to prevent stored XSS from Discourse/GitHub API data
  - Add inline URL validation (`isSafeUrl`) to reject non-HTTP(S) protocols (e.g., `javascript:`)
  - Fix implicit global variables in `github-queries.js`

  Closes #789

  ## Files changed
  - `js/news-collect.js` — landing page news cards
  - `js/forum-fetch.js` — FAQ page topic cards
  - `js/github-queries.js` — GitHub release button and star count

PR #816, opened 2026-02-26
fix: resolve implicit global variables in github-queries.js

praaatap:
## Description
This PR fixes a few places where variables were being assigned without an explicit declaration (`let`, `const`, or `var`) in [js/github-queries.js]

- Pollution of the global namespace.
- Conflict with other scripts.
- Difficult-to-debug side effects.

I have refactored these places to use `const` because these variables are not being reassigned in their scopes.

## Changes
- Declared `date` inside the function.
- Declared [tag]
- Declared `count` inside the stargazers-fetching closure.

## Verification
- Verified by reviewing the script logic.
- Verified locally to ensure GitHub stats are still being fetched and displayed correctly.


js/github-queries.js
```diff
@@ -27,9 +27,9 @@ var github_api_endpoint = 'https://api.github.com/repos/precice/precice/releases
     fetch(github_api_endpoint).then(function(response) {
       if (response.ok) {
         response.json().then(function(data) {
-          tag = data[0].name;
-          published_at = data[0].published_at;
-          url = data[0].html_url
+          const tag = data[0].name;
+          const published_at = data[0].published_at;
+          const url = data[0].html_url
```
hanzalahwaheed: semicolon missing here. I noticed rest of the file uses semicolons at the end. lets follow a common syntax.

pratapsisodiya: ya thanks for help @hanzalahwaheed

MakisH:
This makes sense overall.

@praaatap could you please update and resolve the conflicts? There is also an unresolved conversation.

MakisH:
@praaatap the conflict resolution reverted the changes of #793.

PR #817, opened 2026-02-26
feat(issue-templates): add feature request template and extend bug re…

praaatap:
# Simple Update for preCICE Website

We want to make it easy for people to help us. So we added two things:

1. **New Feature Form**: Use this if you have a cool new idea for the site.
2. **Fixed Bug Form**: We added a "Feedback" box at the bottom for extra notes.

### Why this is good?
It helps us understand your ideas faster. Now we can fix bugs and add new things more easily!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -17,3 +17,8 @@ body:
       placeholder: "The link to X returns a 404. It should point to Y."
     validations:
       required: true
+  - type: textarea
+    id: feedback
+    attributes:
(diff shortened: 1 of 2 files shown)


MakisH:
Hi! There have recently been a few new issue templates (from you, iirc), which are generally useful. I would like to avoid adding more templates, since this introduces friction when opening a new issue: there are already five things to read and select from:

Screenshot from 2026-03-14 12-00-43

Moreover, I don't really think that a "new feature request" makes much sense here, or that the content of the template adds much value. There is always the option of a blank issue.

Thank you for the contribution anyway!

PR #820, opened 2026-02-27
Add a community spotlight to the community page

AmazingDude:
Hi preCICE team

I've been looking into the Website Modernization project and wanted to tackle the "Entry Test" mentioned in the ideas list. I saw the recent discussion on PR #736 and totally get why adding global features to the current Jekyll/Bootstrap 3 setup isn't the priority right now since the Hugo migration is the real goal

That said, I wanted to see if I could build something that fulfills the test adding a new element to the landing page without creating a maintenance headache for you guys

I’ve added a community_spotlight component that’s completely self contained. Here is how it looks:

### Preview:
Screenshot 2026-02-27 155517

### What I focused on:
* **Keeping it isolated:** I used scoped CSS classes so this won't mess with any of the existing legacy Bootstrap styles
* **Data-driven:** Instead of hardcoding anything, I used Liquid to pull directly from _data/developer.yaml. I had to do some filtering/deduplication there to make sure the grid stays clean
* **Modern layout:** I used CSS Grid/Flexbox for the avatar display. It was a good exercise to see how the current data structure handles responsive layouts before the eventual jump to Hugo/Bootstrap 5

I'm mainly submitting this to show I've got the local environment running and can handle the frontend logic. If you'd rather keep the current landing page exactly as is until the migration starts no worries at all

Looking forward to hearing what you think of the direction


MakisH:
Thank you, @AmazingDude, for the contribution!

On the "should we do this", I have mixed feelings: We already have a [join the community](https://precice.org/#join-the-community) section, which used to have a nice photo (now moved to the [Community overview](https://precice.org/community.html)). I think that putting our faces there mainly makes it about "who maintains", not "who the community is".

Maybe a better question arising from this PR is if the community overview is easy to find from the landing page. Similarly to how the "user stories" appear as a preview in the Community Overview, we could have these pictures as a preview in the respective section: https://precice.org/community.html#contributions-to-precice

On the technical side, I would still keep the CSS in a separate file and prefix the classes with something different than `gsoc`.

I would only be able to do a better review later.

AmazingDude:
Thanks for the feedback, @MakisH . That makes total sense, I see how putting faces on the landing page might feel a bit too 'maintainer-focused'

I like the idea of moving this to the Community Overview as a preview for the contributions section. It feels much more at home there

I'll get to work on:

1. Moving the CSS to a dedicated file in assets/css/
2. Renaming the classes to something more permanent (dropping the 'gsoc' prefix)
3. Relocating the include to the Community page to act as that teaser you mentioned
4. I'll push these changes shortly. Thanks for pointing me in the right direction :)

AmazingDude:
Hey @MakisH I’ve pushed the updates we talked about

Here’s what changed in the latest commits:

- Relocated: The component now lives in community.md, right below the ## Contributions to preCICE section, so it works as a visual preview. I also removed the extra HTML headers and buttons to make sure it blends in cleanly with the existing markdown

- CSS cleanup: Moved all styling into assets/css/community-preview.css

- Class renaming: Renamed the classes from gsoc-* to contributor-preview-* to keep the naming more consistent

Here’s a quick look at how it fits into the Community page now:
image

MuhammadAashirAslam:
@AmazingDude, cool, but I have a suggestion regarding how contributors are displayed. Right now, showing the contributors immediately below “Contributions to preCICE” feels a bit out of context, since it’s not clear who these people are and the heading doesn’t quite match. I think it would look better if the text currently below this section were moved above, followed by a new heading like Maintainers / Collaborators, and then display the contributor pictures under that. Additionally, since the pretext is “Contributions to preCICE”, it might make sense to link to all contributors via [GitHub’s contributors page](https://github.com/precice/precice/graphs/contributors)
, so users have a complete view you can add main contributors and then add a see more button (though this is only a suggestion leaving maintainers there would also be correct)

AmazingDude:
Thanks for the suggestions, @MuhammadAashirAslam I actually just restructured this component specifically based on @MakisH's feedback above

He mentioned wanting to avoid the top down 'who maintains' vibe, which is why I dropped the separate headings and buttons so it acts purely as a visual preview that integrates seamlessly into the existing Contributions to preCICE markdown section

Appreciate you taking a look though!

_data/developer.yaml
```diff
@@ -232,7 +232,6 @@ contributors:
   - fullname: Gilberto Lem Carrillo
     name: Carrillo
     institution: Technical University of Munich
-    github: gilbertolem
```
MakisH, line 235: I removed this because it does not exist anymore and the avatar fetching was failing.

MakisH (approved):
Thanks for the updates! I directly applied a few changes, and I think this now looks nice!

PR #821, opened 2026-02-27
fix: correct invalid Liquid syntax in image.html include

praaatap:
## Description
This Pull Request resolves a common syntax error in the [_includes/image.html] file, where Liquid output tags were mistakenly placed inside logic tags. It also introduces minor security hardening to external links.

## Changes Made
- **Corrected Liquid Syntax:** Stripped the nested `{{ }}` tags inside `{% if %}` tags. In Liquid syntax, logic tags are evaluated directly from the context, and placing output tags inside them is actually incorrect syntax that may cause build warnings or unexpected rendering issues.
- **Security Hardening:** Added `rel="noopener noreferrer"` to external links with `target="_blank"`. This is a current best practice for web security to prevent tab-nabbing and improve user privacy when leaving the documentation site.
## How to test
1. Build the Jekyll site locally: `bundle exec jekyll serve`
2. Open any page with the [image.html]include (such as the 404 page).
3. Verify that the image URLs are still properly rendered.


hanzalahwaheed:
@MakisH LGTM!

hanzalahwaheed:
@praaatap solid PR. please rebase and update the branch for merging.

MakisH (approved):
Thanks!

PR #822, opened 2026-02-27
fix: update Font Awesome icon classes to FA6 format

praaatap:
## Description
This Pull Request will update the Font Awesome icon prefixes throughout the site's alert data and custom Liquid blocks to reflect the new FA6 style. Additionally, this PR will correct a specific typo that was causing the Disclaimer icon to not display properly on the site. 

## Changes Made
- **Fixed Icon Prefix:** Updated the deprecated Font Awesome 4 `fa` prefix to the new Font Awesome 6 `fas` (solid) prefix in [_data/alerts.yml]and [_plugins/customblocks.rb]
- **Fixed Redundant Prefix Typo:** Fixed the `fa-fa-hands-wash` typo in [_data/alerts.yml](line 5). The double prefix was preventing the icon from rendering properly on the live site.
- **Updated Deprecated Aliases:** Updated `fa-warning` to `fa-exclamation-triangle` to reflect the new Font Awesome naming conventions. ## Why this is important
Using the new `fas` prefixes in the latest Font Awesome 6 library will provide better performance and consistency across modern browsers. More importantly, this PR will fix the broken Disclaimer icon due to a high-impact typo.

### DIFF
--- a/_data/alerts.yml
+++ b/_data/alerts.yml
@@ -1,9 +1,9 @@
-tip: '<div class="alert alert-success" role="alert"><i class="fa fa-check-square"></i> <b>Tip: </b>'
-note: '<div class="alert alert-info" role="alert"><i class="fa fa-info-circle"></i> <b>Note: </b>'
-important: '<div class="alert alert-warning" role="alert"><i class="fa fa-warning"></i> <b>Important: </b>'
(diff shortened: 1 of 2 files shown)


MakisH:
Thanks for the contribution, but this PR is not clearly motivated.

Both the `fa` and `fas` are valid. The `fas` is with filling, the `fa` without. It is not related to the version, and the alerts currently work.

There used to be a typo in the disclaimer alert, but this is not the case anymore.

PR #824, opened 2026-02-28
Automate Ubuntu dependency baseline generation

### LINKED ISSUE #617: Automate dependency page
Repology provides an unstable API and database dumps for its data:
 
https://repology.org/api/v1
https://dumps.repology.org/README.txt

We could use this to fetch the versions of the dependencies for the Ubuntu baseline and keep the dependency section updated.

Aayushman-nvm:
# Automate Ubuntu dependency baseline generation

addresses #617

## Approach

Repology's API is rate-limited and the dumps are too large (noted in the issue), so this uses Docker + `apt-cache policy` to query package versions directly from Ubuntu's apt repositories instead.

## What's included

- **`scripts/dependencies.yml`** — package list (required/optional), derived from the Ubuntu system guide
- **`scripts/generate_ubuntu_baseline.py`** — queries each package version via Docker and writes `_data/ubuntu-baseline.yml`. Run manually at release time with `python3 scripts/generate_ubuntu_baseline.py --lts 22.04 24.04`. A `--mock` flag is available to test without Docker.
- **`_data/ubuntu-baseline.yml`** — generated output, committed so Jekyll can reference it at build time
- **`.github/workflows/generate-baseline.yml`** — `workflow_dispatch`-only workflow for auto regenerating and adding ubuntu-baseline.yml list locally

## How I tested

- Ran `debug.yml` (a temporary workflow, not included in this PR) on the `issue617` branch to verify all packages in `dependencies.yml` are actually available and resolvable on both Ubuntu 22.04 and 24.04 via apt.
- Ran the generation script locally with Docker to produce the committed `ubuntu-baseline.yml` and confirmed the versions match what `apt-cache policy` reports inside the containers.

## Notes

- The package list in `dependencies.yml` can be updated independently of the script — easy to add or remove packages without touching any logic.
- Page integration `content/docs/installation/building-from-source/installation-source-dependencies` is intentionally left untouched for now. Happy to wire the data file into the dependencies page however you'd prefer.

Screenshots:
Core:
image

Autogenerated:
image


scripts/generate_ubuntu_baseline.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
+
+import subprocess
+import yaml
+import argparse
+from pathlib import Path
+from datetime import datetime, UTC
```
MakisH, line 7: This Python script would still need a `requirements.txt` defining `PyYAML` as a dependency (probably `PyYAML==6`). And then change the CI to replace `pip install PyYAML` with `pip install -r requirements.txt`.

scripts/generate_ubuntu_baseline.py
```diff
@@ -0,0 +1,124 @@
...
+        raw = raw.split(":", 1)[1]
+
+    # Remove Debian revision (e.g. "-1ubuntu1")
+    raw = raw.split("-", 1)[0]
+
+    return raw
+
+
+def get_version_mock(package):
+    return "0.0.0-mock"
```
MakisH, line 65: What is the use case for this one? It just assigns `0.0.0-mock` to every dependency.

scripts/generate_ubuntu_baseline.py
```diff
@@ -0,0 +1,124 @@
...
+
+def get_version_docker(ubuntu_version, package):
+    cmd = [
+        "docker",
+        "run",
+        "--rm",
+        f"ubuntu:{ubuntu_version}",
+        "bash",
+        "-c",
+        f"apt-get update -qq && apt-cache policy {package} | grep Candidate | awk '{{print $2}}'",
```
MakisH, line 47: While this will give the package name, which includes the respective upstream version, it also includes other elements, such as the Debian package revision. For example, `1.83.0.1ubuntu2` or `2.9.14+dfsg`.

I am not sure if these would be useful for the user on the website, or rather confusing. In principle, the first three digits should work (all these packages use semantic versioning).

.github/workflows/generate-baseline.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Generate baseline
+        run: |
+          python3 scripts/generate_ubuntu_baseline.py \
+            --lts 22.04 24.04
+
+      - name: Commit if changed
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
```
MakisH, line 25: In other workflows (`update-submodules.yml` and `update-discourse-data.yml`), we use:


```suggestion
          git config --local user.name "precice-bot"
          git config --local user.email "info@precice.org"
```

.github/workflows/generate-baseline.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Python deps
+        run: pip install PyYAML
+
+      - name: Generate baseline
+        run: |
+          python3 scripts/generate_ubuntu_baseline.py \
+            --lts 22.04 24.04
```
MakisH, line 20: Since https://github.com/precice/precice/pull/2564, the baseline has shifted to 24.04, and we have added 26.04. For the current version (3.4.x), we still need to keep 22.04 listed.

Let's make this:


```suggestion
          python3 scripts/generate_ubuntu_baseline.py \
            --lts 22.04 24.04 26.04
```

.github/workflows/generate-baseline.yml
```diff
@@ -0,0 +1,33 @@
+name: Generate Ubuntu Baseline
+
+on:
+  workflow_dispatch:
+
+jobs:
+  generate:
+    if: github.actor != 'github-actions[bot]'
```
MakisH, line 8: Why this line?

.github/workflows/generate-baseline.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
MakisH, line 12: There is already a v6:


```suggestion
      - uses: actions/checkout@v6
```

MakisH (changes requested):
Thank you for contributing! We are still discussing in #617, but this PR already does something useful in the respective direction, and I think we can merge it independently of the conclusion.

This PR addresses #617 by automatically generating a list of relevant packages provided in an Ubuntu release with a Python script. A manually-triggered CI workflow executes that script and checks the respective YAML file into this repository.

The Python script executes an `apt-cache policy` for each package in a Docker container, which might be a bit overkill, but works. Inputs are specified in `dependency.yml`, the output is saved into `_data/ubuntu-baseline.yml`. This PR already adds such a file, which could in principle then be rendered on the website.

See some suggestions, which should all be easy to implement.

scripts/generate_ubuntu_baseline.py
MakisH, line 1: Ideally, I would put this script (everything you currently have into `scripts/`) into a directory specific to this script (`scripts/baseline/`).

scripts/generate_ubuntu_baseline.py
```diff
@@ -0,0 +1,124 @@
...
+        "--mock",
+        action="store_true",
+        help="Use mock versions instead of querying Docker (useful for testing the pipeline).",
+    )
+    parser.add_argument(
+        "--lts",
+        nargs="+",
+        required=True,
+        help="Ubuntu LTS versions to query (e.g. 22.04 24.04)",
+    )
```
MakisH, line 114: Why restrict that to LTS releases? I would rather name the option `--release`. For example, `--lts 25.10` also works, even though that is not an LTS release.

scripts/generate_ubuntu_baseline.py
```diff
@@ -0,0 +1,124 @@
...
+    cmd = [
+        "docker",
+        "run",
+        "--rm",
+        f"ubuntu:{ubuntu_version}",
+        "bash",
+        "-c",
+        f"apt-get update -qq && apt-cache policy {package} | grep Candidate | awk '{{print $2}}'",
+    ]
+    raw = run_command(cmd)
```
MakisH, line 49: Before running this command, I would expect some output with the command to run, especially since this might take a while. This would also be useful for debugging and for understanding what this script does.

fsimonis (changes requested):
As described in the issue, we have an API key for pkgs.org.
There is no reason to probe a Docker container to get the package versions.

Also, the distinction between optional and required packages is unnecessary. It is handled where we display the information.

Furthermore, we are interested in the version of the packaged software, not the package. Meaning that we want `1.83.0`, not `1.83.0.1ubuntu2`.

MakisH:
I am closing this PR, since the parent issue still needs discussion/conclusion and this PR is far from a state to get merged. Happy to review any follow-up PR following a more lightweight approach.

Issue #411, opened 2024-05-29
Consistent URL endings


fsimonis:
We are mixing `.html` with more modern endling-less URLs, which leads to confusion in search engines as they are indexed as separate pages.

We should decide which version we want to stick to, make it consistent and ideally enforce it.

Example:
* https://precice.org/about and https://precice.org/about.html serve the same `about.html`.
* https://precice.org/ , https://precice.org/index and https://precice.org/index.html serve the `index.html`

What we need to choose:
1. without ending https://precice.org/about
2. with ending https://precice.org/about.html

What we need to do after the decision:
* Adjust canonical metadata
* Fix sitemap (may require changing permalinks)
* Fix links

MakisH:
What happens to links from external resources? Can we have redirects both in case we rename `about -> about.html` and `about.html -> about`?

fsimonis:
I am honestly not sure. This behaviour seems to dependent on how the HTTP-server is configured.

Aayushman-nvm:
After reviewing the current setup, it appears that `.html` URLs are explicitly enforced via `permalink: about.html`, with extensionless paths configured as redirects.

Since migration to **Hugo** is being considered, and Hugo defaults to extensionless (“pretty”) URLs, it may be cleaner to standardize on extensionless URLs already in Jekyll.

This would:

- Avoid changing canonical URL policy twice (once now, once during migration)  
- Align early with Hugo’s default behavior  
- Simplify migration  
- Reduce long-term maintenance complexity  

### A possible approach:

- Remove explicit `.html` permalinks in Jekyll  
- Use pretty permalinks (`/about/`) as canonical  
- Redirect `/about.html` → `/about/`  
- Update canonical metadata and sitemap accordingly  

During migration to Hugo, this structure would map naturally to Hugo’s default output, with optional aliases added for any remaining legacy `.html` URLs.

Would it make sense to standardize on extensionless URLs already at the Jekyll stage to avoid duplicated effort later?

MakisH:
@Aayushman-nvm I commented on the issue. Besides the comments there, I think it is a bit too early to address this, and I would rather only look at it in the context of the migration. There are no related issues with the running system at the moment.

Aayushman-nvm:
@MakisH I understand, thanks for the review by the way. 😄

MakisH:
In the Hugo state (as of #979), http://localhost:1313/about.html would automatically redirect to http://localhost:1313/about/. The same for all other pages I tested.

PR #827, opened 2026-03-01
fix: Consistent URL endings concept (#411)

Aayushman-nvm:
## Summary

Standardizes the `About` page to use an extensionless permalink (`/about/`) and fixes relative asset paths that broke when switching away from `/about.html`.

---

## Changes

- **Permalink**: `about.html` → `/about/`, with a `redirect_from: /about.html` redirect
- **Canonical URL**: Replaced `.html` removal hacks with `{{ page.url | absolute_url }}`
- **Asset paths**: Fixed relative `images/` and `css/` references using `| relative_url` to prevent `/about/images/...` misresolution
- **Internal links**: Replaced `.html` links (e.g. `privacy.html`) with extensionless equivalents (e.g. `/privacy/`)

---

## Note

- The scope of this PR is intentionally limited to the About page to keep the change small, isolated, and easy to review.
- The same pattern can be systematically replicated across other pages to fully address issue #411.
- Updated `head.html` to remove the assumption of a flat URL structure (i.e., direct `.html` file endings). The previous logic did not account for nested paths or directory-based URLs.
- Similar adjustments may be required in other templates where this flat URL assumption exists.

---

## Why

- Removes flat-file permalink assumptions
- Compatible with `permalink: pretty` and Hugo's default URL structure
- Reduces migration friction and improves long-term URL hygiene

---

## Testing

- `/about/` loads correctly
- `/about.html` redirects to `/about/`
- All assets and internal links resolve without issues

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,22 +1,23 @@
 <meta charset="utf-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1">
+
 {% if page.summary %}
 <meta name="description" content="{{ page.summary | strip_html | strip_newlines }}">
 {% endif %}
 <meta name="keywords" content="{{page.tags}}{% if page.tags %}, {% endif %} {{page.keywords}}">
 <meta name="theme-color" content="#0A76BB">
-<meta name="msapplication-square310x310logo" content="images/icon-310x310.png">
+<meta name="msapplication-square310x310logo" content="{{ '/images/icon-310x310.png' | relative_url }}">
 
 <title>{{ page.title }} | {{ site.site_title }}</title>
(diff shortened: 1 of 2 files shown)


MakisH:
@Aayushman-nvm I don't disagree with the need for some of the changes, but this PR touches too many different things at once, for which reason there are already conflicts.

Instead of closing this, I suggest selecting one of the aspects (the most important, in your opinion), and focusing the PR to that aspect. There is no reason to touch all of these now. Please also don't touch the formatting of lines you are not changing for the needs of the PR.

Make sure that the "files changed" view only includes the intended changes: https://github.com/precice/precice.github.io/pull/827/changes

MakisH:
As elaborated in https://github.com/precice/precice.github.io/pull/827#issuecomment-4061367196, I am closing this as it is not easy to merge it.

While smaller PRs in this direction would generally be welcome, this will need some coordination with the planned framework migration.

Thank you, in any case, for contributing!

PR #831, opened 2026-03-04
Fix infinite duplication bugs in subprojects.rb plugin

praaatap:
## Description

This PR fixes two mistakes in the [_plugins/subprojects.rb] Jekyll plugin. These mistakes caused file filtering and endless asset duplication.

### 1. Correct Path Context for Directory Filtering

The plugin checked if a file is a directory using `File.directory?(f)`.. `f` is just the filename, like `logo.png` not the full path. So Ruby checked the project root of the `images` subdirectory.

- **Fix:** I used `File.join(images, f)` to give the path.

### 2. Fixed Infinite Duplication in Static Files

The code that tried to skip existing files did not work. The `existing_statics` Set had paths without a leading slash. The code checked against `static_file.relative_path` which has a leading slash. This meant the same files were re-registered in every build loop.

- **Fix:** I added `.delete_prefix('/')` to the comparison so both paths match.

## Impact

- **Prevents "Endless Rebuild Loops"**: Now Jekyll correctly identifies existing files and stops re-registering and re-writing the same assets.

- **Performance**: We save memory and time by not processing the same assets multiple times.

## Verification

- I checked that the Ruby syntax is correct.

- I confirmed that path comparisons now consistently exclude the leading slash, for matching.


MakisH:
@praaatap I think I understand what you are trying to convey (the PR description needs some editing), but I am wondering how you see the issue: at the moment, both `master` and this branch build correctly for subprojects. How can I see the before and after?

Did you face this issue in practice, or was it a suggestion of a linter? If yes, which?

pratapsisodiya:
@MakisH I identified these issues through reading the code rather than through the site failing in real life. The issues are tricky and will not always cause the site to fail, although they can cause the same files to be registered multiple times in memory. You can verify this by adding a debug statement before the site.static_files += statement and seeing the same files appear multiple times

MakisH (approved):
I understand the change on line 72, but I am not convinced that the change on line 44 is needed.

In any case, it does not seem to break anything, and we will migrate away from Ruby soon anyway.

PR #832, opened 2026-03-04
Unify GitHub Pages deployment action version to @v4

praaatap:
## Description
This PR unifies the version of the `peaceiris/actions-gh-pages` GitHub Action used for website deployment. 

Currently, [deploy.yml]was using `@v3`, while [testing.yml] was already updated to `@v4`. This change ensures consistency across environments and adopts the latest stable version for production deployments.

## Changes
- **[MODIFY]** [.github/workflows/deploy.yml]: Upgraded `peaceiris/actions-gh-pages` from `v3` to `v4`.

## Verification
- Verified version consistency with the existing staging/testing workflow (`testing.yml`).
- Syntax check on the workflow file confirmed.


MuhammadAashirAslam:
Hey @praaatap , I noticed this as well, but I think this workflow is not currently being used since it is disabled. You can see in the Actions tab that the “Build and Deploy the Testing Website” workflow is disabled, so aligning the versions here might not have any real effect right now. But still good PR 😊

MakisH:
Thank you for the contribution, @praaatap. This is valid, but I was also wondering why it only updates the specific action. Note that the quotes change is also not relevant and not needed.

Triggered by this PR, I directly bumped all actions in all workflows in https://github.com/precice/precice.github.io/pull/852.

> I think this workflow is not currently being used since it is disabled. You can see in the Actions tab that the “Build and Deploy the Testing Website” workflow is disabled

Not sure what happened here, but the workflow is actually actively used.

Issue #405, opened 2024-05-07
Documentation footer


fsimonis:
In addition to the edit me button at the top of the page and the general information in the footer, it may also be useful to have an additional documentation footer similar to the one on GitHub.

_This is an extension of #46_

### Content:

* Quick way to give feedback using a form: I expeced this, I missed that, I liked that
* "Edit me" button to quickly edit the page and date of the last edit (Same as top of the page)
* Quick link to the [FAQ](https://precice.discourse.group/tag/faq) + discourse category based on where we are in the documentation:

Documentation Section | Discourse Category
--- | ---
[Fundamentals](https://precice.org/configuration-acceleration.html#) | Is preCICE for me?
[Installation](https://precice.org/configuration-acceleration.html#) | Installing preCICE
[Configuration](https://precice.org/configuration-acceleration.html#) | Using preCICE
[Tooling](https://precice.org/configuration-acceleration.html#) | ?
[Provided adapters](https://precice.org/configuration-acceleration.html#) | Official adapters and tutorials
[Couple your code](https://precice.org/configuration-acceleration.html#) | Using preCICE
[Running simulations](https://precice.org/configuration-acceleration.html#) | Using preCICE
[Dev docs](https://precice.org/configuration-acceleration.html#) | _hide_
[Documentation meta](https://precice.org/configuration-acceleration.html#) | _hide_

### Example GitHub footer

_Note: The suggestion is supposed to be less busy_

![image](https://github.com/precice/precice.github.io/assets/13552216/3a3c6ddc-2baa-4f2e-8370-b01841bcd37d)

![image](https://github.com/precice/precice.github.io/assets/13552216/03f20195-0e9d-4edb-92e5-e5b251f5de2a)

### Extra

Something similar may be useful for the tutorials too.
This is supposed to go above the normal footer.

MakisH:
I like it! And I think we would put exactly the same in the tutorials section.

As an additional motivation for contributing, we could also include all usernames that have edited a page at the bottom, although that would be:

- complicated
- noisy
- and probably leave out many people that contributed to the earlier wiki

fsimonis:
> include all usernames that have edited a page

This would look a bit weird as the same names would repeat over and over.
Also, git gives you an email address, which still needs to be looked up for a username via the GitHub API, which is rate-limited, meaning it likely needs to be cached. Sounds like a complex feature.

MakisH:
[...]

PR #834, opened 2026-03-05
Add: A light weight documentation footer above the site footer. (Issue: #405)

Aayushman-nvm:
## Documentation Footer

Adds a lightweight documentation footer above the site footer.

Addresses Issue #405 

Features:
- "Edit this page" GitHub edit link
- Last updated date
- Quick help links (only FAQ for now - Discourse links planned for later... perhaps after review)

### Design considerations

The implementation intentionally remains minimal and migration-safe since the documentation site will migrate from Jekyll to Hugo in the near future.

Features that require additional infrastructure (feedback forms, contributor lists, GitHub API usage) are intentionally not implemented here and can be reconsidered after migration.

### Future improvements

Once the documentation system migrates to Hugo, themes such as Docsy could provide additional functionality including:
- feedback widgets
- enhanced navigation
- richer page metadata

### Note:

This PR intentionally omits section-aware Discourse links (e.g. linking to "Installing preCICE" when browsing installation docs). This would require a `section:` front matter field to be added across doc pages. Happy to open a follow-up issue for this if the team think it's worth pursuing.

`_data/discourse_map.yml` is included as a reference for the planned section-aware Discourse links but is not actively used in this PR.

`content/docs/fundamentals/fundamentals-overview.md` is the only file where this appears for now. Will add it to other doc pages similarly.

### Screenshot:

image

### DIFF
--- a/_data/discourse_map.yml
+++ b/_data/discourse_map.yml
@@ -1,0 +1,27 @@
+fundamentals:
+  name: "Is preCICE for me?"
+  url: "https://precice.org/docs.html"
+
+installation:
+  name: "Installing preCICE"
+  url: "https://precice.org/installation-overview.html"
+
+configuration:
+  name: "Using preCICE"
+  url: "https://precice.org/configuration-overview.html"
+
+tooling:
+  name: "Using preCICE"
+  url: "https://precice.org/tooling-overview.html"
+
+adapters:
+  name: "Official adapters and tutorials"
+  url: "https://precice.org/adapters-overview.html"
+
+couple-your-code:
+  name: "Using preCICE"
+  url: "https://precice.org/couple-your-code-overview.html"
+
+running:
+  name: "Using preCICE"
+  url: "https://precice.org/running-overview.html"
--- a/_includes/doc_footer.html
+++ b/_includes/doc_footer.html
@@ -1,0 +1,33 @@
+<div class="doc-footer">
+  <div class="doc-footer-inner">
+
+    <div class="doc-footer-col">
+      <p class="doc-footer-heading">Improve this page</p>
+      <a
+        class="doc-edit-btn"
+        href="https://github.com/precice/precice.github.io/edit/master/{{ page.path }}"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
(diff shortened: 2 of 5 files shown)


_data/discourse_map.yml
MakisH, line 1: Since this file is not used, I would rather leave it out in this PR. But having such a list in a YAML file is probably the way to go, in the same way as we have other collections.

_includes/doc_footer.html
```diff
@@ -0,0 +1,33 @@
...
+
+    <div class="doc-footer-col">
+      <p class="doc-footer-heading">Improve this page</p>
+      <a
+        class="doc-edit-btn"
+        href="https://github.com/precice/precice.github.io/edit/master/{{ page.path }}"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <svg width="15" height="15" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><path d="M11 4H4a2 2 0 0 0-2 2v14a2 2 0 0 0 2 2h14a2 2 0 0 0 2-2v-7"/><path d="M18.5 2.5a2.121 2.121 0 0 1 3 3L12 15l-4 1 1-4 9.5-9.5z"/></svg>
```
MakisH, line 12: Having an SVG path described directly here sounds efficient, but rather cryptic. Since we already use FontAwesome elsewhere, I would rather use glyphs from there, instead (see, e.g., header CSS).

content/docs/fundamentals/fundamentals-overview.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: The preCICE documentation
+layout: docs
```
MakisH, line 3: Why implement this like this and not similarly to the main footer? Touching every file to enable a technical element sounds like a bit too much duplication.

css/modern-business.css
MakisH, line 1: Something looks odd regarding the sizes:

Image

MakisH (commented):
This is an nice to have proof-of-concept, but it is currently mainly that.

Since this would need quite some effort to bring into shape and integrate into all pages, while we would need to find HUGO-oriented solutions later on that we would not need to maintain (e.g., from the respective theme or from a plugin), I am closing this PR.

In any case, see some specific comments, and thank you for contributing!

PR #718, opened 2026-02-17
Restructure API page

carme-hp:
Fixes https://github.com/precice/precice.github.io/issues/715

PR #836, opened 2026-03-06
Update the C++ API location / URLs

MakisH:
The "Location" of the C++ API currently links to `Participant.hpp`. This is the direct link to the API (so, easy for quick access), but it is not the header that we want people to include. This is a bit confusing.

This PR:

- As "location", links to the directory https://github.com/precice/precice/tree/main/src/precice, similarly to the C/Fortran bindings
- Adds a link to "header" (`precice.hpp`)
- Keeps the link to "API" (`Participant.hpp`)
- Fixes the locations of the C/Fortran bindings to remove the GitHub URL parts from the locations

From a discussion with @dabele.

@fsimonis what do you think?

### DIFF
--- a/content/docs/couple-your-code/couple-your-code-api.md
+++ b/content/docs/couple-your-code/couple-your-code-api.md
@@ -13,7 +13,7 @@ The definite documentation of the C++ API is available on [the preCICE API doxyg
 
 | Language       | Location                                                                                    | Installation                                                                  |
 |----------------|---------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


fsimonis:
It still looks mixed to me.

Maybe the issue is the heading `Location`. If we change that to `Repository`, then it's clear what we mean and C++, C and Fortran would point to precice/precice.

Instead of the actual location, we could point to the API docs instead.

Not all bindings have a clear location, as they may contain out of multiple files and folders.

Also doesn't this collide with #718?

MakisH:
Right, I forgot that #718 was still open. I ported the changes there, which should cover all points.

PR #837, opened 2026-03-06
docs: update external http links to use https

praaatap:
### Description
This PR updates several external links in the documentation to use secure `https://` instead of `http://`. 

This is a small maintenance improvement that:
* Avoids mixed-content warnings in modern browsers
* Improves overall security
* Follows best practices for SEO

**Links updated:**
* `www.boost.org`
* `www.netlib.org/lapack/`
* `www.calculix.de`
* `www.nutils.org`
* `www.sciencedirect.com`


MakisH (approved):
Makes sense, thanks!

PR #839, opened 2026-03-07
Add organizers email address for workshop 2027

MakisH:
@Crazy-Rich-Meghan is this correct and ready?


Crazy-Rich-Meghan:
Yes, this is correct. I have tested the email address and it works. Thanks!

Issue #838, opened 2026-03-06
Repository structure overview in contributor documentation


Aayushman-nvm:
### What should be improved?

While exploring the repository as a contributor, I noticed that understanding the project structure takes some time, especially for contributors coming from JS/TS ecosystems (Web dev in general).

Since the project currently uses Jekyll (and will migrate to Hugo later), some directories and conventions may not be immediately obvious to new contributors.

It might be helpful to include a short "Repository Structure" section in the documentation explaining things like:

- What each top-level directory contains
- Purpose of key folders (e.g. _layouts, _includes, _data, etc)

Many contributors (myself included) tend to rely on repository documentation more than upstream framework docs when getting started, so this could reduce onboarding friction.

This may require familiarity with the repository history and architecture, so maintainers or long-time contributors might be best suited to define the initial structure.

Aayushman-nvm:
# An example of it could be something like this:

``` bash
├── _layouts/
├── _includes/
├── _data/
└── and more...
```
### Folder Explanation
`_layouts/`

Contains the page layout templates used across the site.

Example:
`default.html` :- Base layout used by most pages

These layouts define the overall HTML structure and usually include headers, footers, and content placeholders.

hanzalahwaheed:
We can introduce an `ARCHITECTURE.md` file maybe

Aayushman-nvm:
Anything is fine as long as it's not an AI-generated PR and it's written by someone who has taken the time to understand the repository. They don't need to know the repo perfectly, but they should be familiar enough with the structure to know what each folder contains and how things are organized.

I’m clarifying this because issues like this can sometimes attract low-effort PRs otherwise 😅

SKM2227229725:
Hi, I’m interested in working on this issue.
I’m currently exploring the repository structure and running the site locally to understand how the folders and layouts work.

Once I understand the structure properly, I’ll propose a documentation update explaining the repository anatomy for new contributors.

SKM2227229725:
\assign me

i40055392-oss:
I think adding a clear repository structure overview in the contributor documentation will be very helpful for beginners.

I would like to work on this issue by documenting the top-level folders and explaining their purpose in a simple way. This should make it easier for new contributors to understand the project quickly.

I will try to make the documentation beginner-friendly so that new contributors can easily understand the project structure without confusion.

i40055392-oss:
> I think adding a clear repository structure overview in the contributor documentation will be very helpful for beginners.
[...]

PR #843, opened 2026-03-10
docs: add repository anatomy to contribute-to-precice #838 

dhanushsaimudari:
add repository anatomy to contribute-to-precice  resolving issue #838 

for better understanding for new contributor who want understand the House Analogy of website repo

repo anatomy

### DIFF
--- a/content/community/contribute/community-contribute-to-precice.md
+++ b/content/community/contribute/community-contribute-to-precice.md
@@ -40,6 +40,27 @@ and merge them as soon as possible.
 These pages are written in [Markdown](https://guides.github.com/features/mastering-markdown/) (a very easy language)
 and we also have a [cheatsheet](docs-meta-cheatsheet.html) specifically for this website.
 
+## Repository Anatomy
+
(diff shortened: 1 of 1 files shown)


Aayushman-nvm:
Hi @dhanushsaimudari 
Thanks for the PR, but I see few mistakes... did you fully rely on some tool or did you actually go through the repo?

MakisH:
I removed some files in https://github.com/precice/precice.github.io/pull/854, this is (at least partly) why the content here looks a bit different.

But definitely, some sections are "fluffier" than they need to be.

Aayushman-nvm:
@MakisH thanks for the reference, I understand now

But still... for an instance look at `assets/`
It doesnt have `css` files

MakisH:
Given that:

- a lot of restructuring is about to happen,
- this needs updates after #854 
- the descriptions would generally need to be rewritten
- it is currently a rather standard Jekyll project

I think the best way forward is to close this PR for now.

Still, point taken that repositories could include something like this. There is always the question of how quickly they become outdated.

PR #846, opened 2026-03-11
Add profiling merge example without arguments

fsimonis:
The first example used merge with arguments, which can be misunderstood to be the Participants to merge.

We now
- first show the non-argument version that should be enough in most cases.
- then explain what arguments mean
- and finally show the example with arguments

### DIFF
--- a/content/docs/tooling/tooling-performance-analysis.md
+++ b/content/docs/tooling/tooling-performance-analysis.md
@@ -222,17 +222,18 @@ where the naming pattern is `participant-rank-file_number`. To find and merge th
 $ ls
 A
 B
-$ precice-cli profiling merge A B
-Searching A : found 1 files in A/precice-profiling
-Searching B : found 1 files in B/precice-profiling
-Found 2 unique event files
+$ precice-cli profiling merge
+Found 2 files in .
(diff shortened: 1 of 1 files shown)


content/docs/tooling/tooling-performance-analysis.md
```diff
@@ -228,6 +229,24 @@ profiling.db
 
 The merge command searches passed directories for the event files.
```
carme-hp: what is the default directory if nothing is specified? I get that you are looking at the current directory, but I think it is not obvious that it searches inside subdirectories as well.

carme-hp (commented):
Thanks! I'd just asked you to add more on the default.

content/docs/tooling/tooling-performance-analysis.md
fsimonis:
```suggestion
The merge command searches the current directory recursively for event files.
You can also pass individual directories to be searched, or individual event files to use.
```

fsimonis: @carme-hp You are right. This isn't clear. Fixed it.

PR #847, opened 2026-03-12
ci: automate gsoc label assignment

aadhil2k4:
## Summary
- Added a PR template by referring existing templates in preCICE repos.
- Added workflow to automatically assign GSoC labels via PR template checkbox.

## Motivation
There is no standard PR template in this repositiory that new contributors can follow.
GSoC labels are currently assigned manually. Considering the number of PRs created each day, this task can become overwhelming. Automating this process can reduce the workload for maintainers.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,32 @@
+## Main Changes of this PR
+
+
+## Motivation and additional information
+
+<!--
+Short rational why preCICE needs this change. If this is already described in an issue a link to that issue (closes #123) is sufficient.
+-->
+
+## Steps to Test
+
+
+## Screenshots (if applicable)
+
+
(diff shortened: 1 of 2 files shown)


pratapsisodiya:
hi @aadhil2k4 i think the both changes of ci and pr templates shoudb be seperate pr for both

MakisH:
Thank you for contributing.

Overall, this is an interesting idea, which we did not end up implementing in time. If we apply again next year or for similar programs, we could use something like this. But then, we need to define what labels we use and how. For example, do we now keep adding the GSoC label to new PRs from the same people that previously applied for GSoC? Do we have a new label per year?

I see why you needed to define a PR template to make the workflow work, but then the order is wrong. Let's revisit this again in the future if needed.

Issue #399, opened 2024-05-03
Formalize strategy for upgrading minimum versions of dependencies


fsimonis:
We have some statements regarding supported distro versions, but we have no clear statement on how we select and upgrade our minimum required versions.

This is essentially a follow-up of https://github.com/precice/precice.github.io/issues/16.

We need a statement on:
* What dictates our minimum required versions.
* That minimum required version updates are not considered breaking releases.
* What upper bound we actively test.
* How forward-compatible the project is and if changes warrant a bugfix release.
* What distros we directly support (2 LTS of Ubuntu)
* How we deal with different MPI implementations (OpenMPI vs MPICH on Fedora, what about Intel MPI / Windows MPI?)

Optionally:
* What distros we indirectly support if they meet the version requirements (Debian, Arch, Fedora)
* How we handle required bug fixes for indirectly supported distros

Where shall we put this in the docs? Under fundamentals? How shall we call this page?

MakisH:
> Where shall we put this in the docs? Under fundamentals? How shall we call this page?

Why not just a section in https://precice.org/installation-source-dependencies.html?

I would definitely not put it in the "Fundamentals": sounds rather a technical detail to me. 

Alternatively, it could be a dedicated page under "Dev docs", just before or part of https://precice.org/dev-docs-release-strategy.html

premxpagar:
Hi! I added this documentation section to clarify how preCICE selects and updates minimum dependency versions.

This is related to the discussion in https://github.com/precice/precice.github.io/issues/399 about formalizing the dependency version strategy.
Let me know if the section should be placed somewhere else in the docs.

MakisH:
Hi @prem-pagar,

this issue is still open because it needs discussion. I don't think that external contributions are useful here, since the blocking factor is not the content, but the decision.

MakisH:
Also related: https://github.com/precice/precice.github.io/issues/617

fsimonis:
> Why not just a section in https://precice.org/installation-source-dependencies.html?

> Alternatively, it could be a dedicated page under "Dev docs", just before or part of https://precice.org/dev-docs-release-strategy.html

IMHO, both of these look too hidden.

> I would definitely not put it in the "Fundamentals": sounds rather a technical detail to me.

I disagree on this. Providing clear information on OS support is important when deciding on your software stack for a potentially long-term project.
It is also a good place to mention the support license for exotic wishes.

PR #848, opened 2026-03-12
docs: add dependency version strategy documentation

premxpagar:
This PR adds a short documentation section explaining how preCICE selects and updates the minimum supported versions of its dependencies.

The intention is to clarify the project's dependency policy for users and contributors. It describes how minimum versions are chosen (mainly following supported Ubuntu LTS releases), when they may be updated, and what environments are regularly tested.


MakisH:
Thank you for the contribution. The content is not that far from what we are actually doing right now, but some decisions are still pending in #399. I leave this open for now.

MakisH:
It looks like we will need to anyway rewrite this differently, more concisely, and in a different place once we have a decision. Closing to keep a better overview of PRs.

PR #857, opened 2026-03-16
github-queries.js: declare constants as const

praaatap:
## Title
Remove `var` and use `const`, add semicolons

## Description
This change enhances the JavaScript coding style and ensures the variables are safe.

### Changes made
1. Replaced `var` with `const` in variable declarations.
2. Declared variables correctly instead of leaving them undeclared.
3. Added missing semicolons to adhere to the standard JavaScript coding style.
4. No functionality was altered. The code functionality is still the same.

### Why this change
`const` is safer than `var` because it does not allow the variables to be reassigned by mistake.  
Adding semicolons also enhances the code readability.

### Summary
This commit only enhances code quality. It updates the coding style of the variables and formatting without altering the functionality of the program.


MakisH (approved):
Thanks! I resolved the conflicts.

This is rather trivial and does not affect functionality, but why not.

PR #106, opened 2021-12-10
Make native doxygen website more visible

ajaust:
This adds an additional section referring to the native C++ API and moves the Doxygen link there. This should make the API documentation more visible and easier to find as not everyone might know Doxygen. Before, the link to the API documentation was hidden in the running text. The change also makes the representation of available APIs more consistent as all APIs are represented by a tables now.

PR #870, opened 2026-03-24
Rework the API links order, name, and targets

MakisH:
Follow-up of #718:

- Renames `Location` to `Source code`
- Moves it to the bottom of each section, to not confuse users
- Changes the C++ API "location" from https://github.com/precice/precice/tree/main/src/precice/Participant.hpp to https://github.com/precice/precice/tree/main/src/precice. The intention here is to have an entry point to find the rest (and was already pointing to `Participant.hpp` before #718).

Original table added in https://github.com/precice/precice.github.io/pull/106


davidscn (approved):

fsimonis:
I don't think moving this is a good solution either.
IMHO these should all simply link to the repository and relevant information should be in the repo Readme. If someone wants to look at source, then they should look at the repository.

MakisH:
Merging this to move forward; follow-up PRs are welcome.

From my side, I like seeing such an overview in a project. The respective `README.md` file(s) can also additionally point to that information as well.

PR #902, opened 2026-05-20
Update mapping config

davidscn:
Consequence of merging https://github.com/precice/precice/pull/2536

This change is not 100% accurate, but probably easier to understand rather than saying: consistent in version 3.4.1 and conservative in 3.5.0. I would simply merge this after the next release.


MakisH (commented):
In some other parts of the documentation, we are using the (newer) `{% version %}` [alerts](https://precice.org/docs-meta-cheatsheet.html#version-information), which also automatically change the text to signify when a release is pending.

I see the difficulty in writing this concisely. The way I would do it would be to leave the table row as-is and add directly below the table something like

```
{% version 3.5.0 %}
The conservative variant of the `rbf-pum-direct` mapping with Kokkos is supported since preCICE v3.5.0. The consistent variant is available since v3.4.1.
{% endversion %}
```

By the way: There should be no hyphen in `Kokkos-Kernels`.

Side note: I would like to get all (if possible) PRs in this repository merged before we start the migration to HUGO.

MakisH (approved):

Issue #867, opened 2026-03-23
Funding page: uneven card gaps and mismatched heights


NishantSinghhhhh:
### Page URL

https://precice.org/about.html

### Description

The funding cards on the `/funding` page have two related layout problems:

1. **Unequal vertical gaps between cards** — the spacing between rows is inconsistent. Some card pairs have a large gap between them while adjacent pairs are much tighter, making the grid feel unbalanced.

2. **Cards are not equal height within a row** — when one card in a row has more content than the other (e.g. a longer description or an extra field like "Funding number"), the two cards end up at different heights instead of stretching to match each other. This breaks the visual grid alignment.

### Steps to Reproduce

1. Navigate to the Funding page
2. Observe the two-column card grid
3. Compare the vertical gap between different card rows — the spacing is visibly uneven
4. Compare card heights within the same row — cards with shorter content are shorter than their neighbors

### Expected Behavior

- All cards in the same row should be equal height (stretch to match the tallest card)
- Vertical and horizontal gaps between all cards should be uniform throughout the grid

### Suggested Fix

we can use grid or flexbox to solve the issue

### Screenshots

https://github.com/user-attachments/assets/2885bb63-4fca-4806-99e6-49984a126965

<img width="1410" height="852" alt="Image" src="https://github.com/user-attachments/assets/b1d8c787-8725-4125-8483-d02bdf6d2597" />

<img width="1410" height="852" alt="Image" src="https://github.com/user-attachments/assets/0c688c70-58dd-455f-a613-797373ac274c" />

Abhijit1018:
Can i solve this issue. i have an idea to solve it and that would help me in gsoc also. /assign

MuhammadAashirAslam:
@Abhijit1018 you can discuss your approach in the comments or have a branch in your fork but for now no PRs are accepted.
More information at : https://precice.discourse.group/t/next-steps-in-gsoc/2827

Abhijit1018:
OK let me see and also can you share me what can do now for GSOC

MuhammadAashirAslam:
First of all, you can introduce yourself in the [Form](https://precice.discourse.group/c/jobs/13)

You can also fork the repository and implement solutions (of the issues) or features you like in separate branches, then reference them in your post on the forum .

However, due to high traffic, most of the bigger issues have already been solved, so there are only a few good first issues left.

MakisH:
[...]

PR #921, opened 2026-05-28
Migration from Bootstrap3.3.7 to Bootstrap 5.3.8

MuhammadAashirAslam:
- Upgraded Bootstrap from 3.3.7 to 5.3.8 CDN bundle (CSS & JS) in head layouts and removed obsolete IE8/9 compatibility layers.
- Simplified HTML layouts by deleting nested row/column wrappers ("div-soup") around section headers, paragraph tags, and obsolete caret helper elements.
- Replaced custom CSS classes with modern native Bootstrap 5 grid, flexbox, and alignment utilities (e.g., col-flex -> d-flex flex-column, vertical-align -> d-flex align-items-center, no-margin -> m-0).
- Consolidated main styles by deleting modern-business.css and merging rules into customstyles.css.
- Purged duplicate CSS rules, legacy browser vendor prefixes, and obsolete panel/card dual-class selectors from stylesheets.
- Compensated for spacing changes by shifting the 20px bottom-margin from the removed headers globally to the .section-header class.
- Realigned page footer components, custom workshops, and team grids using standard Bootstrap 5 container, rounding, and spacing parameters.

Documentation:
- BS5 Migration: https://getbootstrap.com/docs/5.3/migration/
- Grid System: https://getbootstrap.com/docs/5.3/layout/grid/
- Cards (replacing Panels): https://getbootstrap.com/docs/5.3/components/card/
- Spacing: https://getbootstrap.com/docs/5.3/utilities/spacing/
- Flexbox: https://getbootstrap.com/docs/5.3/utilities/flex/
- Dropdowns: https://getbootstrap.com/docs/5.3/components/dropdowns/

PR #922, opened 2026-06-01
Migrate from Bootstrap 3.3.7 to Bootstrap 5.3.8

MuhammadAashirAslam:
- Upgraded Bootstrap from 3.3.7 to 5.3.8 CDN bundle (CSS & JS) in head layouts and removed obsolete IE8/9 compatibility layers.
- Simplified HTML layouts by deleting nested row/column wrappers ("div-soup") around section headers, paragraph tags, and obsolete caret helper elements.
- Replaced custom CSS classes with modern native Bootstrap 5 grid, flexbox, and alignment utilities (e.g., col-flex -> d-flex flex-column, vertical-align -> d-flex align-items-center, no-margin -> m-0).
- Consolidated main styles by deleting modern-business.css and merging rules into customstyles.css.
- Purged duplicate CSS rules, legacy browser vendor prefixes, and obsolete panel/card dual-class selectors from stylesheets.
- Compensated for spacing changes by shifting the 20px bottom-margin from the removed headers globally to the .section-header class.
- Realigned page footer components, custom workshops, and team grids using standard Bootstrap 5 container, rounding, and spacing parameters.
- Removed jquery.min.js and jqueryui.min.js and rewrote all jQuery-dependent JS to vanilla DOMContentLoaded APIs.
- Replaced the navgoco jQuery sidebar plugin with Bootstrap 5 native collapse component using data-bs-toggle attributes.
- Removed unused theme-blue.css and theme-green.css.

Documentation:

- BS5 Migration: https://getbootstrap.com/docs/5.3/migration/
- Grid System: https://getbootstrap.com/docs/5.3/layout/grid/
- Cards (replacing Panels): https://getbootstrap.com/docs/5.3/components/card/
- Spacing: https://getbootstrap.com/docs/5.3/utilities/spacing/
- Flexbox: https://getbootstrap.com/docs/5.3/utilities/flex/
- Dropdowns: https://getbootstrap.com/docs/5.3/components/dropdowns/
- Collapse: https://getbootstrap.com/docs/5.3/components/collapse/
- Navbar: https://getbootstrap.com/docs/5.3/components/navbar/


css/landing-page.css
MakisH, line 1: A misalignment at the bottom of the landing page:

Image

content/community/workshops/precice-workshop-2025.md
```diff
@@ -367,9 +367,9 @@ Did you know? Many past participants only discovered this workshop via a direct
 
 ## Supporters
 
-<div class="col-md-2 col-flex"><a href="https://www.dfg.de/en" title="DFG, German Research Foundation" class="no-icon"><img src="images/funding/dfg.jpg" alt="DFG logo" style="max-height:50px;"></a></div>
-<div class="col-md-6 col-flex"><a href="https://dtecbw.de/" title="dtec.bw" class="no-icon"><img src="images/events/precice2025/precice2025-logo-dtec.bw.jpg" alt="dtec.bw logo" style="max-height:50px;"></a></div>
-<div class="col-md-2 col-flex"><a href="https://dtecbw.de/home/forschung/hsu/projekt-hpcbw" title="hpc.bw" class="no-icon"><img src="images/events/precice2025/precice2025-logo-hpc-bw.png" alt="hpc.bw logo" style="max-height:50px;"></a></div>
+<div class="col-md-2 d-flex flex-column"><a href="https://www.dfg.de/en" title="DFG, German Research Foundation" class="no-icon"><img src="images/funding/dfg.jpg" alt="DFG logo" style="max-height:50px;"></a></div>
```
MakisH: The result now breaks over two rows:

Image

Not so critical to make the old workshop pages render the same, but we need a way to achieve the same result in future ones.

In the 2024 workshop, it looks fine.

content/docs/couple-your-code/step-by-step/couple-your-code-defining-mesh-connectivity.md
```diff
@@ -15,11 +15,11 @@ For volume coupling in 2D, mesh connectivity boils down to defining triangles an
...
 All kind of connectivity can be built up directly from vertices. Triangles and quads also allow us to define them using edge IDs.
 <ul id="apiTabs" class="nav nav-tabs">
-    <li class="active"><a href="#cpp-1" data-toggle="tab">C++</a></li>
-    <li><a href="#python-1" data-toggle="tab">Python</a></li>
+    <li class="nav-item"><a class="nav-link active" data-bs-toggle="tab" role="tab" href="#cpp-1">C++</a></li>
+    <li class="nav-item"><a class="nav-link" data-bs-toggle="tab" role="tab" href="#python-1">Python</a></li>
 </ul>
 <div class="tab-content">
-<div role="tabpanel" class="tab-pane active" id="cpp-1" markdown="1">
+<div role="tabpanel" class="tab-pane fade show active" id="cpp-1" markdown="1">
```
MakisH: The fading part is nice, but also hard-coded in each tab. Why? I guess this could easily become inconsistent if we start having fading in some tabs but not in other. I would rather remove the effect.

By the way, I have tested all tabs in the couple-your-code section and they work fine.

_includes/publication_panel.html
MakisH, line 1: That's a nice and long-overdue addition, thanks!

I wonder if we should use something more unique than the title (e.g., the filename where the data is drawn from), as the title could be duplicated (e.g., the slides and the post-proceedings paper of a conference). But that's not currently an issue.

_config.yml
```diff
@@ -73,6 +73,12 @@ defaults:
       search: true
       sidebar: home_sidebar
       topnav: topnav
+  -
+    scope:
+      path: "content/index.html"
+      type: "pages"
+    values:
+      topnav: topnav
```
MakisH: Why is this needed now but not previously?

_includes/topnav.html
MakisH, line 1: Clicking on the preCICE logo leads to http://127.0.0.1:4000/, which now does not work (only the explicit http://127.0.0.1:4000/index.html works).

_includes/sidebar.html
MakisH, line 1: In narrow resolutions, the sidebar is lacking some margin to the content:

Image

_includes/news_banner.html
MakisH, line 1: In narrow screens, some padding at the top is missing:

Image

_includes/sidebar.html
MakisH, line 1: Do we need to port any of these changes to https://github.com/precice/precice.github.io/blob/master/pdfconfigs/prince-list.txt ?

_includes/sidebar_recursive.html
```diff
@@ -0,0 +1,122 @@
+{% comment %}
+  Recursive sidebar: renders nested doc nav from _data/sidebars via folder.folderitems.
+  Parameters: item (folder node), level (1–4), parent_id (element id for data-bs-parent accordion group).
+  Branch rows use BS5 Collapse (data-bs-toggle, .collapse.show); leaves are plain nav-links.
+  Active state compares page.url (no leading slash) to each item url.
+{% endcomment %}
```
MakisH, line 6: Why is this needed, and how does it relate to `sidebar.html`?

Is this something standard, or something you implemented on top?

_includes/sidebar_recursive.html
```diff
@@ -0,0 +1,122 @@
...
+
+{% assign page_url_clean = page.url | remove_first: "/" %}
+
+{% assign itm = include.item %}
```
MakisH: As a style comment, I would prefer longer and clearer indices instead of constructed shortenings (in this case, `item` instead of `itm`).

_includes/head.html
```diff
@@ -50,17 +49,15 @@
...
-<link rel="stylesheet" href="{{ "css/customstyles-precice.css" }}">
-<link rel="stylesheet" href="{{ "css/boxshadowproperties.css" }}">
-<link rel="stylesheet" href="{{ "css/bookcover.css" }}">
+<link rel="stylesheet" href="{{ "css/customstyles.css" }}?v={{ site.time | date: '%s' }}">
+<link rel="stylesheet" href="{{ "css/customstyles-precice.css" }}?v={{ site.time | date: '%s' }}">
+<link rel="stylesheet" href="{{ "css/boxshadowproperties.css" }}?v={{ site.time | date: '%s' }}">
+<link rel="stylesheet" href="{{ "css/bookcover.css" }}?v={{ site.time | date: '%s' }}">
 <!-- Color styles -->
-<link rel="stylesheet" href="{{ "css/theme-precice.css" }}">
+<link rel="stylesheet" href="{{ "css/theme-precice.css" }}?v={{ site.time | date: '%s' }}">
```
MakisH, line 60: Why is this? To enforce reloading?

MakisH (commented):
I have gone through most of the changes and I have looked through all of the pages in "desktop view" and several pages in "mobile view". Here are some first comments (all the ones from https://github.com/precice/precice.github.io/pull/921 seem to have been resolved).

_includes/head.html
MuhammadAashirAslam: Yes , this enforces the site to serve the newer version (if changed) , so that user might not see the cached version from a previous build

css/customstyles-precice.css
```diff
@@ -30,6 +30,16 @@ main {
   margin-top:-20px;
 }
 
+/* Match the fixed collapsed navbar height on narrow screens. */
+@media (max-width: 1199.98px) {
```
MakisH, line 34: Probably for a next PR: these hard-coded widths in `media` queries look a bit hacky. We should add some CSS preprocessor, or in some other way make these centrally configurable.

content/index.html
```diff
@@ -37,13 +34,13 @@ <h2 class="subtitle no-anchor">The coupling library for partitioned multi-physic
...
-        <p class="no-margin">
+        <p class="m-0">
           The software offers convenient methods for transient equation coupling, communication, time interpolation, and data mapping.
         </p>
       </div>
       <div class="col-md-6">
-        <img class="img-responsive col-xs-12" src="images/lead-image-precice.svg" alt="visualisation of how preCICE couples different solvers">
-        <!-- https://stackoverflow.com/questions/21943108/bootstrap-3-img-responsive-images-are-not-responsive-inside-fieldset-in-firefox -->
+        <img class="img-fluid col-12" src="images/lead-image-precice.svg" alt="visualisation of how preCICE couples different solvers">
+        <!-- https://stackoverflow.com/questions/21943108/bootstrap-3-img-fluid-images-are-not-responsive-inside-fieldset-in-firefox -->
```
MakisH: Is this comment still needed? In any case, the URL should not be changed (it still works because of the encoded id).

content/index.html
```diff
@@ -166,55 +156,67 @@ <h2 class="section-header">Prepared for the next generation of multi-physics sim
...
-        <h2 class="section-header">Ready-to-use with your favourite open-source solver</h2>
-      </div>
-    </div>
-    <div class="row">
-      <div class="col-lg-12">
-        <p>Several free and proprietary codes are currently coupled with preCICE. Pick one of our official adapters below or check out our <a href="community-projects.html">community projects</a>.</p>
-      </div>
-    </div>
+    <h2 class="section-header">Ready-to-use with your favourite open-source solver</h2>
+    <p class="mb-4">Several free and proprietary codes are currently coupled with preCICE. Pick one of our official adapters below or check out our <a href="community-projects.html">community projects</a>.</p>
```
MakisH, line 159: An issue also with the current implementation: In too narrow screens, where only one card fits, still two are displayed. Maybe this is easy to fix.

css/customstyles-precice.css
```diff
@@ -8,58 +8,376 @@ body {
     display: flex;
     flex-direction: column;
 }
+
+.btn {
+  font-size: 14px;
+  line-height: 1.42857143;
```
MakisH, line 14: Why this overly-specific number?

css/customstyles-precice.css
```diff
@@ -8,58 +8,376 @@ body {
...
+  font-size: 14px;
+  line-height: 20px;
+  padding: 4px 10px;
+  font-weight: 500;
+  margin-top: 11px;
 }
+
 /* rightmost icons */
-.nav > li > a.nav-external {
+.nav :is(li > a, .nav-link).nav-external,
```
MakisH, line 126: There are ~250 lines of new rules here, some quite specific. Why are they needed now and not before?

We should simplify as much as possible. It is not so important to look identical as before, as it is to look fine and be maintainable.

MakisH (commented):
Some more comments. I have now looked through the whole code, even though I only skimmed through some longer CSS additions, which I think need to be simplified.

css/customstyles-precice.css
MuhammadAashirAslam: Many of these overrides were added to keep the layout close to Bootstrap 3 after switching to Bootstrap 5. Before, this was handled partly by Bootstrap 3 defaults and partly by the existing theme CSS. After the migration, plain Markdown tables/code blocks render differently because they do not automatically get [Bootstrap 5 table utilities](https://getbootstrap.com/docs/5.0/content/tables/).

If we are ok with it, then we can remove some intentional overriding of things. For example, removing the table overriding will reduce around ~172 lines.

Tables:

image

On original site:

image

css/customstyles-precice.css
MuhammadAashirAslam: This comes from Bootstrap 3’s default line height: 20px / 14px = 1.42857143. The whole .btn block copies the old Bootstrap 3 button metrics (font-size: 14px, line-height: 1.42857143, padding: 6px 12px, border-radius: 4px) so the we keep the previous button height and spacing.

css/customstyles-precice.css
MuhammadAashirAslam: Thanks! i will look in to this

content/index.html
MuhammadAashirAslam: Done

content/index.html
MuhammadAashirAslam: Removed the comment

_includes/sidebar_recursive.html
MuhammadAashirAslam: Sorry for the late reply, I missed this.

sidebar.html is the shell (title + top-level loop). sidebar_recursive.html does the nested part. It replaces navgoco after we dropped jQuery for BS5, walks _data/sidebars by calling itself, and sets active/expanded state in Liquid instead of client-side JS.

Same idea as Hugo's [menu templates](https://gohugo.io/templates/menu/) docs (recursive partial + active detection). We can then port it directly in the Hugo migration.

_includes/topnav.html
MakisH: With the current fix (https://github.com/precice/precice.github.io/pull/922/changes/683e28962596ce345ceeaa08079553292db7a364), note that the behavior is a bit different than in the current site: clicking on the logo on the current site makes the URL `https://precice.org/`, clicking on the current state of this PR makes it `https://precice.org/index.html`.

css/customstyles-precice.css
MakisH: As discussed: No need for the tables to look the same, but applying some common theme would be nice to make them more readable.

css/customstyles-precice.css
MakisH: In case of doubt: Fine to have this PR make things look reasonably similar and have a follow-up PR simplifying/cleaning up.

css/customstyles-precice.css
MuhammadAashirAslam: On it

MuhammadAashirAslam:
I found a few remaining Bootstrap 3-era assets, docs, and helper-class references. Tried my best to remove all of the Bootstrap 3 references 

[440bf9e](https://github.com/precice/precice.github.io/pull/922/commits/440bf9efde3c9d81f020edc23e79f4f95e285882) removes the old Glyphicons font files, since Bootstrap 5 no longer ships or uses them.
[904ad64](https://github.com/precice/precice.github.io/pull/922/commits/904ad64cb6a66d97fac694e01919f48df81ba17a) updates the local Bootstrap assets to 5.3.8 and makes the PDF build use the local stylesheet.
[9bc8e05](https://github.com/precice/precice.github.io/pull/922/commits/9bc8e057e4fb4390e7b5537f94c63b3ea63576f6) widens the PDF print layout so generated pages use more of the available page width.
[69b459b](https://github.com/precice/precice.github.io/pull/922/commits/69b459b46e7d2472c72f79200fdaac286fc08fac) updates stale documentation that still described the site as using Bootstrap 3.x.
[ad133b8](https://github.com/precice/precice.github.io/pull/922/commits/ad133b8ca21b930e07aafa1cf9be4c3ebcddf36b) removes remaining old Bootstrap references and migrates legacy image helper classes to Bootstrap 5 equivalents.

content/community/contribute/community-contribute-open-projects.md
MakisH, line 1: Let's not update the GSoC project descriptions in this PR. This is the idea at the time of announcement. I will archive them later.

_layouts/default_print.html
MakisH, line 1: Printing looks a bit broken now:

Image

MakisH (commented):
Thanks for the overview of updates and the clean history, this really helps!

Some comments on the latest changes.

content/community/contribute/community-contribute-open-projects.md
MuhammadAashirAslam: Yeah i sensed that this might be wrong so i removed it in https://github.com/precice/precice.github.io/pull/922/commits/1d7306c9c26f9f15d902c3e9dd1297952d99b2f1

MakisH: Something looks wrong: I can still see the changes in the latest state (https://github.com/precice/precice.github.io/pull/922/changes) and https://github.com/precice/precice.github.io/commit/1d7306c9c26f9f15d902c3e9dd1297952d99b2f1 does not seem to touch this file.

MuhammadAashirAslam: Removed it in https://github.com/precice/precice.github.io/pull/922/commits/556bf5ac1204771ba70be519c1295d25b244d8e6. I had removed it from that specific commit, but it turns out the file still had changes from an earlier commit that cant be removed by just editing the newest commit .

_layouts/default_print.html
MuhammadAashirAslam: Done (https://github.com/precice/precice.github.io/pull/922/commits/3b957c17ba1f65807f6614e4660c6217cc40ed53)

image

css/customstyles-precice.css
MuhammadAashirAslam: Removed the custom table overriding in https://github.com/precice/precice.github.io/pull/922/commits/7bac2361d186888f9478a989c0bbaf34a064a4e1

The new table styling uses Bootstrap 5’s built-in table classes , hope you like the colors/styling

_layouts/default_print.html
MakisH: Since we are on it: could we have the QR code only on the first page (or the bottom of the last)? In subsequent pages, it sometimes hides part of the text.

css/customstyles-precice.css
MakisH: Nice that it's much simpler now! Could we just switch the heading color to a dark gray again (or the same blue as we use elsewhere), to not conflict with the rest of the theme? Other than that, the defaults are fine.

MuhammadAashirAslam: Done (https://github.com/precice/precice.github.io/pull/922/changes/cf03c24814bc5e8e5300987a0000a95e20e55873) , just changed the table header color

_layouts/default_print.html
MuhammadAashirAslam: Done(https://github.com/precice/precice.github.io/pull/922/changes/896cd82cee94246e366b2ce7e560d00be599e10a) , qr is now only visible on the front page

MakisH (approved):
Thanks again for resolving all details. The unresolved conversations can be handled in follow-up contributions.

Btw, also closes https://github.com/precice/precice.github.io/issues/867 (nice!)

I think this is already good enough to merge. @fsimonis do you see anything critical?

MuhammadAashirAslam:
I rebased bootstrap5-migration onto the latest upstream master to keep the branch current and avoid merge conflicts ( turns out this was not needed 😅 )

MuhammadAashirAslam:
https://github.com/precice/precice.github.io/pull/922/changes/4288212eb03daa5ffd35f2245503971d36f33c94 Just a small tweak: the code_aster overview was rendered twice because the same URL was  declared both on the code_aster parent and its Overview child.

fsimonis:
I'll take a look at this now.

MakisH:
When updating the branch, the trivial change from https://github.com/precice/precice.github.io/pull/931 will need to be ported.

fsimonis (approved):
Reviewing this PR really makes you realize how slow and buggy the GitHub interface has become.

I opened an issue to cleanup these random snippets of JS that are lying around. https://github.com/precice/precice.github.io/issues/932

There is still a lot of CSS in the repo, but hopefully this will shrink when moving to the SASS version of the Hugo site.

I tested it locally and walked through the site with the old and the new side by side using the iPhone 17 preset. 
The new theme is slightly more spacious, which is not a bad thing.
The only functional difference I noticed is that the sidebar of the new version is way snappier. Not sure why.

Anyway. Looks good and ready to go.

MakisH:
Just in case this is needed in the future: https://github.com/precice/precice.github.io/commit/b25d6e19ce9faecb5738632bcf83480336d4f3ae is the last commit with Bootstrap 3.

Issue #934, opened 2026-06-15
XML reference layout broken on mobile


fsimonis:
### Page URL

https://precice.org/configuration-xml-reference.html#precice-configuration

### What is wrong?

The XML reference layout is broken on mobile. The code blocks and tables should be responsive and allow horizontal scrolling.

PR #936, opened 2026-06-16
Fix XML reference layout on mobile

MuhammadAashirAslam:
Fixes the XML reference layout on mobile (#934)  by making wide formulas horizontally scrollable instead of overflowing the page.
 Screenshot:
 
image


MakisH (approved):
Nice catch, thanks!

Issue #897, opened 2026-05-15
Various issues with the PDF export


MakisH:
Some issues with the current [PDF export](https://precice.org/docs-meta-publish-to-pdf.html) (as of May 2026 and while reviewing #734):

- [x] The [XML config reference](https://precice.org/configuration-xml-reference.html) appears in the middle of the document for several pages. It would be better to render it in an appendix. -> Fixed by #898 
- [x] Some pages (e.g., [special systems](https://precice.org/installation-special-systems.html) are too wide. -> Tried to fix in https://github.com/precice/precice.github.io/pull/899, but code blocks are not the issue.
- [x] Pages are centered and some images take only as much width as the text. For example, the [release strategy page](https://precice.org/dev-docs-release-strategy.html). -> Workaround in https://github.com/precice/precice.github.io/pull/899 
- [ ] Mermaid graphs (e.g., in [multi-coupling configuration](https://precice.org/configuration-coupling-multi.html#composition-of-bi-coupling-schemes)) are not rendering.
- [x] The footer (including the Newsletter block) appears at the bottom of the Quickstart.
- [x] There is often an empty trailing page.
- [x] Some (thin) images are too large and take up too much page area. We should restrict the maximum height. For example, the [heat exchanger tutorial](https://precice.org/tutorials-heat-exchanger.html).
- [ ] Prince [throws several warnings](https://precice.org/docs-meta-publish-to-pdf.html#warnings-by-prince). We should eventually look at those.
- [ ] YouTube videos are occupying space without any rendering. For example, in the [OpenFOAM adapter overview](https://precice.org/adapter-openfoam-overview.html). -> The space issue is now mostly solved in https://github.com/precice/precice.github.io/pull/972, we would just need a placeholder link to the video.

Many of these issues would be easy to fix. Most of the relevant files are in the [`pdfconfigs`](https://github.com/precice/precice.github.io/tree/master/pdfconfigs) folder. PRs are welcome, but please keep them small (one per aspect).

Besides the above, some issues that would require discussion in the team:

- [x] The copyright statement is incomplete, only pointing to the core library -> Expanded in https://github.com/precice/precice.github.io/pull/899
- [ ] The "Community" and "About" pages are not incorporated (do we even want that?). Links to these appear as "page 0". From the "Community", we could add the contributing/guidelines pages, and the "About" would be interesting as well.

[...]

PR #938, opened 2026-06-17
Limit PDF image height

MuhammadAashirAslam:
Limit image height in PDF output to prevent tall, narrow images from taking excessive page space e.g the heat exchanger image

Screenshot of the generated Pdf:

image


MakisH (approved):
I have not checked the complete PDF, but the screenshot looks good and the changes make sense. Thanks!

Related to https://github.com/precice/precice.github.io/issues/897

PR #940, opened 2026-06-22
Add SPARTA DSMC to the list of third-party adapters

caozq-China:
Add sparta-precice-adapter in Third-party adapters


MakisH (approved):
Thank you for adding the adapter to the list! I was not aware of your work so far. Coupling rarefied flows is not something that preCICE is used very frequently for! 😄

Note that we are in the process of establishing some standards for adapters: https://precice.org/community-guidelines-adapters.html

This list will eventually be replaced by a more elaborate page.

Issue #590, opened 2025-09-04
Create sections for imported files


fsimonis:
Some mounted locations such as adapters are not single pages but generally complete sections.

We should create the necessary folder structure with the required `_index.md` files in the existing content so that the sections are created correctly.

Example:

The openfoam adapter docs are here https://github.com/precice/openfoam-adapter/tree/develop/docs
These pages and images are to be mounted here `content/docs/adapters/openfoam`.
This still requires the `content/docs/adapters/openfoam/_index.md` to exist.

These need to be created as part of #588

MuhammadAashirAslam:
So working on this issue, here are a few findings.

The symlink approach will not work: https://github.com/gohugoio/hugo/issues/11556

## Module mounts

Instead of creating a real `_index.md`, I mapped the existing imported file into
Hugo's virtual content filesystem:

```toml
[[module.mounts]]
source = "imported/code_aster-adapter/docs/adapter-code_aster.md"
target = "content/docs/adapters/code_aster/_index.md"
```

This keeps the source repository unchanged:

```text
imported/code_aster-adapter/docs/adapter-code_aster.md
```

but Hugo sees it as:

```text
content/docs/adapters/code_aster/_index.md
```

I verified this with:

```bash
hugo list all | grep code_aster
```

The relevant output was:

```text
imported/code_aster-adapter/docs/adapter-code_aster.md,,The code_aster adapter,...,https://precice.org/docs/adapters/code_aster/,section,docs
```

The important part is:

```text
/docs/adapters/code_aster/,section,docs
```

This confirms that Hugo sees the imported Code_Aster document as a `section`,
not only as a normal page.

Resources can also be mounted the same way, for example images and PDFs:

```toml
[[module.mounts]]
source = "imported/code_aster-adapter/docs/images"
target = "content/docs/adapters/code_aster/images"

[[module.mounts]]
source = "imported/code_aster-adapter/docs/thesis-excerpt-cheung.pdf"
target = "content/docs/adapters/code_aster/thesis-excerpt-cheung.pdf"
```

## Aliases

Another thing is aliases, which are demonstrated in the video attached below.

Jekyll currently publishes Code_Aster at:

```text
/adapter-code_aster.html
```

The natural Hugo section URL is:

```text
/docs/adapters/code_aster/
```

To keep the old URL working, I tested a Hugo alias using config cascade:

```toml
[[cascade]]
aliases = ["/adapter-code_aster.html"]

[cascade._target]
[...]

PR #941, opened 2026-06-23
Rename content files to remove unneeded prefixes

### LINKED ISSUE #589: Rename content files to their desired name
We need to rename files of pages to their desired name by removing their prefix.
This shouldn't matter in Jekyll as we already use permalinks for every page.

Examples:

* `content/docs/configuration/advanced/configuration-action.md` becomes `content/docs/configuration/advanced/-action.md`
* `content/docs/installation/building-from-source/installation-source-configuration.md` becomes `content/docs/installation/building-from-source/configuration.md`

MuhammadAashirAslam:
## Rename content files to their desired name

Removes the redundant path-echoing prefixes from native `content/` filenames now
that the folder hierarchy (from #580) provides the context, e.g.:

- `content/docs/configuration/advanced/configuration-action.md` → `.../action.md`
- `content/docs/installation/building-from-source/installation-source-configuration.md` → `.../configuration.md`

Section overview/landing pages become `overview.md` within their folder
(e.g. `tutorials.md` → `tutorials/overview.md`, as suggested in the issue).

### URLs are unchanged

Every page keeps its existing `permalink:` front matter untouched, so Jekyll
publishes the exact same `*.html` URLs as before. `bundle exec jekyll build`
passes and the generated `_site/` URL set is identical. A few internal links
that referenced old source filenames were updated to point at the page permalinks.

### Exception: two porting guides keep their original names

`content/docs/couple-your-code/porting/couple-your-code-porting-v1-2.md` and
`couple-your-code-porting-v2-3.md` are **intentionally not renamed**. These
migration guides legitimately contain old preCICE API names, and the
`check-old-precice` pre-commit hook skips them by matching `"porting"` in the
**filename**. Renaming them (e.g. to `v1-2.md`) would drop that match and make
the hook fail. Rather than weakening the hook, the two files are left as-is.
(If preferred, the hook's `fileFilter` could match the path/directory instead of
just the filename in a follow-up.)

### Scope

Native `content/` only. Imported adapter/tutorial content (`imported/**`) is
handled separately via module mounts in #590.

Closes #589 (and #538).


content/docs/couple-your-code/advanced/moving-or-changing-meshes.md
MakisH, line 1: Good catch!

content/docs/installation/bindings/julia.md
```diff
@@ -7,7 +7,7 @@ summary: "Use add PreCICE to install the Julia language bindings from the offici
 
 ## The versioning scheme
 
-The versioning scheme of the Julia bindings is the major and minor preCICE version which the bindings are tested with, followed by the version of the Julia bindings. This differs from the [matlab](installation-bindings-python.md) and [python](installation-bindings-python.md) bindings due to the [Julia versioning guidelines](https://pkgdocs.julialang.org/v1/toml-files/#The-version-field), which only allow major, minor, and patch versions.
+The versioning scheme of the Julia bindings is the major and minor preCICE version which the bindings are tested with, followed by the version of the Julia bindings. This differs from the [matlab](installation-bindings-matlab.html) and [python](installation-bindings-python.html) bindings due to the [Julia versioning guidelines](https://pkgdocs.julialang.org/v1/toml-files/#The-version-field), which only allow major, minor, and patch versions.
```
MakisH, line 10: Good catch!

MakisH (approved):
I went through all the pages, the names make sense. Nice URL fixes as well!

I would revert the base branch change and merge it already into `master`. I have already renamed the pages in most/all of the imported repositories anyway.

I see that we still have a few repositories that we could still import from submodules, but we are getting there. I would directly address this in the context of Hugo.

PR #944, opened 2026-07-07
Migrate Jekyll _layouts to Hugo layouts/_default

MuhammadAashirAslam:
This PR replaces the legacy Jekyll _layouts with standard Hugo equivalents (layouts/_default). During the port, I also cleaned up some of dead Javascript code (unused map/sidebar scripts) and fixed a bug where the footer was rendering twice on posts.

This establishes the baseline layout structure needed before migrating the actual content sections and shortcodes.

Also added the .gitignore entries for hugo 

Short description of files:
- `layouts/_default/baseof.html`: Provides the shared web document shell, redirects, navigation, sidebar, footer, analytics, and site scripts.
- `layouts/_default/list.html`: Renders section pages with summaries, table of contents, edit metadata, compatibility-aware content, and child-page links.
- `layouts/_default/none.html`: Provides a minimal layout for pages that only need compatibility-aware content rendering.
- `layouts/_default/page.html`: Renders standard pages with summaries, metadata, tags, breadcrumbs, and translated legacy content.
- `layouts/_default/single.html`: Renders individual content pages with summaries, metadata, tags, breadcrumbs, and translated legacy content.
- `layouts/partials/compat_content.html`: Converts imported Jekyll callouts, inline includes, and `site.*` values into Hugo-renderable content.


layouts/partials/compat_content.html
```diff
@@ -0,0 +1,25 @@
...
+
+  {{- /* Translate the legacy Jekyll callouts into the existing Hugo shortcodes. */ -}}
+  {{- range $tag := $blockTags -}}
+    {{- $openPattern := printf `{%%[[:space:]]*%s[[:space:]]*%%}` $tag -}}
+    {{- $closePattern := printf `{%%[[:space:]]*end%s[[:space:]]*%%}` $tag -}}
+    {{- $content = replaceRE $openPattern (printf `{{< %s >}}` $tag) $content -}}
+    {{- $content = replaceRE $closePattern (printf `{{< /%s >}}` $tag) $content -}}
+  {{- end -}}
+
+  {{- /* The imported documentation uses only note and warning inline includes. */ -}}
```
MakisH, line 19: This is not the case. See:

- `disclaimer`: https://precice.org/adapter-openfoam-overview.html
- `important` (added after this PR was opened): https://precice.org/adapter-calculix-get-adapter.html

Or do I misunderstand the comment?

layouts/partials/compat_content.html
MakisH, line 1: Do I understand correctly that this file could be removed once all of the existing alerts have been switched to the hugo-native system? If yes, a comment at the top of the file would help.

We could (in another PR) just replace all of the legacy alerts. There is no particular reason to keep them.

How does https://github.com/precice/precice.github.io/pull/961 fit with this one?

layouts/_default/baseof.html
```diff
@@ -0,0 +1,81 @@
...
+            {{ block "main" . }}{{ end }}
+          </main>
+        </div>
+      {{ end }}
+    </div>
+  {{- end }}
+
+  {{ partial "footer.html" . }}
+  {{ partial "scroll_to_top.html" . }}
+  {{ partial "google_analytics.html" . }}
```
MakisH: We don't use Google Analytics - We use https://plausible.io/precice.org instead. This file can go (in this or in another PR).

MakisH (commented):
I did a quick sanity check, here are some minor comments to consider; nothing blocking at the moment.

I cannot build the website at this state, it still needs https://github.com/precice/precice.github.io/pull/945 (and we should prioritize a CI workflow to test this on every PR).

layouts/partials/compat_content.html
MuhammadAashirAslam: Yes , you are partially right , disclaimer falls in the warning includes but important doesn't , lemme correct that (for now )

layouts/partials/compat_content.html
MuhammadAashirAslam: Yep, this file is to be removed the content in the website repo now uses the shortcodes #964 introduced in #961 , but the imported repos still use the liquid jekyll syntax and we have to update them so this brittle/hack can be removed

layouts/_default/baseof.html
MuhammadAashirAslam: Removed

layouts/partials/compat_content.html
MuhammadAashirAslam: Small correction to my previous reply: disclaimer and important are both handled as block callouts (https://github.com/precice/precice.github.io/pull/944/changes/BASE..0d98c7542247a9c9156d737083acc51c2beea5ad#diff-dde938805e82cece2b29c38a34aedc7b9bbfda723832054ef83a9f21c4647a9cR5) by the compatibility layer. Only the legacy inline-include conversion is limited to note and warning; I’m clarifying the comment accordingly.

layouts/partials/compat_content.html
MuhammadAashirAslam: Added the  TODO comment https://github.com/precice/precice.github.io/pull/944/changes/38a5bb1a914d62c3f46daa84d7d6dff751cba431

MakisH (approved):

PR #948, opened 2026-07-15
Add publication and tagging partials

MuhammadAashirAslam:
This PR migrates the publication and tagging components. It sets up the publication panel templates, archive structure, and the tag logic needed to sort list outputs.
Short description of the files:
- `layouts/partials/archive.html`: Renders the standard archive page heading and archive content container.
- `layouts/partials/fenics_publications.html`: Renders the FEniCS citation and related-literature sections from Hugo publication pages.
- `layouts/partials/publication_panel.html`: Renders publication metadata, DOI and publisher links, and BibTeX downloads without Jekyll collection reads.
- `layouts/partials/publications_by_tag.html`: Filters and sorts Hugo publication pages by tag before rendering their panels.
- `layouts/partials/taglogic.html`: Lists pages matching a tag with links, content type, and a short summary.


layouts/partials/fenics_publications.html
MakisH, line 1: Same question as in https://github.com/precice/precice.github.io/pull/961#discussion_r3702215704: Why a FEniCS-specific partial? We should have one variant that works for all (and, if needed, make the FEniCS-specific details follow the common, card-like variant).

(I think these two PRs could have been one)

MakisH (commented):
A quick sanity check looks fine, but could you point to some rendering examples?

layouts/partials/fenics_publications.html
MuhammadAashirAslam: This is not related to FEniCS , the problem is the rendering of the jekyll liquid syntax from imported repos , this partial is also related to the same compat_content.html hack as FEniCS is the only imported adapter to have a publication loop

MakisH: Let's add a `TODO` comment to such files, to explain when they should be removed.

MuhammadAashirAslam:
Few examples:
https://github.com/precice/precice.github.io/pull/964/changes#diff-9ab63cfb8b664c4731440aa0a1feefd4fb0e7625a381a6a3912cc8167c51055fR17
https://github.com/precice/precice.github.io/pull/964/changes#diff-0f9720651d2a1b3fec584d07b52488ff89b2092723854628c0c8dc80f96e659cR35

Where as the [taglogic.html](https://github.com/precice/precice.github.io/pull/948/changes#diff-cb72b9a3c31f79ba17291090948c8d3c5848ee93e85d001358105b14067e0669) and [archive.html](https://github.com/precice/precice.github.io/pull/948/changes#diff-e05331b2b568ea01a5b8798fa1a90071ed0cda20278281d93eb61089cdb45218)  are the direct equivalent of the https://github.com/precice/precice.github.io/blob/master/_includes/archive.html and https://github.com/precice/precice.github.io/blob/master/_includes/taglogic.html . Although these two are not used in both versions of the website

layouts/partials/fenics_publications.html
MakisH: Still needed, but I think clear at this point.

layouts/partials/archive.html
MakisH, line 1: Is this even used anywhere? If not, it could already be deleted.

layouts/partials/fenics_publications.html
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/948/changes/7ad15aa01fa4ab26ef9096138fbe7f75a21d7d94

layouts/partials/archive.html
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/948/changes/7ad15aa01fa4ab26ef9096138fbe7f75a21d7d94

MakisH (approved):

PR #952, opened 2026-07-15
Add third party integrations and analytics partials

MuhammadAashirAslam:
This PR migrates third-party tools, tracking services, and search integrations. It sets up files for Algolia search, Commento discussions, Google Analytics, and the user feedback widge


layouts/partials/commento.html
MakisH, line 1: Since this is not used on the current website, let's just drop it.

layouts/partials/feedback.html
MakisH, line 1: Since this is not used on the current website, let's just drop it. We have the "edit" me button, instead, and many ways to reach us.

We also don't set a feedback email.

MuhammadAashirAslam: Done

layouts/partials/commento.html
MuhammadAashirAslam: Done

MakisH (approved):
Quick sanity check looks good, and the search seems to work on the test deployment.

MakisH (approved):

PR #951, opened 2026-07-15
Add sidebar layouts

MuhammadAashirAslam:
This PR migrates the sidebar navigation files to Hugo partial layouts. It handles the structure, check for active items, and recursive rendering of the side menu to match the new site structure.
Short description of the files:
- `layouts/partials/sidebar.html`: Selects the page sidebar data and starts rendering the navigation tree.
- `layouts/partials/sidebar_is_web.html`: Checks whether a sidebar item is enabled for web output.
- `layouts/partials/sidebar_item_active.html`: Recursively determines whether a sidebar branch contains the current page.
- `layouts/partials/sidebar_page_matches_url.html`: Matches a page against its permalink, URL, and aliases.
- `layouts/partials/sidebar_recursive.html`: Recursively renders sidebar leaves and Bootstrap collapse branches.
- `layouts/partials/sidebar_resolve_url.html`: Preserves external URLs and resolves internal URLs with `relURL`.
- `layouts/partials/sidebar_url_clean.html`: Normalizes URLs before sidebar comparisons.

PR #953, opened 2026-07-15
Add content utilities and helpers partials

MuhammadAashirAslam:
This PR adds general utility templates. It includes helper layouts for rendering images, custom metadata editors, funding icons, news banners, newsletters, and formatting external links.


layouts/partials/edit_button.html
MakisH, line 1: Why is the `_metadata`? This is the button itself + the "last modified date", right?

What is a `custom metadata editor`? (from the PR description)

A better name would be `edit_button.html`.

layouts/partials/inline_image.html
MakisH, line 1: Why is this needed? Isn't the markdown syntax enough?

layouts/partials/links.html
MakisH, line 1: The `links.html` and `links_entry.html` correspond to the sidebars, right?

https://github.com/precice/precice.github.io/pull/951

Just for me to understand, as this was not mentioned in the PR description.

layouts/partials/news_banner.html
```diff
@@ -0,0 +1,18 @@
+{{/*
+News banner template, disabled by default. To publish a banner, remove this
+outer Hugo comment wrapper and edit only the announcement text and link below.
+The base layout already calls this partial, and the existing classes should be
+kept for consistent styling. To hide the banner from the front page, wrap the
+uncommented block with `{{ if not .onlanding }}` and `{{ end }}`; without that
+condition, it appears on every page.
```
MakisH: I would prefer an always-on check here, as we currently have it: For example, `if false` should disable the banner. Keeping the comment (but ending at line 7) would be useful.

If I remove the comment and then want to disable the banner again, I would need to go to the history to find the comment again.

layouts/partials/url.html
MakisH, line 1: Does this change the style of external URLs to point out to the user that they are external?

layouts/partials/edit_button.html
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/953/changes/29064f0c69e357ee69bd405f2405b35d429f55f1

layouts/partials/news_banner.html
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/953/changes/73a70af86e79547ed4b2b6d75564caa5a056d733

layouts/partials/news_banner.html
```diff
@@ -0,0 +1,20 @@
+{{/*
+News banner template, disabled by default. To publish a banner, change `false`
+below to `true` and edit only the announcement text and link below. The base
+layout already calls this partial, and the existing classes should be kept for
+consistent styling. To hide the banner from the front page, wrap the banner
+block with `{{ if not .onlanding }}` and `{{ end }}`; without that condition,
+it appears on every page.
```
MakisH: The current modes are:

- show nowhere
- show only on landing page
- show on every page

The comment suggests:

- show nowhere
- show everywhere but the langing page
- show everywhere


```suggestion
consistent styling. To only show the baner on the front page, change the banner
block condition to `{{ if .onlanding }}`.
```

PR #961, opened 2026-08-03
Add Hugo shortcodes

MuhammadAashirAslam:
Add the hugo shortcodes directory.
Shortcodes are reusable Hugo components that let authors add structured elements to Markdown without writing HTML. For example, this PR provides shortcodes for notes, warnings, callouts, publication panels, Mermaid diagrams, testimonials, and site parameters. They replace the old Jekyll/Liquid includes and keep these elements consistent across the website.

This also requires documentation as currently the [layouts/partials/compat_content.html](https://github.com/MuhammadAashirAslam/precice.github.io/blob/hugo-test/layouts/partials/compat_content.html) is changing the liquid syntax to hugo one but in the long run it will be better that all the content (in the imported repos too) to use hugo native shortcodes/content , so that there is no need of compat_content.html (as it is brittle)


layouts/shortcodes/fenics_publications.html
MakisH, line 1: Why a file specifically for FEniCS?

MakisH (commented):
Some questions:

- Are these already demonstrated somewhere? 
- What happens to the current alerts?

layouts/shortcodes/fenics_publications.html
MakisH: To be removed once the FEniCS adapter uses the same style as the rest.

MakisH (approved):

PR #963, opened 2026-08-05
Add specialized Hugo layouts

MuhammadAashirAslam:
This PR adds the remaining Hugo layout templates for the homepage, publications, testimonials, and Markdown
  image rendering.

  - Adds the Hugo homepage template.
  - Adds publication list and detail templates.
  - Adds testimonial list and detail templates.
  - Adds the image render hook for resolving page and section resources mounted by Hugo Modules.


MuhammadAashirAslam:
As discussed, specialized layouts can be used to give testimonial and publication pages their own styling if we want to improve them in the future.

layouts/_markup/render-image.html
MakisH, line 1: Same question as in https://github.com/precice/precice.github.io/pull/953#discussion_r3728272212: Why is this needed?

layouts/index.html
MakisH, line 1: In this PR, for example, it would be useful to remove the old `index.html` in the same PR. I could directly compare the files, and we would reduce the chances of getting the content outdated.

MakisH: Th diff looks fine. But still, `git mv` would be great.

MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/963/changes/19d146c090d05bbaf5dac30284ec574c1be2750b

layouts/_markup/render-image.html
MuhammadAashirAslam: This is a Hugo image render hook. It still uses normal Markdown image syntax, but resolvesimages mounted next to imported documentation to their Hugo page-resource URL. Without it, links such as images/example.png remain relative and break on nested tutorial or adapter pages; external image URLs are left unchanged.
Doumentation link : https://gohugo.io/render-hooks/images/

MakisH (approved):

PR #966, opened 2026-08-08
Replace Jekyll XML Outputs and Update Doxygen Sitemap Integration

MuhammadAashirAslam:
Replaced Jekyll feed, sitemap-index, and robots templates with Hugo-native output generation while preserving their public URLs and Doxygen sitemap reference. Updated doxygen/patch.sh for Hugo configuration paths and configurable Doxygen base URLs.


MakisH (approved):
I have not checked what the result is, but sanity check looks good. Merging to proceed with the rest.

PR #968, opened 2026-08-10
Add Hugo Algolia indexing

MuhammadAashirAslam:
This PR replaces the previous Jekyll Algolia plugin with a Hugo JSON export and a standalone Node.js
  indexing tool.

  ## layouts/index.algolia.json

  This is a Hugo output template, not the Algolia index itself. During the Hugo build, it collects searchable
  pages and writes them to public/algolia.json.

  For each page, it stores:

  - The page title and URL
  - Rendered HTML and plain text
  - Headings and section information
  - Tags, categories, keywords, and modification date
  - The configured nodes to index, such as paragraphs, code blocks, and tables

  Pages marked with search: exclude are skipped. The template also uses the existing content compatibility
  layer so imported documentation is indexed after its legacy markup has been converted to Hugo-compatible
  output.

  ## tools/algolia-index.mjs

  This is the standalone command-line indexer. It reads the JSON generated by Hugo and converts it into
  records that Algolia can search.

  The script:

  - Parses the generated HTML with Cheerio
  - Extracts headings, anchors, paragraphs, code blocks, and tables
  - Removes scripts, styles, iframes, and alert content from searchable text
  - Keeps tutorial introductions ahead of informational callouts
  - Generates deterministic object IDs so repeated runs update the same records
  - Adds heading hierarchy, snippets, highlighting fields, and ranking metadata
  - Splits oversized records when they exceed the configured limit
  - Validates the export before any upload takes place
  - Applies the Algolia searchable attributes, facets, snippets, and ranking settings
  - Replaces the target index atomically using Algolia’s replaceAllObjects operation

  The command supports a dry-run mode:

  npm run algolia:index -- --dry-run

  A dry run performs all parsing and validation locally without contacting Algolia.

  For an actual upload, the script reads the write credential from the environment. The application ID and
  index name are supplied by the workflow, while the write key remains a repository secret.

  ## test/algolia-index.test.mjs

  These tests use Node.js’s built-in test runner, so no additional test framework is required.

  The tests cover the important indexing behavior:

  - Valid Hugo exports produce the expected Algolia records
  - Heading and anchor metadata are preserved
  - Object IDs are deterministic
  - Alert boxes are excluded so tutorial results begin with their introduction
  - Oversized records are split without exceeding the configured limit
  - Invalid JSON schemas and invalid command-line options are rejected

  They run with:

  npm run test:algolia

  ## package.json

  This file defines the Node.js tooling used by the Algolia implementation.

  It provides:

  - npm run algolia:index for transforming and uploading records
  - npm run test:algolia for running the test suite
  - The algoliasearch dependency for communicating with Algolia
  - The cheerio dependency for parsing rendered HTML
  - The minimum supported Node.js version

  The workflow installs these dependencies with npm ci, which uses the lockfile rather than resolving
  versions again.

  ## package-lock.json

  package-lock.json is generated automatically by npm from package.json. It records the exact versions,
  download locations, and integrity checksums of the direct and transitive Node.js dependencies.

  It is committed so that local development and GitHub Actions install the same dependency tree every time.
  It should not be edited manually; it should be regenerated with npm when package.json changes.

  ## Workflow

  The workflow performs the indexing in several controlled stages:

  1. Checks out the repository with full history.
  2. Installs the configured Go, Hugo, and Node.js versions.
  3. Runs npm ci using package-lock.json.
  4. Runs the Algolia unit tests.
  5. Verifies Hugo Modules.
  6. Builds the Hugo site and generates public/algolia.json.
  7. Runs the indexer in dry-run mode to validate the generated records.
  8. Uploads the validated records to Algolia.

  The application ID and index name are configured to match the Hugo configuration. The workflow reads only
  ALGOLIA_API_KEY from GitHub repository secrets for the write credential.


.github/workflows/update-algolia.yml
```diff
@@ -1,39 +1,74 @@
...
+concurrency:
+  group: update-algolia
+  cancel-in-progress: false
+
+env:
+  GO_VERSION: "1.26.3"
+  HUGO_VERSION: "0.163.3"
+  NODE_VERSION: "24.18.0"
+  ALGOLIA_APP_ID: "LIT6P0EW26"
+  ALGOLIA_INDEX_NAME: "jekyll"
```
MakisH: Check: Provide a different index name.

MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/968/changes/0267c3f7cc8eaf8d7aa9037058f7bc635dd974d6

tools/package-lock.json
MakisH, line 1: Do we need the `package-lock.json` on Git if we have the `package.json`?

tools/package.json
MakisH, line 1: If this `package.json` is only for the tools, shouldn't it be in the `tools/` directory?

tools/algolia-index.mjs
```diff
@@ -0,0 +1,510 @@
...
+const CALLOUT_SELECTOR = ".alert,[role='alert']";
+
+// Preserve the production Algolia ranking and highlighting contract.
+export const INDEX_SETTINGS = Object.freeze({
+  searchableAttributes: [
+    "title",
+    "headings",
+    "unordered(content)",
+    "collection,categories,tags",
+  ],
```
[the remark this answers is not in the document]
MuhammadAashirAslam: Algolia explicitly supports comma-separated attributes in one searchableAttributes entry to give them equal priority, so this spilliting is not required as they will give them precedence which is not needed here , https://www.algolia.com/doc/api-reference/api-parameters/searchableAttributes

tools/algolia-index.mjs
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    const tagName = node.tagName?.toLowerCase();
+
+    if (tagName && /^h[1-6]$/.test(tagName)) {
+      headingLevel = Number(tagName[1]) - 1;
+      hierarchy[headingLevel] = normaliseWhitespace(element.text());
+      hierarchy.fill(null, headingLevel + 1);
+      anchor = findAnchor($, element) || anchor;
+    }
```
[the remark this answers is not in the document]
MuhammadAashirAslam: Done

package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "precice-website-tools",
+  "private": true,
+  "type": "module",
+  "engines": {
+    "node": ">=20"
+  },
```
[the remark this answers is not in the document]
MuhammadAashirAslam: Done

tools/package-lock.json
MuhammadAashirAslam: No its better that we have package-lock.json tracked , this will pin the exact versions for reproducible CI runs , also npm ci needs it

tools/package.json
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/968/changes/ebb11ec4a0a0f4977f516e6e2443c780eebad3ba

tools/algolia-index.test.mjs
MakisH, line 1: The `test/` directory does not exist at the moment, and it looks like this will be the only test added in the near future. I understand that it makes sense since the script is in-house, but I would keep it in the same folder as the main script (`tools/`).

tools/algolia-index.test.mjs
```diff
@@ -0,0 +1,90 @@
...
+        <script>ignored content</script>
+      `,
+      plain: "Use precice-config to configure the adapter. Version 3.4",
+      section: "docs",
+      kind: "page",
+      tags: ["adapter"],
+      categories: [],
+      keywords: "configuration",
+      date: 1_784_198_400,
+    },
```
MakisH, line 31: I would keep this example more generic: I would remove mentions to preCICE and concrete versions. This could cause some confusion while searching through the codebase.

Similarly below.

tools/package-lock.json
```diff
@@ -0,0 +1,534 @@
...
+      }
+    },
+    "node_modules/encoding-sniffer": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/encoding-sniffer/-/encoding-sniffer-0.2.1.tgz",
+      "integrity": "sha512-5gvq20T6vfpekVtqrYQsSCFZ1wEg5+wW0/QaZMWkFr6BqD3NfKs0rLCx4rrVlSWJeZb5NBJgVLswK/w2MWU+Gw==",
+      "license": "MIT",
+      "dependencies": {
+        "iconv-lite": "^0.6.3",
+        "whatwg-encoding": "^3.1.1"
```
MakisH, line 374: I am getting a deprecation warning here:

```
tools [hugo-algolia]$ npm install
npm warn deprecated whatwg-encoding@3.1.1: Use @exodus/bytes instead for a more spec-conformant and faster implementation
```

tools/package-lock.json
```diff
@@ -0,0 +1,534 @@
...
+      "license": "MIT"
+    },
+    "node_modules/undici": {
+      "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
+      "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.18.1"
+      }
```
MakisH, line 509: I am getting a vulnerability report here:

```
tools [hugo-algolia]$ npm audit
# npm audit report

undici  7.0.0 - 7.28.0
Severity: high
undici vulnerable to downstream response desynchronization via retry interceptor - https://github.com/advisories/GHSA-8xcm-r25x-g524
undici vulnerable to cross-user information disclosure and parse-time crash via degenerate private cache directives - https://github.com/advisories/GHSA-4cwx-7wf7-3272
undici vulnerable to CRLF Injection via blob-like body 'type' property - https://github.com/advisories/GHSA-m8rv-5g2x-5cg5
undici vulnerable to cross-user information disclosure via whitespace around equals in Cache-Control directives - https://github.com/advisories/GHSA-jr45-8vmc-qm54
undici vulnerable to cookie attribute injection via unsanitized domain and unparsed setCookie fields - https://github.com/advisories/GHSA-v3r7-h72x-cjcm
fix available via `npm audit fix`
node_modules/undici

1 high severity vulnerability

To address all issues, run:
  npm audit fix
```

tools/algolia-index.mjs
MakisH, line 1: I tried the dry-run from the PR description, but it failed. What am I missing?

```
tools $ npm run algolia:index -- --dry-run

> algolia:index
> node algolia-index.mjs --dry-run

Algolia indexing failed: Unable to read Algolia export at /home/gc/repos/precice/website/tools/public/algolia.json: ENOENT: no such file or directory, open '/home/gc/repos/precice/website/tools/public/algolia.json'
```

MakisH (commented):
I do not have much contact with Javascript to review the `.mjs` files, so I can only trust what the Copilot review says right now, and that you know what you are doing / the testing you have performed.

On https://muhammadaashiraslam.github.io/precice.github.io/, if I add, `precice.github.io` to e.g., `https://muhammadaashiraslam.github.io/docs/installation/overview/#installing-the-core-library` to the resulting URLs (in this case, to get `https://muhammadaashiraslam.github.io/precice.github.io/docs/installation/overview/`), the search works.

Just a couple of minor comments / questions.

tools/algolia-index.test.mjs
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/968/changes/dc491900e24e45862d459554bfda883724e21fa4

tools/algolia-index.test.mjs
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/968/changes/dc491900e24e45862d459554bfda883724e21fa4

tools/package-lock.json
MuhammadAashirAslam: Yes but this warning comes from the package itself , as it hasnot upgraded to the new dependecy . This is fine in our case

tools/package-lock.json
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/968/changes/f4f4ef0a9988c58ad8a5c8eb39a8a3c957092dcb

tools/algolia-index.mjs
MuhammadAashirAslam: Yeah it needs to have the public/algolia.json , which i assume is not working right now due to some legacy code in content/docs/docs-meta/publish-to-pdf.md but if you want simply replace lines 140–141 in content/docs/docs-meta/publish-to-pdf.md  (this is solved by #972 )
```text
https://precice.org/titlepage.html
https://precice.org/tocpage.html
```
Then run hugo or hugo --gc this will create the public/algolia.json

MakisH: Before that, the issue here is that the script is under `tools/`, and it is looking for `tools/public/algolia.json`. In the meantime, I am building the website in the parent directory.

How should these two be combined? I assume I need some kind of flag to point the `algolia:index` to the `../public`.

`.github/workflows/update-algolia.yml` seems to be doing the same as I am doing (unless I overlook something).

MuhammadAashirAslam: Thanks , very nice catch 😅
Updated to go the root https://github.com/precice/precice.github.io/pull/968/changes/ceef0aa64642eef103bec2e842f86050e0013e5d

MakisH: Worked now, thanks!

MakisH:
Keeping this open until #972 is merged (missing one comment), maybe the CI works then.

PR #969, opened 2026-08-10
Migrating Hugo modules and Discourse data updates worflows

MuhammadAashirAslam:
Replaces the obsolete submodule updater with automated Hugo Module synchronization using go.mod and go.sum. Updates the Discourse news and FAQ data workflows for Hugo’s static/assets paths, and removes the retired submodule workflow.


.github/workflows/update-discourse-data.yml
```diff
@@ -33,19 +32,15 @@ jobs:
...
           python tools/fetch-faq.py
 
-      - name: Configure Git author
+      - name: Commit and push changes
         run: |
+          git add static/assets/data/news.json static/assets/data/faq.json
+          if git diff --cached --quiet; then
+            echo "Discourse data is already up to date."
+            exit 0
+          fi
```
MakisH, line 41: In the current approach, the workflow is a bit more verbose than needed, because of the following fields:

- `"generated_at": "2026-08-12T07:18:28.922820",`
- `"views": 2451,`

Could we exclude these?

.github/workflows/update-submodules.yml
MakisH, line 1: Note that other repositories trigger the `update-submodules.yml`. Maybe keep the same name?

https://github.com/precice/tutorials/blob/develop/.github/workflows/update-website.yml

MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/969/changes/e1c2afcb6ced6f0df02476c268635e6c0c0aebb0

.github/workflows/update-discourse-data.yml
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/969/commits/3bbea9feff9652c8cac76a2147ec5a4a02341adf
 
I've set a view threshold, so a commit is now only made if the diff is greater than the threshold.
 
Some testing:
- A commit was made ([47bd3af](https://github.com/MuhammadAashirAslam/precice.github.io/commit/47bd3af7883ed0b8fc8f2f9a379b21f97d037e39)) because of the testing reduction ([cb17c4c](https://github.com/MuhammadAashirAslam/precice.github.io/commit/cb17c4c1a751b75e366c467a9c89560d968ccc97)).
- After that, no commit was made even when we ran the workflow again: https://github.com/MuhammadAashirAslam/precice.github.io/actions/runs/32112340065
When a commit is made, all of the view counts are updated, meaning that if only one card reaches the threshold, all of them get updated (the threshold right now is 50), but I think it would be better to reduce it.
 
Also used USER_AGENT because of rate limiting.

.github/workflows/update-discourse-data.yml
```diff
@@ -5,26 +5,25 @@ on:
...
-        with:
-          client-id: ${{ vars.WEBSITE_UPDATER_CLIENT_ID }}
-          private-key: ${{ secrets.WEBSITE_UPDATER_KEY }}
-
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
-          token: ${{ steps.app-token.outputs.token }}
+          fetch-depth: 0
```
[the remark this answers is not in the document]
MuhammadAashirAslam: This is intentional choice

.github/workflows/update-discourse-data.yml
```diff
@@ -33,19 +32,15 @@ jobs:
...
-          git add assets/data/news.json
-          git add assets/data/faq.json
-          git commit -m "Update Discourse data data [skip ci]" || echo "No changes to commit"
-
-      - name: Push commit
-        uses: ad-m/github-push-action@master
-        with:
-          github_token: ${{ steps.app-token.outputs.token }}
-          branch: ${{ github.ref }}
+          git commit --message "Update Discourse data [skip ci]"
```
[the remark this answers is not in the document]
MuhammadAashirAslam: This is intentional choice

MakisH (approved):

PR #972, opened 2026-08-10
Port PDF generation to Hugo

MuhammadAashirAslam:
Ports the documentation PDF workflow from Jekyll to Hugo and Prince. Adds Hugo PDF configuration, templates, sidebar-driven contents, local print assets, and an updated pdf-docs.sh generator while removing obsolete Jekyll PDF files.

Generated PDF: https://drive.google.com/file/d/1zuifXDzyl4Xiy0xE8GPifccP-K9mOIsT/view?usp=sharing


MakisH:
Some comments on the attached PDF:

- [ ] In the ToC:
   - [ ] Ideally, keep the section and page number on the same line, if it fits.
   - [ ] The page number has a serif font, the section a sans-serif font (use the sans-serif for both). 
- [x] Internal links should point to the PDF. They currently point to 127.0.0.1.
- [x] In dependencies overview (page 34), there is a `no-icon`.
- [x] In the Roadmap (page 20), the button "click here" should be replaced by the text (unreleased changes).
- [x] Would there be a command to hard-code to start a new page?
- [x] In the Distribution (page 86), the DOI/DaRUS button is centered, the "external link" icon is on another line. Also in page 129.
- [x] Page 99: table headings with code snippets have a wrong color.
- [x] Page 168: no distinction between inline math and math blocks.
- [x] Page 173: table width.
- [x] Page 194: bloken tables.
- [x] Page 273: the image could be smaller.
- [x] Page 297: the tabs have some issues (numbers not shown, active tab name not readable). There are more issues on the page.

MuhammadAashirAslam:
Alot of issues have been addressed just the TOC one remains and the table width one(which was very trivial , and the information was already complete)
The new pdf can be accessed with the same link in the description

MakisH:
Nice job! The ToC is also fine as-is, I think. No reason to keep investing time on it.

In this one:

>  In the Roadmap (page 20), the button "click here" should be replaced by the text (unreleased changes).

I meant actually loading the unreleased changes. The `(unreleased changes)` placeholder is good enough, in case loading that page element when rendering the PDF is complicated.

How does `layouts/shortcodes/pagebreak.html` work now? Does it automatically break at better places? Do we need to manually call it in pages? Are both already possible? Similarly to the other shortcodes: are they picked up automatically by Prince?

MuhammadAashirAslam:
> How does `layouts/shortcodes/pagebreak.html` work now? Does it automatically break at better places? Do we need to manually call it in pages? Are both already possible? Similarly to the other shortcodes: are they picked up automatically by Prince?

Yep , for manual breaks you can just put the page break shortcode it is completely invisible on the web, but tells Prince to force a new page in the PDF. Most of the page breaks are automatically handled by printstyles.css like preventing orphan headings, awkward image splits, and section breaks . 
Yes they are automatically picked just like the other shortcodes

layouts/partials/pdf/toc.html
```diff
@@ -0,0 +1,30 @@
...
+          (dict "title" "Tutorials" "sidebar" hugo.Data.sidebars.tutorials_sidebar)
+          (dict "title" "Appendix" "sidebar" hugo.Data.sidebars.pdf_appendix_sidebar) }}
+          {{- if eq $index 0 }}
+          <h2 style="margin-top: 20px; padding-bottom: 5px; border-bottom: 1px solid #ccc;">{{ $section.title }}</h2>
+          {{- else }}
+          <h2 style="margin-top: 30px; padding-bottom: 5px; border-bottom: 1px solid #ccc;">{{ $section.title }}</h2>
+          {{- end }}
+            {{- range $section.sidebar.entries -}}
+              {{- with .folders }}{{ partial "pdf/toc-nodes.html" (dict "site" site "nodes" . "level" 0) }}{{ end -}}
+            {{- end -}}
```
MakisH, line 23: Running the generation script (simply `./pdf-docs.sh`), I get an error here:

```
Start building sites … 
hugo v0.165.0-76a5e1880ab46688155b02e99bab9be2a6134492+extended linux/amd64 BuildDate=2026-08-12T14:26:28Z VendorInfo=snap:0.165.0

Total in 2686 ms
ERROR error building site: render: [en v1.0.0 guest] failed to render pages: render of "/home/gc/repos/precice/website/content/_index.md" failed: "/home/gc/repos/precice/website/layouts/index.pdftoc.html:1:3": execute of template failed: template: index.pdftoc.html:1:3: executing "index.pdftoc.html" at <partial "pdf/toc.html" .>: error calling partial: "/home/gc/repos/precice/website/layouts/partials/pdf/toc.html:21:30": execute of template failed: template: _partials/pdf/toc.html:21:30: executing "_partials/pdf/toc.html" at <$section.sidebar.entries>: nil pointer evaluating interface {}.entries
```

tools/pdf-docs.sh
MakisH, line 1: In #968, a `tools/` directory is introduced. Shouldn't this script also be under `tools/`?

MakisH (commented):
The provided PDF and the changes look good, I just have an issue running the generation script (and I am going through the updated documentation).

layouts/partials/pdf/toc.html
MuhammadAashirAslam: I think this is caused by the changes by the #977 PR . Can you try to fetch the latest hugo branch and running it again?

MakisH: Worked now, thanks!

tools/pdf-docs.sh
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/972/changes/33f2fb037af86c3ad6b48f4ccf1d12ee039ebdad

PR #973, opened 2026-08-10
Remove remaining Jekyll-related files and port recent content changes

MuhammadAashirAslam:
These are the last remaining artifacts of jekyll .


MakisH (commented):
So far, all the file removals are as expected. Keeping this open until all other migration PRs have been merged.

edit Aug 24: Needs more updates.

MakisH:
@MuhammadAashirAslam I think we should update the `hugo` branch as well, to remove the unrelated changes from the diff.

MuhammadAashirAslam:
Could you explain a bit more ?
> to remove the unrelated changes from the diff.

for the files touched in https://github.com/precice/precice.github.io/pull/973/changes/308cdc68a7f403c36b26162901b1894cfad6d1b4? I think we can't merge the main branch in to hugo directly because the location of the files have been changed in some cases so it will cause merge conflicts, so i resorted them to port the new changes in this PR

MakisH:
Good point about the locations. This is indeed easier to avoid the merge conflicts.

MakisH (approved):

MakisH (commented):
I am also removing:

- `tooltips.json`
- `.vscode/`

Regarding licenses, I am removing:

- `licenses/LICENSE-BSD-NAVGOCO.txt` (Navgoco)
- `licenses/LICENSE` (documentation-theme-jekyll)

I will instead add the [Hugo license file](https://github.com/gohugoio/hugo/blob/master/LICENSE). I will check if this directly applies to the modifications, or if we need something else.

@MuhammadAashirAslam Should we also move `assets/xmlreference.md` to `static/assets/xmlreference.md`, or is there a reason for it to be there?

MuhammadAashirAslam:
> @MuhammadAashirAslam Should we also move `assets/xmlreference.md` to `static/assets/xmlreference.md`, or is there a reason for it to be there?

We can do this , but it is not recommended , everything in static is served as raw XML file (i.e it will be served as https://precice.org/assets/xmlreference.md) which is also a duplication of the current page

MakisH (approved):
Looks good for now. Regarding licenses, we are now in the same state as before.

Let's continue with testing in a PR from `hugo` to `master`.

PR #979, opened 2026-08-28
Jekyll to Hugo migration

MuhammadAashirAslam:
This PR completes the migration of the preCICE website from Jekyll to Hugo.

#### 1. Directory Structure & Asset Migration
- **Static Assets Relocation**: Moved public static assets from legacy root folders (`images/`, `css/`, `js/`) into Hugo's standard `static/` directory (`static/images/`, `static/css/`, `static/js/`, `static/assets/data/`).
- **Content Hierarchy**: Reorganized Markdown files from into Hugo's section-aware `content/` tree (`content/docs/`, `content/community/`, `content/tutorials/`, `content/about.md`).

#### 2. Architecture & Dependency Management
- **Hugo Modules**: Replaced Git submodules and custom Ruby import scripts with native Hugo Modules (`config/_default/module.toml`, `go.mod`, `go.sum`). Imported adapter and tutorial repositories are now fetched and pinned as Hugo modules.
- **Configuration**: Structured site settings into modular TOML files (`config/_default/hugo.toml`, `module.toml`, `params.toml`).

#### 3. Layouts, Partials & Shortcodes
- Reimplemented all Jekyll Liquid templates and includes in Go HTML templates (`layouts/` and `layouts/partials/`).
- Ported custom Jekyll tags to native Hugo shortcodes (`{{< warning >}}`, `{{< note >}}`, `{{< tip >}}`, `{{< pagebreak >}}`, etc.).
- Preserved all existing permalinks and historical URLs using Hugo `aliases` to ensure zero broken bookmarks or external links.

#### 4. Search Integration (Algolia)
- Migrated the search pipeline to a dedicated Node.js indexing tool (`tools/algolia-index.mjs`).
- Hugo generates `public/algolia.json` during the production build, which the CLI validates (schema check, record size limits) and uploads using atomic index replacement.

#### 5. Offline PDF Generation
- Ported the Prince XML documentation build to Hugo (`tools/pdf-docs.sh`).
- Configured dedicated PDF layout templates (`config/pdf/`, `layouts/partials/pdf/`) to build the consolidated offline PDF directly from the Hugo content tree.

#### 6. GitHub Actions & Automation Workflows
- **`build.yml`**: Streamlined CI to test both standard website generation and Algolia dry-run indexing.
- **`update-submodules.yml`**: Modernized to update Hugo module revisions (`tools/sync_hugo_modules.py`), supporting both daily schedules and `repository_dispatch` triggers from external repositories.
[...]

PR #981, opened 2026-09-08
Add OpenRadioss to the third-party adapters

MakisH:
This PR adds the OpenRadioss adapter to https://precice.org/adapters-overview.html, linking to the OpenRadioss repository, @elisa-santoro's thesis, and the accompanying case files.

Integrating one of the examples in the preCICE tutorials will follow. We are currently evaluating with @carme-hp which examples will be best fit for one.

This PR also fixes the rendering of the tables (already fixed in the `hugo` branch, cc @MuhammadAashirAslam for #979) and the MercuryDPM entry.


carme-hp (approved):
Looks good!

PR #986, opened 2026-09-14
Add license details

### LINKED ISSUE #247: Missing license
While the content is licensed under CC BY, the repository itself (which also includes code) seems to have no license file.
I am not sure if CC BY would be a good idea also here, or if maybe MIT would be better.

MakisH:
Closes #247 by implementing the suggestions stated in https://github.com/precice/precice.github.io/issues/247#issuecomment-5526755511

- Adds more details in the "Licenses" section of `README.md`, clarifying the license of the website and dependencies.
   - These include Jekyll, the theme, and the libraries fetched in `head.html`.
- Adds SPDX headers to all `_plugins`, attributing to `The preCICE developers - precice.org`, using the years I could retrieve from the history.
   - Similarly to the files in `js/` and `tools/` that were written specifically for the preCICE website and might be reusable.
- Removes the navgoco license, since that has been removed in https://github.com/precice/precice.github.io/pull/922.

I probably have missed a few files or dependencies, but this should already be much clearer than before.

FYI @fsimonis @MuhammadAashirAslam @uekerman 

This is also relevant and will need to be adapted for Hugo (#979).


fsimonis (approved):
My approval for the record

MuhammadAashirAslam (approved):
For Hugo , jQuery will be removed (as it was removed as a dependency)
(Also we use [Moment.js](https://momentjs.com/) and [Asciinema player](https://github.com/asciinema/asciinema-player) which may require a mention in the Dependencies)

uekerman (approved):
Thank you  🙏

PR #987, opened 2026-09-16
Add local-hugo wrapper for Imported documentation

MuhammadAashirAslam:
Adds a Python wrapper (`tools/local-hugo`) to simplify live-editing imported modules locally, as suggested in https://github.com/precice/precice.github.io/pull/979#issuecomment-5681611422.

Instead of manually setting `HUGO_MODULE_REPLACEMENTS`, you can point to local repository clones using simple flags:

```bash
# Live-edit tutorials
tools/local-hugo --tutorials=../tutorials server

# Live-edit multiple modules
tools/local-hugo --openfoam=../openfoam-adapter --tutorials=../tutorials server
```

The script resolves relative paths to absolute paths, sets HUGO_MODULE_REPLACEMENTS, and executes Hugo. All standard Hugo commands and arguments (like server, --bind 0.0.0.0, etc.) are passed straight through.


MuhammadAashirAslam:
@MakisH should we also document this in README? i added a help command in the script itself `tools/local-hugo --help` .

tools/local-hugo
```diff
@@ -0,0 +1,154 @@
...
+    modules: set[str] = set()
+
+    module_toml = repo_root / "config" / "_default" / "module.toml"
+    if module_toml.exists():
+        for line in module_toml.read_text(encoding="utf-8").splitlines():
+            match = IMPORT_PATH.match(line.strip())
+            if match:
+                modules.add(match.group(1))
+
+    go_mod = repo_root / "go.mod"
```
fsimonis, line 32: Shouldn't this be the "truth"? Why is it necessary to also check the  `module.toml`?

tools/local-hugo
```diff
@@ -0,0 +1,154 @@
...
+                    sys.exit(1)
+                replacements[module_path] = resolved_path
+                i += 1
+                continue
+
+        hugo_args.append(arg)
+        i += 1
+
+    if not hugo_args:
+        hugo_args = ["server"]
```
fsimonis: Let's use arparse here.
You can use the `modules` to generate options with default None value.
When you use parse args, the resulting namespace has only defines options.
There is also an option to return non-consumed arguments, these you can directly forward to hugo.

tools/local-hugo
```diff
@@ -0,0 +1,154 @@
...
+        print("[local-hugo] Module replacements active:")
+        for rep in replacement_strings:
+            print(f"  • {rep}")
+        os.environ["HUGO_MODULE_REPLACEMENTS"] = ", ".join(all_replacements)
+
+    print(f"[local-hugo] Executing: hugo {' '.join(hugo_args)}\n")
+    sys.stdout.flush()
+
+    try:
+        os.execvp("hugo", ["hugo"] + hugo_args)
```
fsimonis: The standard is to use `subprocess.run` with the env option. You can passthrough the hugo exit code by giving it to `sys.exit()`.

tools/local-hugo
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/987/changes/f247b6dfc138fc883f0bf7a55379209f9d696aff

tools/local-hugo
MuhammadAashirAslam: Done [`059af8a`](https://github.com/precice/precice.github.io/pull/987/commits/059af8a00cf195acc7eb7b726069b1f1646e78a5)
Just a question : with `os.execvp`, the Python process is replaced entirely by Hugo, whereas `subprocess.run` keeps Python running in the background waiting for the child process? Isn't it better to have Hugo as the main process?

tools/local-hugo
MuhammadAashirAslam: Partially true , all the main modules are in go.mod , checking `module.toml` is meant as a fallback safety step in case someone is integrating a brand new repository/adapter that isn't in `go.mod` yet (e.g. testing mounts locally before running `hugo mod get`).

fsimonis: > in case someone is integrating a brand new repository/adapter that isn't in go.mod yet

Why not add the new repo with the local path for testing to `go.mod`. Seems way easier to me than having to fix all mount paths when I am done testing. 

The main problem I see is that processing the go.mod straight-forward while processing the modules mapping is messy.

tools/local-hugo
```diff
@@ -0,0 +1,116 @@
...
+            if not resolved.is_dir():
+                print(
+                    f"Error: Directory for '{mod_path}' does not exist: {resolved}",
+                    file=sys.stderr,
+                )
+                sys.exit(1)
+            replacements[mod_path] = resolved
+
+    if not hugo_args:
+        hugo_args = ["server"]
```
fsimonis, line 80: No need for extra treatment, just pass everything to hugo

tools/local-hugo
```diff
@@ -0,0 +1,116 @@
...
+        usage="%(prog)s [MODULE_OPTIONS] [HUGO_ARGS...]",
+    )
+
+    for mod in canonical_modules:
+        short = mod.removeprefix(PRECICE_MODULE_PREFIX)
+        aliases = [f"--{short}"]
+        if short.endswith("-adapter"):
+            aliases.append(f"--{short.removesuffix('-adapter')}")
+        parser.add_argument(
+            *aliases,
```
fsimonis, line 57: This alias unnecessarily complicates things. Let's keep it easy

```suggestion
        short = mod.removeprefix(PRECICE_MODULE_PREFIX)
        parser.add_argument(
            f"--{short}",
```

tools/local-hugo
```diff
@@ -0,0 +1,116 @@
...
+
+    parsed, hugo_args = parser.parse_known_args()
+
+    replacements: dict[str, Path] = {}
+    for mod_path, local_path in vars(parsed).items():
+        if local_path is not None:
+            resolved = local_path.expanduser().resolve()
+            if not resolved.is_dir():
+                print(
+                    f"Error: Directory for '{mod_path}' does not exist: {resolved}",
```
fsimonis, line 73: You can build this check into argparse by providing a function as type, resolving the path and raise an exception when the path doesn't exist.

This completely avoids this block.

tools/local-hugo
fsimonis: > Isn't it better to have Hugo as the main process?

Low level calls can be surprising. Let's just use what most python devs know, and that is subprocess.

tools/local-hugo
MuhammadAashirAslam: Removed it in https://github.com/precice/precice.github.io/pull/987/changes/ed497c8f952810b36f8aa161775c12875e03f6ef , let's just avoid unnecessary complexity.

tools/local-hugo
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/987/changes/a69154ed4189c60582b91f0cdfac2d43778b0172

tools/local-hugo
MuhammadAashirAslam: Done https://github.com/precice/precice.github.io/pull/987/changes/524cb0ac2f67a03edac1c845230239a25d20bfc2

